PR #35, opened 2016-09-21
Support for disabled buttons

tolu360:
Perhaps you can add built-in support for 'disabled` buttons. Great work so far, by the way!

### DIFF
--- a/Readme.MD
+++ b/Readme.MD
@@ -153,6 +153,7 @@ import { Button } from 'react-native-elements'
 | fontSize | 18 | number | font size (optional)  |
 | underlayColor | transparent | string(color) | underlay color for button press (optional)  |
 | raised | false | boolean | flag to add raised button styling (optional)  |
+| disabled | false | boolean | prop to indicate button is disabled (optional) |
 
 
 ## Social Icons & Buttons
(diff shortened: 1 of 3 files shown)


dabit3:
@tolu360 Thanks, this is a great starting point. What do you think about having a default disabled color such as a light shade of grey? Also, how about the possibility of disabling any onPress method if the button is disabled? If you want to update the pr with something like this go ahead, but if you would like me to add this to the next version I can also do so. 

Thanks again for putting this here :)

tolu360:
Hi @dabit3, so I took your suggestion because it was just brilliant. Have updated the PR to include a default color for the disabled state for buttons, please feel free to alter my color choice (added it to the colors config to allow for easy updates). As per, preventing `onPress` from firing when button is disabled, we get that for free - out of the box, RN handles that quite well.

dabit3:
@tolu360 Hey thanks so much! I Have not forgotten about you, just been busy and need to test this out. I'll hopefully have this merge by the end of the week. Thanks again :)

MelanieCLopez:
ok thank you ;)

tolu360:
That's fine @dabit3

dabit3:
@tolu360 thank so much for this pr, it's a great addition!

PR #40, opened 2016-09-22
allow for static image resource

hiaw:
This will allow user to use static image as well as network images


hiaw:
by the way, thanks for doing the react native radio. i enjoy listening to it.

dabit3:
@hiaw Great PR. Thanks so much! Also, thanks for listening to the show, we really enjoy doing it :)

PR #53, opened 2016-09-28
fixes #52

### LINKED ISSUE #52: ListItem rightIcon failed prop type - Warning
propTypes for rightIcon in ListItem.js defaults to type string whereas it requires an object.

TestCase: provide an object with name and type to rightIcon
Observed behavior: An yellow warning for invalid prop type and a "?" displayed in UI
Required Behavior: UI shows the icon from icon type

futhrevo:
changed proptype for rightIcon to object and default value to  an object


dabit3:
@futhrevo thanks!

PR #59, opened 2016-10-04
Allow for multi selection for button group

hiaw:
able to select multiple buttons in button group


Monte9:
@hiaw This PR looks good. Thanks for submitting it. Although I am not sure how useful it is. 
The button group is suppose to behave similar to UISegmentedControl on iOS and from what I understand the point it to have only 1 selected at any given point. 
What use case where you thinking?

dc-smish:
For example, you have a filter list of buttons. Like for memory you can have 
16GB
32GB
64GB
and users are able to see both 16GB and 64GB.

Monte9:
@dc-smish makes sense. Thanks for the example. 

@hiaw I'll look into the code and merge it this evening.

Monte9:
The code lgtm, although this PR changes the default behaviour of `Button Group` to multi-selection. 

Imo, `multi-selection` should be optional and the user should be able to pass a prop `multiselection={true}` in order to enable this. By default, the behaviour should be that of single selction like a UISegmentedControl. 

@binoy14 What are your thoughts on this PR?

src/buttons/ButtonGroup.js
```diff
@@ -49,11 +50,9 @@ const ButtonGroup = ({
...
                     styles.buttonText,
                     textStyle && textStyle,
-                    selectedIndex === i && {color: colors.grey1},
-                    selectedIndex === i && selectedTextStyle
-                  ]}>{button}</Text> 
-                )
-              }
+                    selectedIndex.includes(i) && {color: colors.grey1},
+                    selectedIndex.includes(i) && selectedTextStyle && selectedTextStyle
+                  ]}>{button.toUpperCase()}</Text>
```
binoy14: @hiaw Your missing closing `)}`

binoy14:
@Monte9 I agree that multi selection should be optional available thru a prop. The other thing i also noticed was that the ButtonGroup did not have a default height is that the case or was I doing something wrong?

Monte9:
Closing this PR for now. @hiaw is MIA plus this PR could use some work before it is actually merged.

PR #60, opened 2016-10-05
Hint for reverse trick: make icons look like button

mbeaudru:
It didn't sound really obvious to me that in order to use "Icon Buttons" as stated in the title of the section we have to put "reverse" as a props to the Icon component.

In fact, I had to try all samples to understand that, so it might be the case for others too.


dabit3:
Great point, thanks so much for the pr!

dabit3:
@mbeaudru

Issue #56, opened 2016-10-01
Create Activity Indicator for elements


Jeanmsilva89:
I was wondering if other people would be interested in Loading Indicator prop for elements such as a Button and FormInput. I'm thinking of just adding a prop called "loading" and it would wrap the ActivityIndicator in the button container for example. 

Before creating a PR, I wanted to know if there would be interest from the community on such feature.

Thanks.

dabit3:
I think this would be a great addition for sure! @Jeanmsilva89

Firaenix:
I'd love something like this for my project!

Jeanmsilva89:
#62

Jeanmsilva89:
I'd love to add this feature to other components, and I'll have to do it for my own projects so I'll submit another PR soon for that.
@dabit3 Another topic but figured I'd ask you here instead of opening another issue. 
Is there a reason for why you decided to create the SocialIcon component from scratch as opposed to building it on top of your button component? In a lot of ways the social media icons are used as buttons and could share lot of the button features. 
For example. I just created a branch where I replicated the props. "small, disabled, loading" into the socialIcon component. I initially assumed it was built using the button so I thought it would've just worked. 

Btw thanks a lot for this cool project and for you awesome React Native Radio, been really enjoying it big time. Keep up!

Monte9:
@Jeanmsilva89 I really like what you have done with the activity indicators on the button elements. gj!
What other components were you thinking of adding the activity indicator too?

Also I think, as part of your PR you forgot to update the readme with the new button props that you implemented. Feel free to submit a PR for this if you are not busy.

dekm:
Would be cool to have this added to the Searchbar as well.  I was just trying to look into adding a spinner on the right side of the searchbar as I'm calling search as the user types.

Looks like inside the search component there is only the left icon currently added.

Jeanmsilva89:
@Monte9 Sorry I have been too busy in the past few weeks, and missed your comment. I'll send a PR for the README by this weekend. 

@dekm I'm not using the SearchBar currently but yeah I think it's useful for just about every component that has an onPress event so that a loading state can be added.

Monte9:
@Jeanmsilva89 any updates?

Monte9:
@binoy14 since we haven't heard from Jean, maybe this is something we can do ourselves. We can talk about what elements it would make sense to add the activityIndicator to. 

[...]

PR #62, opened 2016-10-06
Activity Indicator on Button Element Issue

Jeanmsilva89:
**Issue** #56 
Accepts the following new props. 
-   **loading**: PropTypes.bool
-   **activityIndicatorStyle**: PropTypes.any
-   **loadingLeft**: PropTypes.bool

By default the activityIndicator shows up on the right side
It also uses the small prop from the button

### DIFF
--- a/src/buttons/Button.js
+++ b/src/buttons/Button.js
@@ -15,7 +15,7 @@ const Button = ({
   Component,
   disabled,
   loading,
-  loadingLeft,
+  loadingRight,
   activityIndicatorStyle,
   buttonStyle,
   borderRadius,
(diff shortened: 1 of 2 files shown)


dabit3:
@Jeanmsilva89 Thanks so much for the pr! Question: should `!loadingLeft` instead be on the right of the Text component or am I missing something ( still not fully awake ).

Jeanmsilva89:
@dabit3 Oops. You're actually awake, I'm the one who must've been sleeping. 
The prop should actually be loadingRight instead. I made the loading component to be on the left by default. Unless you think it should be on the Right by Default. If you think that will be more standard if we add this other components. 

I'll update when I get a break at work today.

dabit3:
No this looks great. Once you update it I will merge it and get to releasing a new version :) Thanks again for the pr @Jeanmsilva89

PR #63, opened 2016-10-07
Fix styling prop for ButtonGroup

darjanin:
The `containerStyle` should style the border around button group according to the documentation.
> containerStyle	inherited styling	object (style)	specify styling for main button container (optional)
It doesn't work, because it's on component and there is no other way to change the border's color.


dabit3:
Thanks so much @darjanin !

PR #68, opened 2016-10-13
Add normalize text helper function

### LINKED ISSUE #67: Font size normalization across pixel ratios
Font sizes on iphone6 and iphone4 are particularly small.  It would be good to use a "resizing" function to help with this across pixel ratios.

If you would like me to submit a PR please let me know.

const React = require('react-native')
const {
  PixelRatio
} = React;
const pixelRatio = PixelRatio.get();

const normalize = (size) => {
    if (pixelRatio == 2) {
        return size * 1.15;
    } 
    if (pixelRatio == 3) {
        return size * 1.35;
    }
    return size * pixelRatio;
}

export default normalize;

Usage:
import normalize from '../helpers/normalizeText';

fontSize:  normalize(14)

====
Code example from: https://jsfiddle.net/97ty7yjk/ and https://stackoverflow.com/questions/34837342/font-size-on-iphone-6s-plus.

xiaoneng:
Fix #67 with helper function to normalize text size across devices.

Tested:
Devices:  iphone 5, iphone 5s, iphone 6, xiaomi 4, xiaomi 5
Simulators:  iphone 6s, iphone 6s plus, iphone 7, iphone 7 plus, Samsung Galaxy S4

This still needs some work to catch all android devices and get all ratios equal but hopefully decent starting point.

### DIFF
--- a/src/buttons/Button.js
+++ b/src/buttons/Button.js
@@ -4,6 +4,7 @@ import colors from '../config/colors'
 import Text from '../text/Text'
 import MaterialIcon from 'react-native-vector-icons/MaterialIcons'
 import getIconType from '../helpers/getIconType'
+import normalize from '../helpers/normalizeText'
 
 let styles = {}
 
@@ -190,7 +191,7 @@ styles = StyleSheet.create({
   },
   text: {
     color: 'white',
-    fontSize: 18
+    fontSize: normalize(18)
   },
   icon: {
     marginRight: 10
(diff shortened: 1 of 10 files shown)


dabit3:
Hey @xiaoneng thanks so much for the pr ! :)

turnipdabeets:
@xiaoneng This is just for my own curiosity - why do we need to normalize fontSize if React Native uses logical pixels by design? My understanding is that logical pixels translate to iOS's points of Android's dp measure so it would automatically handle different screen scales.

xiaoneng:
@turnipdabeets i didn't put that much thought into it...necessity was the guide.  had 10+ different ios / android devices ( some of them older) in an environment where text needed to look relatively the same size on each, not run off the screen or wrap and text readability was absolutely necessary.  at the time, just sticking a 12pt/14pt font in there wasn't working so i did this and scratched my itch...

PR #75, opened 2016-10-26
Pass onButtonPress prop to PricingCard Component

zibs:
I noticed when using the PricingCard component that I was unable to pass an onPress function to the component's button. This fixes that! 

Thanks for the great lib.


dabit3:
Thanks so much, this is great! I appreciate the pr.

PR #79, opened 2016-10-30
Fix propTypes for Icon component

peteychuk:
Got a warning when I try to use style created via StyleSheet.create. This PR fixing it.

I also double checked propTypes for rest components - they haven't this issue because all propTypes are correct (equal to PropTypes.any).

Let me know if you have any questions.


dabit3:
Thanks, this looks great!

PR #83, opened 2016-11-08
Wrong anchor link and order in the summary

mbeaudru:


dabit3:
Thanks so much @mbeaudru !

PR #89, opened 2016-11-11
ButtonGroup text labels shouldn't force upper case

andreioprisan:
By default, ButtonGroup elements convert the text labels to upper case. Since react native doesn't support text transforms via css styles (and they'd be overwritten anyways by the button.toUpperCase() call, can we not force the upper casing on all text and instead just pass through the initial button text labels?


dabit3:
@andreioprisan great point, thanks!

PR #93, opened 2016-11-15
[FIX] add hitSlop to button

note89:
The TouchableWithoutFeedback prop `hitSlop` was missing from button. This adds that.
https://facebook.github.io/react-native/docs/touchablewithoutfeedback.html#props

### DIFF
--- a/src/buttons/Button.js
+++ b/src/buttons/Button.js
@@ -29,6 +29,7 @@ const Button = ({
   onLongPress,
   onPressIn,
   onPressOut,
+  hitSlop,
   activeOpacity,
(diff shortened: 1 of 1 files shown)


note89:
Is there a way to user `...` instead to avoid similar issues in the future ?

note89:
This does not work on android. 
```
Error while updating property 'hitSlop' of a view managed by: RTC View
```

note89:
Okey on android you need to supply all `4` fields ! `top, bottom, right, left` or this crashes.

dabit3:
Thanks @note89 !

PR #103, opened 2016-12-01
Allow list item title and subtitle to accept object proptype as well …

xiaoneng:
…as string proptype

This allows one to embed more complex <View> </View> into title or subtitle of a ListItem component.


dabit3:
Thanks for the pr! Will be releasing a new version sometime this evening with these changes.

PR #111, opened 2016-12-09
Fix typo (remove duplicated condition)

javiercr:
I guess that `borderStyle && borderStyle` was a typo :)


dabit3:
Thanks @javiercr !

PR #131, opened 2016-12-29
Fix: Button - Defaulting to small broke some props

mcnamee:
`!large ...` conditional placement was ignoring fontSize and buttonStyles props


dabit3:
thanks @mcnamee !

PR #134, opened 2016-12-31
V0.9.0

Monte9:
@dabit3 Please review. Thanks.


src/slider/Slider.js
```diff
@@ -35,7 +35,7 @@ Rect.prototype.containsPoint = function(x, y) {
     && y <= this.y + this.height)
 }
 
-class Slider extends Component {
+export default class Slider extends Component {
```
Monte9: @dabit3 @binoy14 need your help debugging something. 

When I try to import Slider I get a `Element type is invalid:`.  Spent some time digging into this and I think it has to do with module.export Here's the exact error:

screen shot 2017-01-01 at 9 33 02 am

`import Slider from './Slider'` although it breaks when I import it from `react-native-elements`. If you have some time, can you look into this? Thanks.

Monte9:
Refer to [this](https://github.com/react-native-community/react-native-elements/issues/125) for release notes.

Issue #133, opened 2016-12-29
ListItem title & subtitle should be View tags


ngduc:
Hi,
Right now, ListItem subtitle is rendered as a Text tag which causes a JS error with Android if subtitle contains a View tag.

https://github.com/react-native-community/react-native-elements/blob/4a73fdf1f8de2556eb36a8497df5e7cb2963144e/src/list/ListItem.js#L85-L91

Solution:
=> render {subtitle} as a View
- Similar to title element also.
Thanks.

![image](https://cloud.githubusercontent.com/assets/639189/21555116/07c5a238-cdca-11e6-8243-41dc29189a6a.png)

Monte9:
hi @ngduc! This has been requested before and it has been already implemented ->`ListItem subtitle prop can now render custom Views`.
This is part of our next release v0.9.0 that will be out soon. To keep up with the release check this out: https://github.com/react-native-community/react-native-elements/issues/125

ngduc:
@Monte9 How about {title} element, please also make it a View for easy styling. It's hard to style when we have nested Text tags. Many thanks!

https://github.com/react-native-community/react-native-elements/blob/v0.9.0/src/list/ListItem.js#L76-L82

Monte9:
I thought that was already the case, but apparently not. Thanks for pointing it out. 
Reopening this issue until this is added into v0.9.0

PR #164, opened 2017-01-28
ListItem title can be a View tag

amaurymartiny:
Implements #133.


dabit3:
Very nice @amaurymartiny thanks :)

PR #166, opened 2017-02-01
feat(ListItem): support both right title and chevron

Blinnikov:
Support both right title and chevron


dabit3:
Looks great, thx @Blinnikov :)

PR #175, opened 2017-02-10
added some entries to the Search Bar documentation regarding props

temitope:
specifically for placeholder, placeholderTextColor and onChangeText


Monte9:
@temitope Thanks for the PR. 👍

PR #176, opened 2017-02-11
Fixed #78 Invalid Icon 'component' propType

### LINKED ISSUE #78: Invalid prop 'component' on Icon
```
when I specify a component on an Icon like so:

<Icon
  reverse
  raised
  name="chat"
  color="lightblue"
  component={TouchableOpacity}
  onPress={() => console.log('hello')}
/>
```

it gives this warning Failed prop type: Invalid prop 'component' of type 'function' supplied to 'Icon', expected a single ReactElement

max2three:
I think React Native Components are mostly 'func' type. 'element' type returns true if it is constructed using `React.createElement()`.

### DIFF
--- a/src/icons/Icon.js
+++ b/src/icons/Icon.js
@@ -70,7 +70,7 @@ Icon.propTypes = {
   name: PropTypes.string,
   size: PropTypes.number,
   color: PropTypes.string,
-  component: PropTypes.element,
+  component: PropTypes.func,
   underlayColor: PropTypes.string,
   reverse: PropTypes.bool,
   raised: PropTypes.bool,


Monte9:
@max2three can you link me to any documention that explicilty states this? Thanks.

max2three:
@Monte9 
[React.PropTypes.element](https://facebook.github.io/react/docs/react-api.html#react.proptypes.element)
[React Element](https://facebook.github.io/react/docs/react-api.html#creating-react-elements) 
[React Component](https://facebook.github.io/react/docs/components-and-props.html)

So, I think validating a component against element type will always gives you a warning as they are not the same.

Correction: Calling `createElement()` explicitly if you are not using JSX.

max2three:
Closed this PR as issue #78 is closed.

PR #194, opened 2017-02-22
Updated ListItem in README

max2three:
`rightIcon` will be hidden only if `hideChevron` is set after PR #166 was merged to master.


Monte9:
Thanks for the PR @max2three!

PR #197, opened 2017-02-24
Avatar component

Monte9:
Please review @binoy14 

TODO:
- [x] Update props to be deconstructed within the component itself and pass in ( ...props) outside.
- [x] Add ability to render initials too (either image or initials for avatar)
- [x] Update README to include example and docs for Avatar component

Example:
``` js
<Avatar
  small
  rounded
  icon={{type: 'home', size: 23}}
  onPress={() => console.log("Works!")}
  activeOpacity={0.7}
  containerStyle={{flex: 2, marginLeft: 20, marginTop: 115}}
/>
<Avatar
  medium
  source={{uri: "https://s3.amazonaws.com/uifaces/faces/twitter/ladylexy/128.jpg"}}
  onPress={() => console.log("Works!")}
  activeOpacity={0.7}
  containerStyle={{flex: 3, marginTop: 100}}
/>
<Avatar
  large
  rounded
  title="MT"
  onPress={() => console.log("Works!")}
  activeOpacity={0.7}
  containerStyle={{flex: 4, marginTop: 75}}
/>
<Avatar
  xlarge
  rounded
  title="MT"
  source={{uri: "https://s3.amazonaws.com/uifaces/faces/twitter/ladylexy/128.jpg"}}
  onPress={() => console.log("Works!")}
  activeOpacity={0.7}
  containerStyle={{flex: 5, marginRight: 60}}
/>
```

Here's what it looks like: 

screen shot 2017-02-26 at 7 07 37 pm


src/avatar/Avatar.js
```diff
@@ -0,0 +1,131 @@
...
+  } else if (!height) {
+    height = width
+  }
+
+  const styles = StyleSheet.create({
+    container: {
+      paddingTop: 10,
+      paddingRight: 10,
+      paddingBottom: 10,
+      borderBottomColor: '#ededed',
```
binoy14: @Monte9 What is the reason for adding a borderBottom?

binoy14:
@Monte9 Few Issues I found.

1. The component is not exported in index.js
2. When the width is less than ~150 the title does not show up. It is probably because font is bigger and it's overflowing 
3. Based on the previous comment I am not sure what is the need for having a bottom border 
4. For destructuring you can use the following snippet.

```js
const Avatar = (props) => {
  const {
    component,
    onPress,
    onLongPress,
    containerStyle,
    source,
    avatarStyle,
    roundAvatar,
    title,
    titleStyle,
    overlayContainerStyle,
    activeOpacity,
  } = props;

// Because we are overriding width and height it can't be a const
  let {
    width,
    height,
  } = props;

  if (!width && !height) {
    width = 34
    height = 34
  } else if (!width) {
    width = height
  } else if (!height) {
    height = width
  }

  let Component = onPress || onLongPress ? TouchableOpacity : View
  if (component) {
    Component = component
  }

  return (
    <Component
      {...props}
      style={[styles.container, containerStyle && containerStyle]}>
```

src/avatar/Avatar.js
Monte9: That's unnecessary. I'll there caused I copy pasted the style from another component. 
I'll remove it.

Monte9:
@binoy14 Thanks for your feedback. I have addressed all PR comments. And I have also added support for Icons & initials as discussed. 
Check out the updated PR description for latest avatar samples.

binoy14:
@Monte9 I added the Props in the docs as we discussed. Everything lgtm and can be merged anytime

PR #200, opened 2017-02-25
when set image property on card container - throws layout error (fixe…

vaderdan:
It appears that if parent element don't have style flex:1, but child have (or parent don't have height hard set), the react throws error.

I'm not sure what is the idea about adding flex property here - if I remove it - visually the component looks right - it expands image to full width of the Card element.

### DIFF
--- a/src/containers/Card.js
+++ b/src/containers/Card.js
@@ -38,10 +38,10 @@ const Card = ({
       }
       {
         image && (
-          <View style={{flex: 1}}>
+          <View>
             <Image
               resizeMode='cover'
-              style={[{flex:1, width: null, height: 150}, imageStyle && imageStyle]}
+              style={[{width: null, height: 150}, imageStyle && imageStyle]}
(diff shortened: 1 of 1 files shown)


vaderdan:
the error is thrown from this line here on react-native
https://github.com/facebook/react-native/blob/master/React/Views/RCTShadowView.m#L137

binoy14:
@vaderdan Can you provide screenshots of before and after. I tried it with and without your changes and didn't change anything in the card.

vaderdan:
ok, 
this is my test repository, so you can see all dependencies (react-native 0.41.2)
https://github.com/vaderdan/testCard


The error from issue, shows when I add `flexDirection: 'row'` style on the container
Otherwise it looks like this:



Or when I add height it could shrink



But it should look like this (auto expand card height when add child elements):

binoy14:
@vaderdan Got it. Thank you for the PR!

Issue #182, opened 2017-02-15
[Feature] Add close icon props in SearchBar


abdurrahmanekr:
Hi, What do I have to do when I want to delete what I have written? 
Like SearchBar.closeIcon={true}. Thank you.

binoy14:
@abdurrahmanekr There doesn't seem to be an close icon. I think it will be a great addition. Thanks for reporting

Monte9:
Thanks for the PR @binoy14! 

@abdurrahmanekr the PR has been reviewed and will me merged soon. Keep an eye out for our next release v0.10. You can follow along with updates [here](https://github.com/react-native-community/react-native-elements/issues/195)

PR #202, opened 2017-02-28
Added clear icon in Search box

binoy14:
- Updated docs for the clear icon

In regards to #182 

screen shot 2017-02-27 at 9 20 20 pm

### DIFF
--- a/Readme.MD
+++ b/Readme.MD
@@ -639,7 +639,8 @@ import { SearchBar } from 'react-native-elements'
 | placeholder | '' | string | set the placeholder text |
 | placeholderTextColor | '#86939e' | string | set the color of the placeholder text |
 | onChangeText | none | function | method to fire when text is changed |
-| clearIcon | { color: '#86939e', name: 'search' } | object {name (string), color (string), style (object)} | specify color, styling, or another [Material Icon Name](https://design.google.com/icons/) Note: pressing on this icon clears text inside the searchbar |
+| clearIcon | { color: '#86939e', name: 'search' } | object {name (string), color (string), style (object)} | specify color, styling, or another [Material Icon Name](https://design.google.com/icons/)
+(Note: pressing on this icon clears text inside the searchbar) |
 
 ## Tab Bar Component
 
--- a/Readme.MD
+++ b/Readme.MD
@@ -592,7 +592,7 @@ render () {
 
 ## Search Bar
 
-![Search Bar](https://i.imgur.com/mvPgPfg.png)
+<img src=“https://i.imgur.com/mvPgPfg.png” width=“300” />
 
 ```js
 import { SearchBar } from 'react-native-elements'
--- a/Readme.MD
+++ b/Readme.MD
@@ -639,7 +639,7 @@ import { SearchBar } from 'react-native-elements'
 | placeholder | '' | string | set the placeholder text |
 | placeholderTextColor | '#86939e' | string | set the color of the placeholder text |
 | onChangeText | none | function | method to fire when text is changed |
-
+| clearIcon | { color: '#86939e', name: 'search' } | object {name (string), color (string), style (object)} | specify color, styling, or another [Material Icon Name](https://design.google.com/icons/) Note: pressing on this icon clears text inside the searchbar |
 
 ## Tab Bar Component
 
--- a/Readme.MD
+++ b/Readme.MD
@@ -592,7 +592,7 @@ render () {
 
 ## Search Bar
 
-<img src=“https://i.imgur.com/mvPgPfg.png” width=“300” />
+<img src="https://i.imgur.com/mvPgPfg.png" width="300" >
 
 ```js
 import { SearchBar } from 'react-native-elements'
--- a/Readme.MD
+++ b/Readme.MD
@@ -592,7 +592,7 @@ render () {
 
 ## Search Bar
 
-![Search Bar](http://i.imgur.com/QvSrF62.png)
+![Search Bar](https://i.imgur.com/mvPgPfg.png)
 
 ```js
 import { SearchBar } from 'react-native-elements'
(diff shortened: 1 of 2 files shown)


Readme.MD
```diff
@@ -1424,4 +1424,4 @@ A component with full size image and with text either inside the image or under
 | overlayContainerStyle | none | object (style) | style for the view outside image or icon |
 | activeOpacity | 0.2 | number | opacity when pressed |
 | icon | none | object {name: string, color: string, size: number, type: string (default is material, or choose one of material-community, simple-line-icon, zocial, font-awesome, octicon, ionicon, foundation, evilicon, or entypo), iconStyle: object(style)} |
-| iconStyle | none | object (style) | extra styling for icon component (optional) |
\ No newline at end of file
+| iconStyle | none | object (style) | extra styling for icon component (optional) |
```
Monte9: This change is not related to this PR. Revert if not unintended?

Readme.MD
```diff
@@ -639,7 +639,7 @@ import { SearchBar } from 'react-native-elements'
 | placeholder | '' | string | set the placeholder text |
 | placeholderTextColor | '#86939e' | string | set the color of the placeholder text |
 | onChangeText | none | function | method to fire when text is changed |
-
+| clearIcon | { color: '#86939e', name: 'search' } | object {name (string), color (string), style (object)} | specify color, styling, or another [Material Icon Name](https://design.google.com/icons/) Note: pressing on this icon clears text inside the searchbar |
```
Monte9: Add punctuation between `..[Material Icon Name](https://design.google.com/icons/) Note: pressing..`

Maybe even put Note in a () for better readability.

src/input/Search.js
```diff
@@ -9,6 +9,10 @@ class Search extends Component {
     const ref = this.props.textInputRef
     this.refs[ref].focus()
   }
+  clear() {
```
Monte9: nitpicky. But newline between functions for better readability.

src/input/Search.js
```diff
@@ -9,6 +9,10 @@ class Search extends Component {
...
+    const ref = this.props.textInputRef
+    this.refs[ref].clear();
+  }
   render () {
```
Monte9: Same as above ^^

src/input/Search.js
```diff
@@ -9,6 +9,10 @@ class Search extends Component {
     const ref = this.props.textInputRef
     this.refs[ref].focus()
   }
+  clear() {
+    const ref = this.props.textInputRef
+    this.refs[ref].clear();
+  }
```
Monte9: Same as above ^^

src/input/Search.js
```diff
@@ -132,6 +137,20 @@ class Search extends Component {
...
         {
+          clearIcon && (
+            <Icon
+              size={16}
+              style={[
+                styles.clearIcon,
+                clearIcon.style && clearIcon.style
+              ]}
+              name={clearIcon.name || 'close'}
+              onPress={this.clear.bind(this)}
```
Monte9: rename function to make intent more explicit? Something like: `clearText` would work too.

src/input/Search.js
```diff
@@ -224,6 +243,17 @@ const styles = StyleSheet.create({
   },
   inputLight: {
     backgroundColor: colors.grey4
+  },
+  clearIcon: {
+    backgroundColor: 'transparent',
+    position: 'absolute',
+    right: 16,
+    top: 15.5,
```
Monte9: Can we avoid fractions?

Monte9:
Lgtm. Few minor comments. 

Also, do you think it would be too much work to update the image on the readme as well? There are several `searchBars` on the left.. maybe we can update just one of them that shows the `clearIcon`. This is not very important, just thinking about loud. Thoughts?

danqing:
Does the clear icon simply get rid of the text in the search bar, or is there a functionality to dismiss (un-focus) the search bar? I think some sort of "Cancel" button will be very useful.

binoy14:
@danqing it just clears the text inside the text input. Can you elaborate more on clear button? Maybe expose the onPress function so users can add custom function?

danqing:
@binoy14 so most search bars (at least on iOS) have this Cancel button at the right (when focused). When you click on the button, it unfocuses the search bar and dismisses the keyboard.

img_0757

binoy14:
@danqing I feel that is not part of the search bar but can be done using a button and searchbar in a Grid component. Do you think it should be part of the searchbar?

@Monte9 Thougts?

Monte9:
@danqing this is a great suggestion. But I agree with @binoy14. Currently the best approach to getting the above search bar would be to couple it with a button using a grid layout or flex. And I like this approach better for now as it give you full control of what to do with the cancel onPress.

src/input/Search.js
binoy14: @Monte9 I copied it from the style of search icon. Do you still want me to change it?

danqing:
Understood.

I personally think that a full-featured search box (that includes this standard cancel button) will be very welcomed, but I agree that it can be achieved with a combo of things.

binoy14:
@danqing Maybe we can create another component with all the options. Feel free to submit a PR and we will review it.

danqing:
Will look into it.

src/input/Search.js
Monte9: I see it. Ok, let's leave it as is then.

Monte9:
@binoy14 PR looks ready to me. 

What do you think about this?
> Also, do you think it would be too much work to update the image on the readme as well? There are several searchBars on the left.. maybe we can update just one of them that shows the clearIcon. This is not very important, just thinking out loud.

binoy14:
@Monte9 Working on the screenshots. Will also create a draft of 0.10 release 👍

Issue #210, opened 2017-03-04
invalid style propTypes on Col and Row 


binoy14:
When using StyleSheet.create

Monte9:
Thanks for reporting the issue & submitting a PR. 

The PR has been reviewed. Feel free to close the issue when you merge the PR.

Issue #209, opened 2017-03-04
When passing styles from stylesheet in grid causes error


binoy14:
![screen shot 2017-03-04 at 6 00 32 pm](https://cloud.githubusercontent.com/assets/6476108/23582931/7e88f31c-0104-11e7-9a7b-621969e707a3.png)

This is because of this code snippet

```js
this.styles = {
   flex: 1,
   flexDirection: this.isRow() ? 'column' : 'row',
   ...style,
};
```

Problem Detail: When using StyleSheet.create it is a number type which is blocked by the Object.assign polyfill for performance. 

Solution: Change the style to how other components in react-native-elements are.

Monte9:
Thanks for reporting the issue & submitting a PR. 

The PR has been reviewed. Feel free to close the issue when you merge the PR.

PR #211, opened 2017-03-04
Fix issues with Grid, Col and Row

binoy14:
- When passing styles using StyleSheet.create the type of of number
and Object.assign cannot deal with numbers
- Change style prop to containerStyle  **This is a breaking change**
- Check if child exists when checking for type

In Reference to #209 and #210

### DIFF
--- a/src/grid/Col.js
+++ b/src/grid/Col.js
@@ -1,30 +1,41 @@
 import React, { PropTypes } from 'react';
-import { View, TouchableOpacity } from 'react-native';
+import { View, StyleSheet, TouchableOpacity } from 'react-native';
 
 const Col = (props) => {
-  const {style, size, onPress, activeOpacity} = props;
+  const {containerStyle, size, onPress, activeOpacity} = props;
 
-  const styles = {
-    flex: (size) ? size : (style && style.width) ? 0 : 1,
-    flexDirection: 'column',
-    ...style,
-  };
+  const styles = StyleSheet.create({
+    container: {
+      flex: (size) ? size : (containerStyle && containerStyle.width) ? 0 : 1,
+      flexDirection: 'column',
+    },
+  });
 
   if (onPress) {
     return (
-      <TouchableOpacity style={{...styles}} activeOpacity={activeOpacity} onPress={onPress}>
+      <TouchableOpacity
+        style={[
+          styles.container,
+          containerStyle && containerStyle,
+        ]}
+        activeOpacity={activeOpacity}
+        onPress={onPress}
+      >
         <View
           {...props}
         >
           {props.children}
         </View>
       </TouchableOpacity>
-    )
+    );
   }
 
   return (
     <View
-      {...styles}
+      style={[
+        styles.container,
+        containerStyle && containerStyle,
+      ]}
       {...props}
     >
       {props.children}
@@ -34,7 +45,7 @@ const Col = (props) => {
 
 Col.propTypes = {
   size: PropTypes.number,
-  style: PropTypes.object,
+  containerStyle: PropTypes.any,
   onPress: PropTypes.func,
   activeOpacity: PropTypes.number,
 };
--- a/src/grid/Grid.js
+++ b/src/grid/Grid.js
@@ -3,23 +3,10 @@ import { View, StyleSheet, TouchableOpacity } from 'react-native';
 import Row from './Row';
 
 class Grid extends Component {
-  static propTypes = {
-    containerStyle: PropTypes.any,
-    onPress: PropTypes.func,
-    activeOpacity: PropTypes.number,
-  }
-
-  static defaultProps = {
-    activeOpacity: 1,
-  }
-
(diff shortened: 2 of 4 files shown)


src/grid/Col.js
```diff
@@ -1,30 +1,41 @@
...
 const Col = (props) => {
-  const {style, size, onPress, activeOpacity} = props;
+  const {containerStyle, size, onPress, activeOpacity} = props;
 
-  const styles = {
-    flex: (size) ? size : (style && style.width) ? 0 : 1,
-    flexDirection: 'column',
-    ...style,
-  };
+  const styles = StyleSheet.create({
```
Monte9: let's move `const styles = StyleSheet.create({` outside our `Col` object.

src/grid/Grid.js
```diff
@@ -1,12 +1,10 @@
 import React, { Component, PropTypes } from 'react';
-import { View, TouchableOpacity } from 'react-native';
+import { View, StyleSheet, TouchableOpacity } from 'react-native';
 import Row from './Row';
 
 class Grid extends Component {
-  styles;
-
   static propTypes = {
```
Monte9: Let's try to be consistent here. Move propTypes outside the `Grid` Class

Monte9: `Grid.propTypes = {
...
}`

src/grid/Grid.js
```diff
@@ -15,35 +13,38 @@ class Grid extends Component {
     activeOpacity: 1,
   }
 
+  styles = StyleSheet.create({
```
Monte9: Let's move this outside the `Grid` class to keep it consistent.

src/grid/Grid.js
```diff
@@ -15,35 +13,38 @@ class Grid extends Component {
...
-  }
-
   render() {
-    const {onPress, activeOpacity} = this.props;
+    const {onPress, activeOpacity, containerStyle} = this.props;
 
     if (onPress) {
       return (
-        <TouchableOpacity style={{flex: 1}} activeOpacity={activeOpacity} onPress={onPress}>
+        <TouchableOpacity style={opacityContainer} activeOpacity={activeOpacity} onPress={onPress}>
```
Monte9: this should be `styles.opacityContainer`. 
I doubt it's doing anything rn.

src/grid/Grid.js
```diff
@@ -15,35 +13,38 @@ class Grid extends Component {
...
           <View
-            {...this.styles}
+            style={[
+              this.styles.container,
```
Monte9: Also once you move `styles` outside the class, this should be refactored as `styles.container`

src/grid/Row.js
```diff
@@ -1,18 +1,26 @@
...
 const Row = (props) => {
-  const {style, size, onPress, activeOpacity} = props;
+  const {containerStyle, size, onPress, activeOpacity} = props;
 
-  const styles = {
-    flex: (size) ? size : (style && style.height) ? 0 : 1,
-    flexDirection: 'row',
-    ...style,
-  };
+  const styles = StyleSheet.create({
```
Monte9: Move this outside `Row` object.

Monte9:
PR looks good. Comments mostly related to code organization for consistency. 

> Change style prop to containerStyle This is a breaking change

Also, this ^ should go into the release notes for v0.10

src/grid/Col.js
binoy14: @Monte9 The reason to have styles inside the `Col` object is because we need access to the props `containerStyle` and `containerStyle.width`

src/grid/Grid.js
binoy14: @Monte9 The reason to have styles inside the `Grid` object is because we need access to the `isRow` function. If you know of a better way please let me know

src/grid/Grid.js
binoy14: Good catch!

src/grid/Grid.js
binoy14: Refer https://github.com/react-native-community/react-native-elements/pull/211/files/ff25371eb616200fd42a85e5fdd038e8cac8db53#r105789402

src/grid/Row.js
binoy14: @Monte9 The reason to have styles inside the `Row` object is because we need access to the props `containerStyle` and `containerStyle.height`

src/grid/Col.js
Monte9: Ah k. That makes sense.

src/grid/Grid.js
Monte9: Ah I didn't see that. Ok that's fine. 

I can think of a couple other ways, but for this it seems like an overkill since only one stlye prop is being changed.

src/grid/Grid.js
Monte9: refactor pending here? ^^

PR #225, opened 2017-03-16
Typo - it should be button not card?

nonotest:


Monte9:
@nonotest Good catch. 

Lgtm. Thanks for the fix.

PR #237, opened 2017-03-22
Add missing save param to the npm install command in CRNA readme

wpasinski:
Since `yarn add` command adds installed package to project dependencies, we should add a save param to the `npm install` command so they are both equal.


binoy14:
@wpasinski Thanks for the Pull Request!

PR #241, opened 2017-03-22
Use fontWeight and fontFamily rather than fontFamily for everything

brentvatne:
It turns out the previous commit almost fixed it but in a few places it was necessary to add fontWeight rather than specify it as fontFamily.


Monte9:
Lgtm. Thanks for the fixes @brentvatne

PR #254, opened 2017-03-30
enhancement(Button): Using custom icons

rajchourasia:
Enabling use of CustomFonts (https://github.com/oblador/react-native-vector-icons#custom-fonts) in Buttons


binoy14:
@rajchourasia this is great, thanks for the PR. Do you think you can add it to other components using icons?

rajchourasia:
@binoy14 Sure. Will try and get it done tomorrow.

rajchourasia:
@binoy14 Updated for Avatar component.

There was is exiting issue where Icon "type" is being used in place of Icon "name". Fixed it to similar implementation as in Button.

The default type used in Avatar is "FontAwesome" instead of "MaterialIocns" (Default type is "MaterialIcons in Buttons"), I left it as is as the default icon name "user" is not present in "MaterialIcons".

Please advice.

binoy14:
@rajchourasia Thanks for pointing out, @Monte9 might be working on fixing the icons issue in Avatar. Let's wait for his input

Monte9:
Thanks for the PR @rajchourasia!

I find that there are a couple fundamental flaws with this PR. 
1. The changes to Avatar components that pertain to enhancing the existing icon functionality should be done in a separate PR. This is to make sure every PR is self-contained and does only 1 thing well. 
2. I don't think we should have an entire section on `Custom Icons with Buttons` with examples. This is already documented under `react-native-vector-icons#custom-fonts` and we should not have to reproduce this on our readme again. IMO, we should just add a one-liner in the `Notes` section of our readme that says something like `All our components that use icons, also support custom icons. For examples, please refer [here](link)`.

@binoy14 curious to hear what you think of the above suggestions?

rajchourasia:
@Monte9 
Have made changes as per your suggestions. Let me know if any further changes are required.

Monte9:
@rajchourasia thanks for the refactoring. Although I am still not convinced we need a separate .md file explaining how to use custom fonts. Seems redundant to me as it is already documented in detail on the vector-icons repo. 

I think we can get away with just adding this in the notes section:
```All React Native Elements components using icons also support [CustomFonts feature](https://github.com/oblador/react-native-vector-icons#custom-fonts) in [react-native-vector-icons](https://github.com/oblador/react-native-vector-icons).```

@binoy14 what are your thoughts on this?

binoy14:
@Monte9 Agreed, can be a note

Monte9:
Closing this PR due to the following:
1. Author inactive
2. The docs are now moved out of the Readme. Hence this PR is outdated. 

@rajchourasia or anyone else that's interested in this, please look at the implementation in this PR for supporting custom fonts and and submit a new PR. Thanks.

PR #258, opened 2017-03-30
ImageWrapperStyle prop for Card component to fix card image flex

Joshuabaker2:
Fixes the bug introduced in commit #200 where we are no longer able to set flex for image and wrapper, as the parent View component has no flexbox. With this, a style of `{flex: 1}` can once again be set on the View. 

Addresses issue #259


binoy14:
@Joshuabaker2 Thanks for the PR!

Issue #260, opened 2017-03-31
Optional text input component for list item


Joshuabaker2:
It would be nice to have the ability to input text where the rightTitle is in the listitem. Making a PR for this.

PR #261, opened 2017-03-31
Add optional text input component for list item

Joshuabaker2:
This addresses #260, giving the ability to have a text input component within the listitem component.

### DIFF
--- a/src/list/ListItem.js
+++ b/src/list/ListItem.js
@@ -1,5 +1,5 @@
 import React, { PropTypes } from 'react'
-import { View, StyleSheet, TouchableHighlight, Image, Platform } from 'react-native'
+import { View, StyleSheet, TouchableHighlight, Image, Platform, TextInput } from 'react-native'
 import Badge from '../badge/badge'
 import Icon from '../icons/Icon'
 import Text from '../text/Text'
@@ -37,6 +37,22 @@ const ListItem = ({
   badgeTextStyle,
   label,
   onLongPress,
+  textInput,
+  textInputAutoCapitalize,
+  textInputAutoCorrect,
+  textInputAutoFocus,
+  textInputEditable,
+  textInputKeyboardType,
+  textInputMaxLength,
+  textInputMultiline,
+  textInputOnChangeText,
+  textInputOnFocus,
+  textInputOnBlur,
+  textInputSelectTextOnFocus,
+  textInputReturnKeyType,
+  textInputValue,
+  textInputStyle,
+  textInputContainerStyle
 }) => {
   let Component = onPress || onLongPress ? TouchableHighlight : View
   if (component) {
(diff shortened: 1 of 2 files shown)


src/list/ListItem.js
```diff
@@ -222,8 +278,11 @@ styles = StyleSheet.create({
     marginRight: 5,
     color: colors.grey4
   },
-  chevron: {
-  }
+  textInputStyle: {
+    height: 20,
+    textAlign: 'right'
+  },
+  chevron: {}
```
binoy14: @Joshuabaker2 small thing but can you remove unused styles. Rest all looks good 👍

Joshuabaker2:
Unused styles removed & commits squashed!

binoy14:
@Joshuabaker2 Your awesome! Thanks for the PR

alien3d:
I'm just checking here for the same thing..  How do i use it ?  I have manually changed the listitem in node module..  I write like this  "          <ListItem title="Username"  textInput={true}  textInputKeyboardType="default"></ListItem>" seem nothing happen

binoy14:
@Joshuabaker2 any thoughts on this?

Joshuabaker2:
Sure thing! It's just like the [React Native TextInput](https://facebook.github.io/react-native/docs/textinput.html), in that you need to set the value while utilizing a callback. In the React Native TextInput props, it is using an `onChangeText` callback and the `value` to set it (check out the docs) and for ours, since we are wrapping this component with a lot more stuff, the names were changed to `textInputOnChangeText` and `textInputValue` respectively. 

As a really basic example, it would look like: 

```
...
      <ListItem
        textInputOnChangeText={(text) => this.setState({text})}
        textInputValue ={this.state.text}
      />
...
```

Monte9:
@Joshuabaker2 It would be helpful for everyone if you added this example to the readme.

alien3d:
Agree .. I would love to see good example  basic login..  @Monte9  ..

alien3d:
@Joshuabaker2 How do you solve focus issue? since refs cannot be access?
[test file](https://pastebin.com/xUjt1fd7)
List item change
[List item](https://pastebin.com/sxT8pBTXl)
@Monte9  still incomplete example :(

alien3d:
Even thou close.. i'm still working on real example list view text input and now it's working. It may not the same with current file..
[test file](https://pastebin.com/dWZDVLdM)
[listitem](https://pastebin.com/d8UkLSVB)

A full sample login will be complete next weekend  :) .. login /register using react element with  php /c#

Monte9:
@alien3d awesome! Keep us updated and kindly post your solution back here. Thanks

marksbren:
I cannot get the TextInput to display on the list item. The following code does not work for me. I set the this.state.newMax value to "20" to start but there is nothing showing on the right and the keyboard does not activate when a tap there.

```
<ListItem
	key={1}
	title="Maximum Value"
	hideChevron={true}
	textInput={true}
	textInputMaxLength={3}
	textInputKeyboardType='numeric'
	textInputValue={this.state.newMax}
	textInputSecureTextEntry={false}
	textInputOnChangeText={(text) => this.setState({newMax:text})}
/>
```
iphone 6 ios 10 3 14e269 2017-04-28 08-04-29

alien3d:
Dear  @marksbren,  textInputSecureTextEntry i think not in the current path yet 
[Current file](https://raw.githubusercontent.com/react-native-training/react-native-elements/master/src/list/ListItem.js)

What my latest sample even not complete yet still can be used..

You may check out this before i release full functional with back end ..
[test file](https://gist.github.com/alien3d/7a2ff42647958ecee8e77085a99e6c21)
[listitem.js](https://gist.github.com/alien3d/1d3100ca43e643a8e8ab83a636415370)

PR #265, opened 2017-03-31
Simple, temporary fix for android

Draer:
On android devices the input width has some problems and always sticks to the right edge.

https://cloud.githubusercontent.com/assets/26611935/24225107/fe30c7ee-0f5e-11e7-8a74-cf61755b8ad0.png

This fix just makes it to have 100% width of parent's component instead of android's device width.

### DIFF
--- a/src/form/FormInput.js
+++ b/src/form/FormInput.js
@@ -122,12 +122,13 @@ styles = StyleSheet.create({
     ...Platform.select({
       android: {
         height: 46,
+        width: '100%',
       },
       ios: {
+        width: width,
         height: 36,
       }
     }),
-    width: width,
     color: colors.grey3,
     fontSize: normalize(14)
   }


binoy14:
@Draer thanks for the pull request! My only concern with your solution is that percent sizing was introduced in recent version of react native. This would break for anyone using older version of react native. Can we use flex somehow? (Haven't looked at the full code yet)

Draer:
Oh! I was not aware of that, as I am pretty new to RN. I will look into it tomorrow :)

binoy14:
@Draer No worries dude.

src/form/FormInput.js
```diff
@@ -122,12 +122,13 @@ styles = StyleSheet.create({
     ...Platform.select({
       android: {
         height: 46,
+        width: '100%',
```
binoy14: @Draer Instead of this change this to `width: width - 25` you can also move `width: width` before the ...Platform.select so you don't have to define in ios

binoy14:
@Draer Please see my comment and you can use https://snack.expo.io/ByvCHV10g as a reference fix. Let us know if you need any help. We want to gets this merge soon :)

binoy14:
Closing this due to inactivity

PR #266, opened 2017-04-01
Documentaion setup using mkdocs

binoy14:
First of all big thanks to @bozzmob for setting everything up using mkdocs. 

I have testing using codeship and everything works. @Monte9 You will need to configure codeship with the following commands before you merge it and everything should just work. We will use the same configurations for testing as well. 

## Setup Commands

```sh
git fetch --unshallow || true
git fetch origin "+refs/heads/*:refs/remotes/origin/*"
pip install mkdocs mkdocs-material
npm install --global yarn
yarn
```

## Test Commands
```sh
yarn test
echo "Deplying Docs"
rev=$(git rev-parse --short HEAD)
mkdocs gh-deploy --clean --force --message "[ci skip] Docs for ${rev}"
```


src/text/Text.js
```diff
@@ -25,21 +25,15 @@ const TextElement = ({style, children, h1, h2, h3, h4, h5, h6, fontFamily}) => (
...
     })
   },
   bold: {
     ...Platform.select({
-      ios: {
-        fontFamily: fonts.ios.bold
-      },
       android: {
-        fontFamily: fonts.android.bold
+        ...fonts.android.bold
```
bozzmob: Its more like a doubt than a comment. How does this work? I mean how does the spread operator replace `fontFamily`?
```
-        fontFamily: fonts.android.bold
+        ...fonts.android.bold
```

I've used spread operator in different places, was curious to know how this worked.

binoy14: @bozzmob since fonts is a object that has fontFamily when it spreads it replaces. See the changes in config/fonts

PR #268, opened 2017-04-02
Fixing Link for create react native

prashanthmadi:
Link doesn't work without https:// in front


binoy14:
@prashanthmadi Thanks for the PR

Issue #245, opened 2017-03-23
Why is Avatar Component using a different Icon component?


joncursi:
https://github.com/react-native-community/react-native-elements/blob/master/src/avatar/Avatar.js#L9

This library has it's own `<Icon />` component which brings support for half a dozen icon libraries. Why is `<Avatar />` not using it, and instead using a font-awesome specific icon component from an external lib?

The API docs are written with *this* library's `<Icon />` component in mind:

<img width="812" alt="screen shot 2017-03-23 at 1 57 32 pm" src="https://cloud.githubusercontent.com/assets/4974609/24262549/c43f7202-0fd0-11e7-88c8-c2e448417b7c.png">

So right now, the implementation does not match the API docs.

See also: https://github.com/react-native-community/react-native-elements/issues/244

Monte9:
@joncursi Good catch. I was the one that worked on the Avatar component and this is clearly a mistake. 

Will push a fix for it later this evening. Keep an eye out for it.

Monte9:
PR #244 has been merged. This fix should be out in our next release `v0.11.0`

Issue #244, opened 2017-03-23
Icon bug in Avatar component


joncursi:
Bug: https://github.com/react-native-community/react-native-elements/blob/master/src/avatar/Avatar.js#L98

`icon.type` is being fed into the `name` prop on icon, which is incorrect. It should be fed into the `type` prop:

```jsx
<Icon
   ...
  name={icon.name || 'user'}
  name={icon.type || 'material'}
  ...
/>
```

darleikroth:
Yes. The documentation is completely different about this point.
[Avatar Props](https://github.com/react-native-community/react-native-elements#avatar-props)

`prop icon -> type ->
object {name: string, color: string, size: number, type: string (default is material, or choose one of material-community, simple-line-icon, zocial, font-awesome, octicon, ionicon, foundation, evilicon, or entypo), iconStyle: object(style)}`

Monte9:
Will review this as part of the fix for #245.

binoy14:
Closing this in favor of #245

Monte9:
PR #269 has been merged!

PR #269, opened 2017-04-02
Fix for avatar name and type

binoy14:
In reference to #244 #245


Monte9:
@binoy14 this was a quick fix. Thanks for doing it! 

Lgtm. Merging it now.

Issue #274, opened 2017-04-03
Card titleStyle proptype incorrect


Joshuabaker2:
The titleStyle is for a text component, but has a View proptype currently. Making a fix.

Monte9:
PR merged. Thanks for reporting the issue @Joshuabaker2!

PR #275, opened 2017-04-03
Fix Card titleStyle to use Text Proptype instead of View Proptype

Joshuabaker2:
Addresses #274


Monte9:
lgtm! Thanks for the quick fix @Joshuabaker2!

Issue #239, opened 2017-03-22
FormInput width (Android only)


szakhlypa:
By default TextInput receives width equals to window width, but it's HOC View has marginHorizontal equals to 15. That creates issue on android:
![screenshot](https://cloud.githubusercontent.com/assets/26611935/24225107/fe30c7ee-0f5e-11e7-8a74-cf61755b8ad0.png)
Also HOC View margins can be redefined using containerStyle prop.
Seems that this issue connected with #70. Maybe @binoy14 should take a look at it.
Thanks for great package

binoy14:
@szakhlypa hmm makes sense. If you think you have solved the issue feel free to send a PR and I can review it.

binoy14:
@szakhlypa This should be fixed in latest release. Let me know if its still an issue

Monte9:
@binoy14 can you point me to the PR that fixes this issue? Thanks.

binoy14:
@Monte9 looks like it was closed not merged. My bad, I will send a PR soon

mwarger:
I was able to work around this by explicitly setting `marginRight: 0` on the component style.  Any update on the PR?

binoy14:
I will fix this in the next release. I had fixed it but some reason i didn't create a PR and it got lost. I will fix it again

Monte9:
This has been fixed and will be released in `v0.12.3`

DenisDov:
`width: '100%'`

Issue #236, opened 2017-03-21
ADD Toasts


SheppardX:
I use your Elements and now i slide in some user interactions and wanted to use toasts for successful actions... but there are no toasts :(

binoy14:
@SheppardX thanks for reporting. We are currently swamped with feature requests and this won't be on the immediate roadmap. That being said check out @dabit3 's article on [Creating A Custom Toast Module for React Native](https://medium.com/@dabit3/creating-a-custom-toast-module-for-react-native-770fd1c0dcf5) and feel free to send a PR.

Monte9:
We are looking to implement "custom toasts" soon. Since this issue closely resembles it, I am closing this. Follow along [here]( #190 )

Issue #190, opened 2017-02-20
[Feature] Add notification component


nebula-ai:
Hi,

I wonder if anyone has a recommendation of a library to use with react native elements to provide a on screen notification generated by a backend server.  This is outside of the push notifications build into the OS.  At the simplest this is an Alert that goes away after a few seconds.   I am using reactnavigation as well.  

Thanks in advance

corymsmith:
@nebula-ai In a couple apps I just created my own, it was just a component that was in my main application component, when a notification came in it just slides down from the top, stays for a few seconds then goes away. Something like that would be a good addition to this lib for sure.

nebula-ai:
Thanks for suggestion.  I'll give it a try and publish if I get it to work cleanly

binoy14:
@corymsmith @nebula-ai What are your thoughts on having something like [this](https://github.com/devBrian/react-native-dropdownalert) in this library?

![](https://camo.githubusercontent.com/f6913fda265b52d81f731079425ecb012dca4cbe/68747470733a2f2f7261772e6769746875622e636f6d2f646576427269616e2f72656163742d6e61746976652d64726f70646f776e616c6572742f6d61737465722f73637265656e73686f74732f64656d6f2e676966)

corymsmith:
@binoy14 Thats pretty much what I was referring to, as long as its highly customizable, background color, text color, font, icon, etc. The right X icon is a bit too bold in my opinion, I think it should be more subtle or just remove it since you can swipe the alert out of the way OR it just goes away on its own after a few seconds.

binoy14:
Yes, customization is the key. I'll add it to the roadmap. Thanks for the suggestion.

Monte9:
In an effort to streamline all RNE issues such that it is easier to get to solutions faster/ submit a new issue we have decided to close any enhancement/roadmap features as "issues". 

Feel free to use this issue to talk about this feature and ping @binoy14 or me (@monte9) if you feel like you can submit a PR for this. Thanks in advance.

Also you can track the status of this feature and all other roadmap features [here](https://github.com/react-native-community/react-native-elements/issues/195#issuecomment-286213704).

tarang9211:
@nebula-ai Opening this issue again for easy accessibility.

camillo777:
But will this notification show up on a mobile phone even if the app is minimized or closed?

binoy14:
[...]

PR #276, opened 2017-04-03
make toast for ios and android and change width text input

sutani:
#239 change width
#236 add toast

please review my PR


Monte9:
@sutani Thanks for the PR. I'll review it soon. 

Can you also please attach screenshots of what the component looks like? Thanks.

sutani:
ok @Monte9 
i have attach screenshots

Monte9:
Looks good. This is a great start and I can see Toast being a great addition to React Native Elements. 

With that being said, I think the UI and Component API for Toast can still be improved. 
At RNE the 🔑  for all components is Customization + Polish + Easy of use. So consider flushing the component to handle several props for containerStyling etc. Look at `Button.js` for examples.

Also I wonder what you think of this? https://github.com/react-native-training/react-native-elements/issues/190
Do you think, this could be a good use case for Toasts with gestures?

sutani:
yes, I think it is much better, I made it once so that the base can be used on ios and android, so I'm going to discover again for toast, using message notification bar like #190 @Monte9

package.json
```diff
@@ -23,6 +23,9 @@
   },
   "homepage": "https://github.com/dabit3/React-Native-Elements#readme",
   "dependencies": {
+    "i": "^0.3.5",
```
binoy14: What's the I and npm library needed for?

src/form/FormInput.js
```diff
@@ -127,7 +127,7 @@ styles = StyleSheet.create({
         height: 36,
       }
     }),
-    width: width,
+    width: width * 0.9,
```
binoy14: What's the reason for this change? It doesn't seem to relate to Toasts?

Monte9: I believe this is addressing Issue #239. 
I would prefer to have this in a different PR. Also look at [PR 265](https://github.com/react-native-training/react-native-elements/pull/265/files) as this is an attempt to fix the same issue.

Monte9:
I think we will have to move away from this implementation for `Toast`. Here's what we have in mind: https://camo.githubusercontent.com/f6913fda265b52d81f731079425ecb012dca4cbe/68747470733a2f2f7261772e6769746875622e636f6d2f646576427269616e2f72656163742d6e61746976652d64726f70646f776e616c6572742f6d61737465722f73637265656e73686f74732f64656d6f2e676966

I think adding a custom toast component could be more useful and it would help to build it from the ground-up. I don't see the need to depend on an npm package for this. 

@sutani Thoughts? 

/cc @binoy14

binoy14:
Agreed, we want to implement a custom Toast component. Let's try to brain storm some design ideas. The functionality of the Toast component can be a base for the notification component as well.

binoy14:
@sutani Closing this PR because we want to implement a custom component and not use a third part library. Thanks for taking time and effort. We really appreciate your hard work you put into this.

Issue #198, opened 2017-02-25
Which files need to be hosted?


sufuf3:
Hi @react-native-community , 
I'm a member of [cdnjs](https://github.com/cdnjs/cdnjs). We are going to host this library.

What the files should need to be added in cdnjs?
(I guess the files under `src/` need to be added in cdnjs.)

And could you provide the minified file, because I can't minify the files with [uglify-js](https://www.npmjs.com/package/uglify-js).

If you have any suggestions, please let me know.

Thank you.

[cdnjs/cdnjs#10418](https://github.com/cdnjs/cdnjs/pull/10418)

Monte9:
@sufuf3 That's great! Yes, I believe the files under `src/` should do. 

We'll follow up on the minified files /soon.

Monte9:
@sufuf3 I looked into minifying files but I am not sure what the best approach to do so is. 

What problem did you run into while minifying it with uglify-js? And can you suggest an alternative? Also I am not too experienced in this regard, so if you could elaborate a bit with regard to the process that would be great.

Also @binoy14 and I help maintain this library together. So I am cc'ing him on this issue so either of us can follow-up on this.

binoy14:
I got minification working with webpack but I am not sure what use case you're looking to serve. As far as I know you can't require files from CDN in react maybe.

sufuf3:
@binoy14 @Monte9 We host the lib with `.js` and `min.js`.
CDNJS is the front-end development resources. And we also host [react](https://cdnjs.com/libraries/react).
So if this library provide the minified file, it would be great.

binoy14:
@sufuf3 Take a look at this file https://github.com/react-native-training/react-native-elements/blob/master/build/bundle.js does that work?

sufuf3:
@binoy14 It's awesome. Thank you for your great work.

binoy14:
Your welcome, Closing this issue. Let me know if you need anything else

PR #283, opened 2017-04-06
Minification using webpack

binoy14:
Reference to #198


binoy14:
@Monte9 or @bozzmob merge when ready

PR #295, opened 2017-04-12
Avatar - default iconSize and titleSize to half of width

mguida22:
Currently when setting a width and/or height for an Avatar, the iconSize and titleSize don't update unless they are explicitly set.

This allows a user to set their own width and height for an avatar, and have the iconSize and titleSize adjust accordingly.

### DIFF
--- a/src/avatar/Avatar.js
+++ b/src/avatar/Avatar.js
@@ -36,39 +36,29 @@ const Avatar = (props) => {
     height,
   } = props;
 
-  let titleSize = 17;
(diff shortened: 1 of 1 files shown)


binoy14:
@mguida22 Could you please rebase and fix conflicts so we can review and merge. Thanks for taking the time to create a pull request

mguida22:
@binoy14 just rebased!

binoy14:
@mguida22 Lgtm. Thanks for the PR!

Issue #47, opened 2016-09-25
Header element


gavindinubilo:
I'm trying to make a view that has a header at the top, similar to what [react-native-navbar](https://github.com/react-native-community/react-native-navbar) is doing, I understand that it may not be great to copy what they are doing, but it would be great to have a unified style out of the box.

Monte9:
@gavindinubilo that's a good idea. Although in my experience, if you are using any navigation library, NavigationExperimental for instance, the header comes with it as well. 

Were you thinking of using this "unified style" throughout your app that's independent of any sort of navigation? Just trying to understand how useful it would be in general.

Monte9:
Ran into this while prototyping a simple UI for an app. I didn't have any sort of navigation although I needed a header and had to write one myself. I think it could be pretty useful if added as part of RN Elements. Will try to implement this soon.

gavindinubilo:
I ran into it when building a fairly simple app, all I needed was a header for a view, it didn't have any navigation components to it. Though having that option would be great.

Carpetfizz:
This seems pretty essential for ListView based apps. How does the Scene change when a row is tapped?

binoy14:
@Monte9 What are your thoughts on having functionality like http://shoutem.github.io/docs/ui-toolkit/components/navigation-bar There is a wide variety of options available

gavindinubilo:
@binoy14 Something like that would be awesome!

Monte9:
@binoy14 I think that's a great idea. Let me know if you have some time on your hands to start it off.. I can pitch in as well.

Monte9:
In an effort to streamline all RNE issues such that it is easier to get to solutions faster/ submit a new issue we have decided to close any enhancement/roadmap features as "issues". 

Feel free to use this issue to talk about this feature and ping @binoy14 or me (@monte9) if you feel like you can submit a PR for this. Thanks in advance.

Also you can track the status of this feature and all other roadmap features [here](https://github.com/react-native-community/react-native-elements/issues/195#issuecomment-286213704).

filiphosko:
@binoy14 @Monte9 I'm working on this, mostly inspired by NavigationBar (shoutem/ui) and Header (nativebase) components. I could have it done by the end of next week (I hope).

binoy14:
@filiphosko go for it! Let me know if you need any help.

filiphosko:
[...]

PR #286, opened 2017-04-09
Improve coverage

cpenarrieta:
The purpose of this PR is to improve the code coverage of the test cases. There is still some room to improve coverage for the `Slider` component but I think its a good start.
image

PR #285, opened 2017-04-09
Refactor props to use ES6 spread

binoy14:
The advantage of doing this would be that all parent components can
be given native props for that component. This is also useful in case
any of react native props change/added/deleted we do not have to do
manual update of all components using that props.

PR #299, opened 2017-04-13
Feature: Header Component

filiphosko:
This is the first version of header/navbar element. I need feedback about what to change/improve. There are still things missing (tests, proptypes validation), I will add them later after reviews and tweaks. This PR targets the issue https://github.com/react-native-training/react-native-elements/issues/47.


Monte9:
@filiphosko This is a great start! Thanks for owning this roadmap feature. 💯 

Here's a list of all the things that need to be addressed:
1. Add screenshots for all possible combinations of props. This helps improve the component from a visual standpoint as we can ensure we are making it as customizable as possible with clean UI. 
2. Refactor to deconstruct props within the class/object. Look at this PR #285 to see what I am talking about. 
3. Add 1 simple test case. Just a simple passing test is a good start too. Look at this PR if you need a reference #286 
4. Add barebones docs (look in the `docs/` folder)

filiphosko:
I refactored the component to support passing in components through props and also added some simple (and superficial) tests. I don't know why the Codecov checks are not passing tho (all tests were passing locally).

I will add some documentation and screenshots next.

filiphosko:
I'm finishing the docs, I have to make some screenshots and push the changes here.

binoy14:
@filiphosko This looks amazing! Will review it soon. Thanks for the PR

filiphosko:
Thanks, I will probably revert the 'unification' of styling that I did with the second last commit (basically there is a top offset by -20 on Android which pushes the header under the StatusBar and makes it look like on iOS). Now I don't see it as a good idea. Apart from that everything else was meant to be ;) I'm open to improvements.

src/header/Header.js
```diff
@@ -0,0 +1,129 @@
+import React from 'react';
+import { StatusBar, StyleSheet, Text, View } from 'react-native';
+import isEmpty from 'lodash.isempty';
+import Icon from '../icons/Icon';
+
+export const DummyNavButton = () => (
```
binoy14: Is this temporary? P.s I like the name

src/header/Header.js
```diff
@@ -0,0 +1,129 @@
...
+  childrenArray.push(
+    leftComponent,
+    centerComponent,
+    rightComponent,
+  );
+
+  return childrenArray;
+}
+
+const Header = (props) => {
```
binoy14: I am thinking let's split the components in seperate files. That way we can maintain it easily as well as test them easily. What do you think?

src/header/Header.js
```diff
@@ -0,0 +1,129 @@
+import React from 'react';
+import { StatusBar, StyleSheet, Text, View } from 'react-native';
+import isEmpty from 'lodash.isempty';
+import Icon from '../icons/Icon';
+
+export const DummyNavButton = () => (
+  <View style={styles.dummyNavBtn} />
+);
+
+export const Title = (props) => {
```
binoy14: This is just a text component right? Why export it?

src/header/Header.js
```diff
@@ -0,0 +1,129 @@
...
+  const {
+    text,
+  ...attributes,
+  } = props;
+
+  return (
+    <Text
```
binoy14: Let's use text from react native elements as it handles font size based on pixel ratio

src/header/Header.js
```diff
@@ -0,0 +1,129 @@
...
+      numberOfLines={1}
+      {...attributes}
+    >
+      {text}
+    </Text>
+  );
+};
+
+export const NavButton = (props) => {
```
binoy14: Again it's just the icon component why export it?

src/header/Header.js
Monte9: can we rename to `DefaultNavButton` if that makes sense in this context? 
I'll leave it up to you to the decide.

src/header/Header.js
```diff
@@ -0,0 +1,129 @@
...
+      <StatusBar {...statusBarProps} />
+      <View style={[styles.innerContainer, innerContainerStyles]}>
+        {propChildren.length > 0 ? propChildren : children}
+      </View>
+    </View>
+  );
+};
+
+const styles = StyleSheet.create({
+  dummyNavBtn: {
```
Monte9: rename to `defaultNavButton`?

src/header/__tests__/Header.js
```diff
@@ -0,0 +1,170 @@
...
+
+describe('Title Component', () => {
+  it('should render without issues', () => {
+    const component = shallow(<Title />);
+
+    expect(component.length).toBe(1);
+    expect(toJson(component)).toMatchSnapshot();
+  });
+
+  it('should accept props', () => {
```
Monte9: tests are great. 💯 gj!

src/header/Header.js
filiphosko: I exported it only to be able to test it (to have 100% test coverage).

src/header/Header.js
filiphosko: Ok, I'll swap it 👍

src/header/Header.js
filiphosko: To be able to test it, would I be able to do it otherwise?

src/header/Header.js
filiphosko: Sure, we can. In my first interation I had them split in separate files, then I changed my mind since they were very small and I could avoid unnecessary imports. Sometimes I try to balance it (like when components are really tiny), but maybe it would be better also in this case to have them separate.

src/header/Header.js
filiphosko: Yep, we can, but semantically it's just a placeholder and not even a button, that's why I named it 'dummy'. But we can rename it for sure.

src/header/__tests__/Header.js
filiphosko: Thanks, I tried to keep codecov satisfied.

binoy14:
@filiphosko Is this ready to be merged? I personally feel that we should split small components in separate files. Seems like you agree too.

cc @Monte9

filiphosko:
@binoy14 Yep, I agree, but sometimes I hesitate to do it since I try to avoid unnecessary import where I can. If I would split them, 1) they would need to be imported into Header.js, 2) each of them will need to import React

filiphosko:
Sorry, I didn't have time to do it, so hopefully it's ok now.

binoy14:
Thanks @filiphosko

PR #310, opened 2017-04-16
feat(ListItem): Allow a react component to be passed for icon

### LINKED ISSUE #301: Allow passing a component to rightIcon/leftIcon in ListItem
### Description

1. Explain what you did
- I'm using the ListItem component but want to able to use custom icons for my `leftIcon`/`rightIcon` chevrons

2. What you expected to happen
- I thought it would allow passing a React component

3. What actually happens
- ListItem only allows an object with type and/or name properties to define which icon to use amongst the one provided by react-native-vector-icons

### Also please include the version of `react-native-elements` that you are using.
All versions

### Solution
Modify the implementation of rightIcon and leftIcon to allow an object or a react component. Update the prop types to expect an object or a component. This shouldn't be breaking change as it would still default to expecting an object. If this is something that you would like to consider, I can send a PR for it.

karanjthakkar:
Fixes #301 

// @Monte9 @binoy14

### DIFF
--- a/docs/API/lists.md
+++ b/docs/API/lists.md
@@ -166,8 +166,8 @@ styles = StyleSheet.create({
 | component | View or TouchableHighlight if onPress method is added as prop | React Native element | replace element with custom element (optional) |
 | containerStyle | none | object (style) | additional main container styling (optional) |
 | hideChevron | false | boolean | set if you do not want a chevron (optional) |
-| leftIcon | none | object {name, color, style, type} (type defaults to material icons) | icon configuration for left icon (optional) |
-| rightIcon | {name: 'chevron-right'} | object {name, color, style, type} (type defaults to material icons) | icon configuration for right icon (optional). Shows up unless hideChevron is set |
+| leftIcon | none | object {name, color, style, type} (type defaults to material icons) <br/><br/> **OR** <br/><br/> React Native element | icon configuration for left icon (optional) |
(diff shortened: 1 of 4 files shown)


JulianKingman:
@karanjthakkar Any chance you could add this for other components as well (e.g. checkbox, etc...)?

binoy14:
@karanjthakkar Could you please add tests in the ListItem.js file in here https://github.com/karanjthakkar/react-native-elements/blob/e57f3ccb82d63702f0015d0a09b07e007d8a2420/src/list/__tests__/ListItem.js and it would be nice to add it to the docs as well

Monte9:
This would be awesome to have on the `v0.11.2` release going out in a day or so. 

@karanjthakkar do you think you could get those changes that @binoy14 requested in soon?

karanjthakkar:
@Monte9 I'm traveling right now but I can get those changes done by tomorrow once I'm back. Do you also want me to add it to other components like @JulianKingman suggested?

Monte9:
In that case, let's not rush it. We can merge this PR in the upcming release `v0.11.3` going out in a week or so. Take your time. 

Yes, let's also add this to the other components were applicable.

karanjthakkar:
@Monte9 Have added the tests that @binoy14 asked for. I might need some more time to check and add the different components where icons are used. I think we should create an issue for this and merge this PR. Reason being, since you're actively accepting PR's and pushing newer versions, this PR might get stale soon if I'm not able to pick it up as soon as I want to. What do you think?

binoy14:
@karanjthakkar Agree, if a component is done then lets ship it and revisit other components in a different PR

PR #313, opened 2017-04-16
Avatar: uses default Icon type if none is given

### LINKED ISSUE #304: [Avatar] icon -> type prop should default to 'material', not 'font-awesome'
Every other component that uses `<Icon />`, [including the `<Icon />` component itself](https://react-native-training.github.io/react-native-elements/API/icons/), uses `material` as the default icon library if no `type` is specified. Avatar should be no different.

See: https://github.com/react-native-training/react-native-elements/blob/master/src/avatar/Avatar.js#L100

mguida22:
fixes #304


Monte9:
Thanks for the quick fix @mguida22

PR #316, opened 2017-04-17
Fix android overflow: ButtonGroup

AudyOdi:
ButtonGroup: when the first or last button is selected (has a background color), the overflow on Android is broken, causing square corners. This problem does not affect iOS.

Before:

After:

This fix allows a border radius on the button which can work-around the issue.

A better solution would be to fix overflow in Android in react-native core. There is an open issue for that but no ongoing work at this time (as per the documentation).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-native-elements",
-  "version": "0.11.1",
+  "version": "0.11.0",
   "description": "React Native Elements & UI Toolkit",
   "main": "src/index.js",
   "repository": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-native-elements",
-  "version": "0.11.0",
+  "version": "0.11.1",
   "description": "React Native Elements & UI Toolkit",
   "main": "src/index.js",
   "repository": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-native-elements",
-  "version": "0.11.0",
+  "version": "0.11.1",
```
Monte9: shouldn't need to change this. Let's revert/discard this change.

AudyOdi:
@Monte9 You're welcome. Glad I can contribute to this project

Monte9:
I should have brought this up earlier, but `containerBorderRadius` needs to be added to the docs. 

@bozzmob can you make sure it's added before it get's merged into master? Thanks.

MrHubble:
@Monte9 @bozzmob Thanks for the React Native Elements! It's been extremely helpful with getting started with React Native.

I believe `containerBorderRadius` hasn't been added to the docs. Any chance this could be done? I tried using it like:

 `containerBorderRadius={{borderBottomLeftRadius: 0, borderTopLeftRadius: 0}}`

but that threw an error so I assume I'm doing it wrong.

Monte9:
That's good to know @MrHubble! Welcome to the community. 😄 

Yes, you are right. It should have been added to the docs as part fo this PR. Sorry about that. 

if you look at the code `containerBorderRadius` has a `propType` of `number`:
```containerBorderRadius: PropTypes.number,```

So the right way to use it would be to do:
`containerBorderRadius={10}`

Let me know if this works for you.

PR #317, opened 2017-04-17
enhancement: added onPress event for badge component

tarang9211:


src/badge/badge.js
```diff
@@ -10,11 +10,28 @@ const Badge = props => {
 
   if (badge.element) return badge.element;
```
binoy14: instead of doing this, lets use prop de-structuring similar to those in other components for examples

```js
const {
  component,
  containerStyle,
  textStyle,
  title,
  ...attributes
} = props;

return (
  <Component style={[styles.badge, containerStyle && containerStyle]} {...attributes}>
      <Text style={[styles.text, textStyle && textStyle]}>{title}</Text>
  </Component>
)
```

src/badge/badge.js
```diff
@@ -10,11 +10,28 @@ const Badge = props => {
...
+
   return (
-    <View style={[ styles.badge, badge.badgeContainerStyle ]}>
+    <Component style={[ styles.badge, badge.badgeContainerStyle ]} onPress={badge.onPress}>
       <Text style={[ styles.text, badge.badgeTextStyle ]}>{badge.value}</Text>
-    </View>
+    </Component>
   );
+
+  // if (badge.onPress) {
```
binoy14: Please remove this comments for the final iteration :)

binoy14:
@tarang9211 Thanks for the PR! Please checkout my comment

tarang9211:
@binoy14 I made a few changes as per your comment, let me know what you think of it.

src/badge/badge.js
```diff
@@ -1,20 +1,31 @@
 import React from 'react';
-import { Text, View, StyleSheet } from 'react-native';
+import { Text, View, StyleSheet, TouchableOpacity } from 'react-native';
 
 let styles = {};
 
 const Badge = props => {
   const { badge } = props;
```
Monte9: you are deconstructing from props below too. Just add this there

src/badge/badge.js
```diff
@@ -1,20 +1,31 @@
 import React from 'react';
-import { Text, View, StyleSheet } from 'react-native';
+import { Text, View, StyleSheet, TouchableOpacity } from 'react-native';
 
 let styles = {};
```
Monte9: I think we moved away from this. @binoy14 I remember you refactored this as part of the props PR right?

src/badge/badge.js
```diff
@@ -6,32 +6,26 @@ let styles = {};
...
 
+  const {
+    badgeContainerStyle,
+    badgeTextStyle,
+    onPress,
+    element,
+    value
+  } = props;
+
   if (!badge) throw Error('badge prop is required');
```
binoy14: This line is not needed anymore

src/badge/badge.js
```diff
@@ -1,20 +1,31 @@
...
 
-  if (badge.element) return badge.element;
+  if (element) return badge.element;
```
binoy14: This is weird, basically we are letting user pass a whole element and returning that. If they really need to do that and not utilize badge styles then why use badge in the first place. I don't think this is needed @monte9 thoughts?

src/badge/badge.js
```diff
@@ -1,20 +1,31 @@
...
+
+  const Component = onPress !== undefined ? TouchableOpacity : View;
 
   return (
-    <View style={[ styles.badge, badge.badgeContainerStyle ]}>
-      <Text style={[ styles.text, badge.badgeTextStyle ]}>{badge.value}</Text>
-    </View>
+    <Component style={[ styles.badge, badgeContainerStyle ]} onPress={onPress}>
```
binoy14: You should pass `{...attributes}` this let's user pass all available props for any component they use.

src/badge/badge.js
```diff
@@ -1,20 +1,31 @@
...
 
 const Badge = props => {
   const { badge } = props;
 
+  const {
+    badgeContainerStyle,
+    badgeTextStyle,
+    onPress,
+    element,
+    value
```
binoy14: Add `...attributes` as the last element which basically adds rest of the pros in attributes ​object

src/badge/badge.js
tarang9211: Sure, but I've never undersood what `{...attributes}` does. If we allow more props than listed in the react propTypes validation, won't it throw a warning/error?

tarang9211:
@binoy14 made a few of the changes. These are the props that need validation

- `badgeContainerStyle` type object?
- `badgeTextStyle` type object?
- `element` (not sure if we need this based on your comment above)
- `value` type string?

src/badge/badge.js
tarang9211: I think this line will get removed with this refactor.

src/badge/badge.js
tarang9211: I'm not sure I understand this. I would still need to import `TouchableOpacity` no?

tarang9211:
@binoy14 @Monte9 please some comments whenever you get a chance, looking to close this out soon.

src/badge/badge.js
binoy14: @tarang9211 I think @Monte9 comment means that we use const styles before the component, no need to define this style variable

src/badge/badge.js
```diff
@@ -1,24 +1,36 @@
...
+import { Text, View, StyleSheet, TouchableOpacity } from 'react-native';
 
 let styles = {};
 
 const Badge = props => {
   const { badge } = props;
 
-  if (!badge) throw Error('badge prop is required');
+  const {
+    badgeContainerStyle,
```
binoy14: lets rename them from `badgeContainerStyle` => `containerStyle` and `badgeTextStyle` => `textStyle`

src/badge/badge.js
```diff
@@ -1,24 +1,36 @@
...
+    badgeTextStyle,
+    onPress,
+    element,
+    value,
+    ...attributes
+  } = props;
 
-  if (badge.element) return badge.element;
+  if (element) return badge.element;
```
binoy14: You don't need this line

src/badge/badge.js
```diff
@@ -1,24 +1,36 @@
...
 
 const Badge = props => {
   const { badge } = props;
 
-  if (!badge) throw Error('badge prop is required');
+  const {
+    badgeContainerStyle,
+    badgeTextStyle,
+    onPress,
+    element,
```
binoy14: don't need the element

src/badge/badge.js
binoy14: Yes, please remove this

src/badge/badge.js
```diff
@@ -1,24 +1,36 @@
...
+    <Component style={[ styles.badge, badgeContainerStyle ]} onPress={onPress} {...attributes}>
+      <Text style={[ styles.text, badgeTextStyle ]}>{value}</Text>
+    </Component>
   );
+
 };
 
 Badge.propTypes = {
   badge: React.PropTypes.any,
+
```
binoy14: containerStyle will have prop types of `View.propTypes.style`
textStyle will have prop types of `Text.propTypes.style`
value should have prop types of `PropTypes.oneOf([PropTypes.string, PropTypes.number, PropTypes.element])

@monte9 does PropTypes.element take care of anyone passing custom text component?

Monte9:
@tarang9211 I apologise for this in advanced. It seems like we need a label/last comment on issue that there's a PR for already. 

Kindly look at #330. It is addressing the same enhancement as this PR. Let me know if you would be ok with closing this PR in favor of that. 

I hope the feedback you got on this PR has helped.

binoy14:
@tarang9211 Don't feel dis hearted I combined your code with #330 Thanks for working on this!

PR #337, opened 2017-04-25
add a props(autoCapitalize) for searchbar

shixiaoquan:
I found a problem when I use react-native-elements in my project .
When I use the component searchbar, the first letter in the textinput is automatically capitalized.


Monte9:
@shixiaoquan thanks for the PR. I think you can already do this with the existing component. There is the spread operator that will handle this for you. ` {...attributes}`

Just pass in and it should work. 
```
<TextInput
  autoCapitalize={autoCapitalize} 
  ...
/>
```

Let me know if it doesn't work.

shixiaoquan:
@Monte9  I have do it by modifying the source code of SearchBar 
`<TextInput
  autoCapitalize="none"
  ...
/>`,
I would like to set the searchbar by setting the property, will be more convenient

binoy14:
Closing this since this seems to already work. @shixiaoquan let me know if you think other wise. I could be missing something

PR #339, opened 2017-04-25
added test for search clearInput, closes #234

### LINKED ISSUE #234: problem on SearchBar with clear button
Hello guys,

I am having a problem with clear button on the SearchBar.
In the SearchBar the clear button is clearing the text but is not triggering the **onChangeText** event...

ericwooley:
The response on the react issue was to change the code to make the behavior work. https://github.com/facebook/react-native/issues/13251

This calls onChangeText when the clearButton is pressed, as expected.

I also added unit tests for this.

Closes https://github.com/react-native-training/react-native-elements/issues/234

### DIFF
--- a/src/input/Search.js
+++ b/src/input/Search.js
@@ -18,8 +18,18 @@ class Search extends Component {
   }
 
   clearText() {
-    const ref = this.props.textInputRef;
-    this.refs[ref].clear();
+    if (this.props.onChangeText) {
+      this.props.onChangeText('');
+    }
+    try {
+      const ref = this.props.textInputRef;
+      this.refs[ref].clear();
+    } catch (e) {
+      if (__DEV__)
+        console.warn(
+          'Could not access textInput reference, make sure you supplied the textInputRef'
+        );
+    }
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


src/input/Search.js
```diff
@@ -18,8 +18,18 @@ class Search extends Component {
   }
 
   clearText() {
-    const ref = this.props.textInputRef;
-    this.refs[ref].clear();
+    if (this.props.onChangeText) {
+      this.props.onChangeText('');
```
binoy14: If i am thinking about this right, it's this just selling value to an empty string?

binoy14:
@ericwooley Thanks for the PR, from what it looks to me, it looks like a work around to the actual bug in react native. Is that correct? if not can you provide a sketch link so can confirm it is not related to our project. You can use https://snack.expo.io/BJDhWjJJ- as a starting point

ericwooley:
As for the sketch, unfortunately I just started vacation until May 5. Hopefully someone else can pick up my slack on this issue. 

Good luck!

binoy14:
@ericwooley I think the workaround makes sense to add, was just curious if it's an actual bug or not. I will review it soon and get it merged. Thanks for the PR, enjoy your vacation 🏖

src/input/Search.js
bartimaeus: I need onChangeText called after text is cleared so tha my onChangeText function can reload my data. I might just not show my clearIcon until this is merged.

neo125874: many many thanks like this workaround!!

PR #342, opened 2017-04-26
imageComponent prop for Card element

tarang9211:


binoy14:
For image especially someone can either pass in to use a different image component or the full component itself, meaning either `<Card imageComponent={CustomImage} />` or `<Card imageComponent={<MyCustomImage />} />` and we need to support both. 

You will need to check if it's a component being passed or an element and you can do that simply by checking if the component is a function then it will be the first case and using `React.isValidElement` that will be for the second case. Based on this you will render the image component. Let me know if you have any more questions regarding this.

binoy14:
Closing this, reopen when ready

PR #347, opened 2017-05-02
Updated PropTypes to reference prop-types package

binoy14:


binoy14:
Lol how is this even possible ?

PR #361, opened 2017-05-09
Enable ListItem to set more props for avatar

tonimoeckel:
With this change, you are able to set props to the avatar Image
component of the ListItem component (like „defaultSource“...) 

```
<ListItem
                avatarProps={{
                    defaultSource: 'https://yourimage/path.jpeg'
                }}
.../>

### DIFF
--- a/src/list/ListItem.js
+++ b/src/list/ListItem.js
@@ -24,6 +24,7 @@ const ListItem = props => {
     rightIcon,
     leftIconContainerStyle,
     avatarStyle,
+    avatarProps,
     underlayColor,
     subtitle,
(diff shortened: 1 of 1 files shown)


binoy14:
@tonimoeckel just so you know we are changing the avatar in the list item to use the avatar component we recently released. The avatar component supports passing pros to set an image. I really appreciate you sending a PR but this code will go away when the avatar component is used.

tonimoeckel:
@binoy14 thank you! Makes perfect sense to me. I'll close my PR and hope to get that changes soon 👍

PR #362, opened 2017-05-10
Make sure the focus and blur method always works

Nickman87:
Currently, when you do not specify a ref string (which is discouraged anyway!), calling focus just crashes for no obvious reason.
This adds a fallback to store the ref trough a callback method to avoid this issue.


src/form/FormInput.js
```diff
@@ -14,13 +14,22 @@ import normalize from '../helpers/normalizeText';
...
 class FormInput extends Component {
+  constructor(props) {
+    super(props);
+    
+    this.getRef = this.getRef.bind(this);
+    this.focus = this.focus.bind(this);
+    this.blur = this.blur.bind(this);
+  }
+  
+  getRef() {
```
binoy14: use `getRef = () => {}` so you won't have to bind and you can also remove the contructor

src/form/FormInput.js
```diff
@@ -14,13 +14,22 @@ import normalize from '../helpers/normalizeText';
...
+    return this.props.textInputRef ? this.refs[this.props.textInputRef] : this.input;
+  }
   focus() {
-    const ref = this.props.textInputRef;
-    this.refs[ref].focus();
+    getRef() && getRef().focus();
```
binoy14: not sure why do it this way. If this just returning value why not just do it in constructor? Also shouldn't it be `this.getRef()`

Nickman87:
This was a quick PR to see if you were interested in it. I'll check your remarks and check if other components might benefit from this change as well.

Thanks for the feedback!

binoy14:
@Nickman87 Any update on this?

src/form/FormInput.js
Nickman87: It's just a failsafe to make sure the ref actually exists before calling focus on it. It has to be this.getRef indeed

Nickman87:
@binoy14 I just got round to take another look at this. I applied some changes according to your review, merged in the latest master, also applied the change to the search bar and updated the docs to reflect the changes.

filiphosko:
@Nickman87 @binoy14 I'm wondering why we still need this functionality. Passing a string literal into refs is deprecated (as is using this.refs). If you'll pass a callback function into the textInputRef, you can assign it to a class variable and call focus and blur methods on it that way. In my opinion we should remove that deprecated functionality and encourage a correct usage of refs. Also I don't see a real added benefit by hooking up the blur and focus methods directly to the component, why not just to let users access them from the textInputRef? What do you think?

Nickman87:
@filiphosko I was wondering this as well as passing in a ref name and then accessing it from the parent  component just seems dirty. I did not want to change the API of the components to keep from breaking existing code.

If this is not a problem we can tweak this further.
I feel that exposing the methods as they are now is cleaner compared to passing in a ref callback and leaving the rest up to the parent component. For now, a TextInput component is used, and this component happens to have the focus method, but what if the implementation were to change?

Keeping the methods focus, blur and clear on the FormInput/Search component, as they are now, provides a stable API for the users of the react-native-elements components while the implemantation can change without issues.

filiphosko:
@Nickman87 I would personally use the blur/focus methods now like this:

```jsx
class Form extends React.Component {
  inputs;

  componentDidMount() {
    this.inputs.name.focus(); // I would use autoFocus normally, this is just for demonstration
  }  

  render() {
    return(
      <View>
        <FormInput
          textInpuRef={(c) => { this.inputs.name = c; }}
          value="My name"
        />
        <FormInput
          textInpuRef={(c) => { this.inputs.email = c; }}
          value="my@email.com"
        />
      </View>
    );
  }  
}
```
Basically now we can use 'ref' and 'textInputRef' on the FormInput to do the same. What I meant is just that the API can be confusing to someone (and definitely the way it was written in the docs). But of course the way you did is good from the standpoint of backwards compatibility (not messing up the option of passing a string to refs).

Nickman87:
I Understand @filiphosko but now you are exposing the internal workings of the FormInput component. In my opinion it is better to only allow external ref usage and leave that usage up to the user and expose the focus method on the formInput itself instead of exposing the inner TestInput.

binoy14:
I agree with @filiphosko We should not use using strings as refs. I wasn't aware they are deprecated. Now there are a questions we need to figure out, which I don't know the answer to.

1. Is using functions as refs supported in older versions of RN? If not which version specifically starts supporting it?

We do need to get rid of string ref as in the new version it will break but it is important for us to answer the above question before we do so.

Now in terms of this PR. I think lets wait on answering the above questions before we put more work in to this.

Nickman87:
@binoy14 I don't think they have yet announced a version in which they want to get rid of the string refs, but they are indeed already marked as depricated in the documentation.
Using function callbacks in refs has been there since 0.13.0 it seems: https://github.com/facebook/react/releases/tag/v0.13.0 So I'm guessing it is safe to switch to function callbacks entirely?

binoy14:
Thanks @Nickman87 so would you mind refactoring your PR to use functional refs. I am not sure if updating this PR is easier or creating a new one is. Let me know what you think after seeing the conflicts

Nickman87:
@binoy14 Sure, I'll take a look at it on wednesday probably

Nickman87:
I pushed a new commit, refactoring the props to allow for strings and functions.
I also removed reference to the props from the docs to discourage their usage (https://github.com/react-native-training/react-native-elements/issues/392) and added additional info on how to use/access the methods.

I would love to write some tests for this, but was unable to figure out how to do this. Any ideas?
Everything I found about enzyme and jest suggest we need full rendering to test this out.

Let me know what you think

filiphosko:
@Nickman87 Great work on this. I would love to at least help you with the tests, but I'm totally without any spare time these days.

filiphosko:
'Solving' one of the open issues confirmed me in my theory that having so many ref props as we have is really confusing to users. In my opinion we could override the default ref on the component and use it only for TextInput. What do you think?

Monte9:
@filiphosko @Nickman87 what's the status of this PR? Could it be a part of the upcoming release `v0.14.0`?

Nickman87:
@Monte9 I believe these changes can go live but I would rather have some tests in to verify everything is working as intended...

src/form/FormInput.js
```diff
@@ -51,8 +73,10 @@ FormInput.propTypes = {
   containerStyle: View.propTypes.style,
   inputStyle: NativeText.propTypes.style,
   selectionColor: PropTypes.string,
-  textInputRef: PropTypes.string,
-  containerRef: PropTypes.string,
+  // Deprecated
+  textInputRef: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
+  // Deprecated
+  containerRef: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
```
Monte9: @iRoachie can we add deprecate warning for these two props. Thanks.

src/input/Search.js
```diff
@@ -99,8 +119,10 @@ Search.propTypes = {
   showLoadingIcon: PropTypes.bool,
   loadingIcon: PropTypes.object,
   clearIcon: PropTypes.oneOfType([PropTypes.object, PropTypes.bool]),
-  textInputRef: PropTypes.string,
-  containerRef: PropTypes.string,
+  // Deprecated
+  textInputRef: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
+  // Deprecated
+  containerRef: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
```
Monte9: add deprecate warnings. /cc @iRoachie

docs/API/searchbar.md
```diff
@@ -36,14 +36,34 @@ import { SearchBar } from 'react-native-elements'
...
+  ref={search => this.search = search}
+  ...
+/>
+```
+You can then access SearchBar methods like so:
+
+```
+this.search.focus();
+this.search.blur();
+this.search.clearText();
```
Monte9: the docs are solid. 💯 thanks

Monte9:
@Nickman87 I reimplemted Input component for `v1` and the way you handled the refs in this PR are right. [Here's](https://github.com/react-native-training/react-native-elements/blob/v1/example/v1/input/Input.js#L51) the implemented it on `v1` for reference. 

This PR lgtm and can be merged into the `next` branch. /cc @iRoachie

PR #365, opened 2017-05-11
Add custom icon component into Button component

tamdao:


build/bundle.js
```diff
@@ -1 +1,4671 @@
...
\ No newline at end of file
+!(function(e) {
```
binoy14: @tamdao Whats the reason to change this file? We will be using travis to build this.

binoy14:
@tamdao Can you please update the test. You can run `npm run test -- -u` and it would be good to update the docs on how to use this as well as add a new test for it

build/bundle.js
tamdao: Sorry. I don't know about that. So Do I need to reset it?

tamdao:
@binoy14 Yeah. I will add the test and update document for that thia weekend

binoy14:
@tamdao Thanks for the PR! 🙌

PR #367, opened 2017-05-12
add secureTextEntry prop for sensitive text

paynegreen:
This is to obscure any sensitive input in the textInput field such as passwords. Using the ListItem elements for a custom account page, I found secureTextEntry prop missing to be able to use it for passwords without implementing the component myself.


Monte9:
Thanks for taking the time to submit a PR @paynegreen! 👍

PR #368, opened 2017-05-13
Add onClearText props so it is able to ejecute a callback when the text i…

delvallejonatan:
…s clear

### DIFF
--- a/src/input/Search.js
+++ b/src/input/Search.js
@@ -21,6 +21,7 @@ class Search extends Component {
   clearText() {
     const ref = this.props.textInputRef;
     this.refs[ref].clear();
+    this.props.onClearText && this.props.onClearText();
   }
 
   render() {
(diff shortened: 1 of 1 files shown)


binoy14:
@delvallejonatan Take a look at the latest change which was merged. It calls onChangeText due to bug in react native. https://github.com/react-native-training/react-native-elements/blob/next/src/input/Search.js#L23 Is this PR still relevant ?

delvallejonatan:
I think this may work, thanks. How can I get this version ?

binoy14:
We will be release new version later this evening. It will be part of that fixes. If you want to experiment, check out the `next` branch

Silvercast:
I think having an extra props `onClearText` makes sense. How can we know if user deletes all characters, or he removes the last character, or he taps on the clear button?

Some other conversation here https://github.com/react-native-training/react-native-elements/issues/389

PR #372, opened 2017-05-14
Added /example expo to RNE core

Monte9:
Added expo app which demo's all the elements in RNE core. 

https://expo.io/@monte9/react-native-elements-app


binoy14:
Lgtm

Issue #350, opened 2017-05-02
Make rounded Button (like large Social Icon)


camillo777:
Make a rounded button (like large social icon) same size so that I can stack social icons and custom buttons of the same aspect, style and size.

binoy14:
@camillo777 you should be able to use props in the button dedicated for styling to achieve this. I would be open to adding that as an available prop. Wondering what @Monte9 thinks

filiphosko:
@binoy14 I agree with @camillo777, it would be nice to have these common button designs configurable through props, for example like [this](https://docs.nativebase.io/Components.html#button-outline-headref). We could add rounded, outline and transparent button options. This is something I have on my todo list for a while, so maybe I could look at it.

binoy14:
@filiphosko go for it

Monte9:
Thanks @filiphosko for the PR! 💯 

The PR has been merged and will be released in `v0.12.3`

PR #382, opened 2017-05-17
Add button styles (transparent, outline, rounded)

filiphosko:
This PR addresses https://github.com/react-native-training/react-native-elements/issues/350 and adds two more styles. Also for convenience while prototyping I opted for having a stylesObject which I'm using to access nested style properties (like stylesObject.text.fontSize) rather than moving properties into separate keys in a StyleSheet (like styles.textFontSize). I don't know if you'll be ok with this solution.

buttons

### DIFF
--- a/src/buttons/Button.js
+++ b/src/buttons/Button.js
@@ -48,6 +48,9 @@ const Button = props => {
     disabledStyle,
     fontFamily,
     containerViewStyle,
+    rounded,
+    outline,
+    transparent,
     ...attributes
   } = props;
   let { Component } = props;
@@ -102,6 +105,14 @@ const Button = props => {
     );
   }
 
+  const baseFont = {
+    color: (textStyle && textStyle.color) || color || stylesObject.text.color,
+    size: (textStyle && textStyle.fontSize) ||
+      fontSize ||
+      (!large && stylesObject.smallFont.fontSize) ||
+      stylesObject.text.fontSize,
+  };
+
   return (
     <View
       style={[styles.container, raised && styles.raised, containerViewStyle]}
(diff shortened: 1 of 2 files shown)


src/buttons/Button.js
```diff
@@ -123,6 +134,21 @@ const Button = props => {
             backgroundColor && { backgroundColor: backgroundColor },
             borderRadius && { borderRadius },
             !large && styles.small,
+            rounded && {
```
binoy14: @filiphosko instead of passing object here can you move this inside stylesheet.create.

src/buttons/Button.js
```diff
@@ -181,7 +207,7 @@ Button.propTypes = {
   fontFamily: PropTypes.string,
 };
 
-const styles = StyleSheet.create({
+const stylesObject = {
```
binoy14: Trying to understand, what's the point of doing this?

filiphosko: The point was to access nested properties, as I mentioned in the PR description - I needed to calculate dimensions based on font size and I also pick correct color based on what the user specified, this is then passed to baseFont const. With StyleSheet you can't access nested properties (like styles.text.fontSize), the only way is to have styles.textFontSize and to me that can add up, but when you use an actual object, you can access them. That's why I wrote in the description that probably you won't agree with that, but it helped me while prototyping. It was just a proposition, we can revert to having only StyleSheet and using keys like styles.textFontSize for this.

binoy14: @filiphosko Totally overlooked the description 😄 I will take a look at in more detail and let u know

Monte9:
@binoy14 @filiphosko have this changes been added to the docs? 

Any PR that contains users facing component changes should always contain doc updates on them as well.

filiphosko:
@Monte9 Yes, you are right. I didn't add them into docs since I waited for review and merging this caught me a little off guard ;) I'll add it.

binoy14:
@filiphosko @Monte9 My bad, I should have checked as well

JLLLinn:
Hello, may I ask when are these going to master? Or is there any timeline that we are looking at to use the "next" version with documentation?

Thanks so much for the work!

Monte9:
@JLLLinn these changes are already in master and it should be available to use with `v0.15.0`.

Here's the changes on [master](https://github.com/react-native-training/react-native-elements/blob/master/src/buttons/Button.js#L52)

Unfortunately, these changes still need to be added to the docs. If you would like to submit a PR for it, I'd be more than happy to merge it in. ⚡️

PR #384, opened 2017-05-17
Change View.propTypes to ViewPropTypes to support React Native v44

alxvallejo:
In response to https://github.com/react-community/react-navigation/issues/1352


binoy14:
@alxvallejo What would happen if someone is using react native version older than v44? Won't that cause errors?

raarts:
@binoy14: yes that would surely give errors. But at least you're not alone, many packages experience this problem...

binoy14:
Closing this as this soultion would not work. Thanks for the PR!

PR #386, opened 2017-05-18
fix(socialIcon): fontStyle prototype to Text

madhutv:
I get a warning message on socialIcon component when using fontStyle prop.
screen shot 2017-05-18 at 7 25 02 pm
Wondering if prototype for fontStyle should be Text instead of View. 

Submitting a PR with this change, can you please review.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-native-elements",
-  "version": "0.12.1",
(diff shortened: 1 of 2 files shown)


binoy14:
@madhutv looks like missed this one. Thanks for the PR

PR #396, opened 2017-05-27
Update image refs to use HTTPS

ryanguest:


binoy14:
@ryanguest This still doesn't solve the issue of gif not showing up. I believe this is because imgur uses gifv format which is not support by markdown. I am gonna close this, thanks for the PR!

Monte9:
@binoy14 what is a possible solution to this issue? I noticed it was broken, but didn't really dive too deep into why it was doing so.

binoy14:
imgur uses gifv which is a video format, which is not supported by github readme. I have added gif to images folder in docs and this would have been fixed

PR #403, opened 2017-06-03
Enhancements to the Rating component

### LINKED ISSUE #388: Rating value
How to set rating value?

leonfeng:
1. A `readonly` prop that disables user interactions
2. Ability to set an initial rating value
3. Speed up the rendering of rating images by using native drivers
4. Ability to set the number of decimal places for rating values

Closes #388 

Please feel free to let me know if any further changes are required. Thanks for reviewing my work :)

### DIFF
--- a/docs/API/rating.md
+++ b/docs/API/rating.md
@@ -42,16 +42,40 @@ const WATER_IMAGE = require('./water.png')
 />
 ```
 
+##### Read-only mode
+```js
+const { rating } = this.props;
+
+<Rating
+  imageSize={20}
+  readonly
+  startingValue={rating}
+  style={{ styles.rating }}
+/>
+```
+
+![readonly demo jpg](https://i.imgur.com/LSVE8sN.jpg)
+
+##### Fractions
+```html
+<Rating showRating fractions={1} startingValue={3.3} />
+```
+
+![Fractions demo gif](https://cloud.githubusercontent.com/assets/241553/26780040/e8cd1a2c-49f8-11e7-8859-6dd9b4e0a779.gif)
+
 #### Rating Props
 
-| prop | default | type | description |
-| ---- | ---- | ----| ---- |
-| **onFinishRating** | none | function | Callback method when the user finishes rating. Gives you the final rating value as a whole number **(required)** |
-| type | star | string | Choose one of the built-in types: `star`, `rocket`, `bell`, `heart` or use type `custom` to render a custom image (optional) |
-| ratingImage | star | string | Pass in a custom image source; use this along with `type='custom'` prop above (optional) |
-| ratingColor | #f1c40f | string (color) | Pass in a custom fill-color for the rating icon; use this along with `type='custom'` prop above (optional) |
-| ratingBackgroundColor | white | string (color) | Pass in a custom background-fill-color for the rating icon; use this along with `type='custom'` prop above (optional) |
-| ratingCount | 5 | number | The number of rating images to display (optional) |
-| imageSize | 50 | number | The size of each rating image (optional) |
-| showRating | none | boolean | Displays the Built-in Rating UI to show the rating value in real-time (optional) |
-| style | none | function | Exposes style prop to add additonal styling to the container view (optional) |
+| prop                  | default         | type           | description                                                                                                                  |
+| ----                  | ----            | ----           | ----                                                                                                                         |
+| **onFinishRating**    | none            | function       | Callback method when the user finishes rating. Gives you the final rating value as a whole number **(required)**             |
+| type                  | star            | string         | Choose one of the built-in types: `star`, `rocket`, `bell`, `heart` or use type `custom` to render a custom image (optional) |
+| ratingImage           | star            | string         | Pass in a custom image source; use this along with `type='custom'` prop above (optional)                                     |
+| ratingColor           | #f1c40f         | string (color) | Pass in a custom fill-color for the rating icon; use this along with `type='custom'` prop above (optional)                   |
+| ratingBackgroundColor | white           | string (color) | Pass in a custom background-fill-color for the rating icon; use this along with `type='custom'` prop above (optional)        |
+| ratingCount           | 5               | number         | The number of rating images to display (optional)                                                                            |
+| imageSize             | 50              | number         | The size of each rating image (optional)                                                                                     |
+| showRating            | none            | boolean        | Displays the Built-in Rating UI to show the rating value in real-time (optional)                                             |
+| readonly              | false           | boolean        | Whether the rating can be modiefied by the user (optional)                                                                   |
+| startingValue         | ratingCount / 2 | number         | The initial rating to render (optional)                                                                                      |
+| fractions             | undefined       | number         | The number of decimal places for the rating value; must be between 0 and 20 (optional)                                       |
+| style                 | none            | function       | Exposes style prop to add additonal styling to the container view (optional)                                                 |
--- a/src/rating/Rating.js
+++ b/src/rating/Rating.js
@@ -56,24 +56,32 @@ export default class Rating extends Component {
   constructor(props) {
     super(props);
 
-    const { onFinishRating } = this.props;
+    const { onFinishRating, fractions } = this.props;
 
     const position = new Animated.ValueXY();
-    const newValue = new Animated.ValueXY();
-    newValue.setValue({ x: 0, y: 500 });
 
     const panResponder = PanResponder.create({
       onStartShouldSetPanResponder: () => true,
       onPanResponderMove: (event, gesture) => {
-        position.setValue({ x: gesture.dx, y: gesture.dy });
-        this.setState({ value: gesture.dx });
+        const newPosition = new Animated.ValueXY();
+        newPosition.setValue({ x: gesture.dx, y: 0 });
+        this.setState({ position: newPosition, value: gesture.dx });
       },
       onPanResponderRelease: event => {
-        onFinishRating(this.getCurrentRating());
+        const rating = this.getCurrentRating();
+        if (!fractions) {
+          // "round up" to the nearest star/rocket/whatever
+          this.setCurrentRating(rating);
+        }
+        onFinishRating(rating);
       },
     });
 
     this.state = { panResponder, position };
+  }
+
+  componentDidMount() {
+    this.setCurrentRating(this.props.startingValue);
   }
 
   getPrimaryViewStyle() {
@@ -82,15 +90,18 @@ export default class Rating extends Component {
 
     const color = TYPES[type].color;
 
-    const width = position.x.interpolate({
-      inputRange: [
-        -ratingCount * (imageSize / 2),
-        0,
-        ratingCount * (imageSize / 2),
-      ],
-      outputRange: [0, ratingCount * imageSize / 2, ratingCount * imageSize],
-      extrapolate: 'clamp',
-    });
+    const width = position.x.interpolate(
+      {
+        inputRange: [
+          -ratingCount * (imageSize / 2),
+          0,
+          ratingCount * (imageSize / 2),
+        ],
+        outputRange: [0, ratingCount * imageSize / 2, ratingCount * imageSize],
+        extrapolate: 'clamp',
+      },
+      { useNativeDriver: true }
+    );
 
     return {
       backgroundColor: color,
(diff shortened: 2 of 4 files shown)


src/rating/Rating.js
```diff
@@ -69,28 +69,40 @@ export default class Rating extends Component {
         this.setState({ value: gesture.dx });
       },
       onPanResponderRelease: event => {
-        onFinishRating(this.getCurrentRating());
+        const rating = this.getCurrentRating();
+        if (!fractions) {
+          // "round up" to the nearest star/rocket/whatever
+          this.setCurrentRating(rating);
```
leonfeng: I've also tweaked the image rendering so that if `fractions` is 0 or undefined, individual images will "fill up" to match the rounding up of rating values. Setting `fractions` to a value between 0 and 20 disables this behaviour.

src/rating/Rating.js
```diff
@@ -56,11 +56,11 @@ export default class Rating extends Component {
...
     super(props);
 
-    const { onFinishRating } = this.props;
+    const { onFinishRating, fractions } = this.props;
 
     const position = new Animated.ValueXY();
-    const newValue = new Animated.ValueXY();
-    newValue.setValue({ x: 0, y: 500 });
+    // const newValue = new Animated.ValueXY();
+    // newValue.setValue({ x: 0, y: 500 });
```
leonfeng: These two lines don't seem to be doing anything… so I commented them out.

binoy14:
@leonfeng Thanks for the PR!
@Monte9 Should probably review this.

src/rating/Rating.js
Monte9: Yup, you are right. These can be removed.

src/rating/Rating.js
```diff
@@ -69,28 +69,40 @@ export default class Rating extends Component {
...
+      {
+        inputRange: [
+          -ratingCount * (imageSize / 2),
+          0,
+          ratingCount * (imageSize / 2),
+        ],
+        outputRange: [0, ratingCount * imageSize / 2, ratingCount * imageSize],
+        extrapolate: 'clamp',
+      },
+      { useNativeDriver: true }
```
Monte9: Was meaning to do this. Thanks for adding it in.

src/rating/Rating.js
```diff
@@ -138,25 +153,57 @@ export default class Rating extends Component {
...
     let currentRating = 0;
 
     if (value > ratingCount * imageSize / 2) {
       currentRating = ratingCount;
     } else if (value > imageSize) {
-      currentRating = Math.ceil(startingValue + value / imageSize);
+      currentRating = startingValue + value / imageSize;
+      currentRating = !fractions
+        ? Math.ceil(currentRating)
+        : +currentRating.toFixed(fractions);
```
Monte9: never used the toFixed func before. How does it work?
Say I pass in a value of `fractions={30}`

src/rating/Rating.js
```diff
@@ -138,25 +153,57 @@ export default class Rating extends Component {
...
 
     return currentRating;
   }
 
+  setCurrentRating(rating) {
+    const { imageSize, ratingCount } = this.props;
+    // `initialRating` corresponds to `startingValue` in the getter. Naming it
+    // differently here avoids confusion with `value` below.
+    const initialRating = ratingCount / 2;
+    let value = -ratingCount * imageSize / 2;
```
Monte9: the value assigned to the value variable is not being used. Can we refactor to `let value = nil`

src/rating/Rating.js
```diff
@@ -138,25 +153,57 @@ export default class Rating extends Component {
...
+    } else if (rating < 0) {
+      value = -ratingCount * imageSize / 2;
+    } else if (rating < ratingCount / 2) {
+      value = (rating - initialRating) * imageSize;
+    } else {
+      value = 0;
+    }
+
+    const { position } = this.state;
+    position.setValue({ x: value, y: 0 });
```
Monte9: why do we need to reset the value of `rating` when we set a rating? 

Once the rating is rendered, I would think that the UI for the user interaction should begin from that point and not 0. 

Please correct me if this is not the resulting implementation.

src/rating/Rating.js
```diff
@@ -245,6 +293,23 @@ const styles = StyleSheet.create({
...
 });
 
+const fractionsType = (props, propName, componentName) => {
+  if (props[propName]) {
+    const value = props[propName];
+    if (typeof value === 'number') {
+      return value >= 0 && value <= 20
+        ? null
+        : new Error(
+            `\`${propName}\` in \`${componentName}\` must be between 0 and 20`
```
Monte9: this is a great addition to validate if the fractions is between 0 and 20. 

Although seems like an overkill to me. Is this absolutely necessary? Can't the user get away with not passing a `fractions` prop if they don't need fractions?

src/rating/Rating.js
```diff
@@ -255,4 +320,7 @@ Rating.propTypes = {
   onFinishRating: PropTypes.func,
   showRating: PropTypes.bool,
   style: View.propTypes.style,
+  readonly: PropTypes.bool,
+  startingValue: PropTypes.number,
+  fractions: fractionsType,
```
Monte9: This is great! The `readonly` and `startingValue` props have been on my to-do list. Glad you got to it, and the implementation looks pretty clean too. 👏

Monte9: Also please can you add these props to the docs? Thanks.

src/rating/Rating.js
Monte9: > individual images will "fill up" to match the rounding up of rating values.

Can you explain this in a  bit more detail? Adding a couple screenshots for a few fractions would be great.

Monte9:
This is a 💯 PR! I appreciate your help on making the Rating component better. 😄 

Couple things:
1. Maybe this is asking too much but can you record a short`gif` to show the `startingValue` prop working? Just want to make sure it's exactly how I would imagine it worked. 
2. Please add the 3 additional props to the docs. This is the file to do that: https://github.com/react-native-training/react-native-elements/blob/master/docs/API/rating.md

src/rating/Rating.js
leonfeng: For instance, `<Rating showRating />`:
* Without the `if` block, the 4th star is only partially filled even though the rating is at 4:
before

I'm probably nitpicking here, but this small discrepancy may still confuse some users.

* With the `if` block, the 4th star is now entirely filled on pan release:
after

And here is `<Rating showRating fractions={1} startingValue={3.3} />`:
fractions

src/rating/Rating.js
leonfeng: Simply put, `toFixed` returns a string representation of the number rounded to the specified decimal places. With `fractions={13}` and a rating of `2.5`, you get `2.5000000000000`.

I wanted a way to display Facebook ratings, which contain 1 decimal place. But I saw no reason to limit the number of decimal places supported by `toFixed`.

src/rating/Rating.js
leonfeng: Sure thing.

src/rating/Rating.js
leonfeng: I tried just `this.setState({ value });`, and the images were not rendered correctly:
img_0330

src/rating/Rating.js
leonfeng: Yeah, I'm probably overthinking it. This is for (accidental) invalid `fractions` values, like `-1` or `true`. The red screen of death only says you're using `toFixed` wrong, which I imagine will leave some users scratching their heads. This way you also get a message in the terminal:
screen shot 2017-06-05 at 15 09 06

Not passing `fractions` works as expected :)

src/rating/Rating.js
binoy14: Just a blind shot at this, but this might be because of how setState batches changes. You might be able to leverage a functional setState. Eg.

```js
this.setState(prevState => ({
  x: prevState.x + 1
})
```

binoy14:
@leonfeng @monte9 Also let's update the tests with the new props so we are not dropping code coverage by 3% :D

leonfeng:
@binoy14 Oh, I was wondering about that… Well, I'm still at caveman level when it comes to tests. So please be patient while I lookup some tutorials…

binoy14:
@leonfeng ya no worries, let me know if you need any help.

src/rating/Rating.js
leonfeng: I think I got it. Here are lines 202 and 203 modified:
```js
    const newPosition = new Animated.ValueXY();
    newPosition.setValue({ x: value, y: 0 });
    this.setState({ position: newPosition });
```
The only catch is I have to change the pan responder move handler to do the same, otherwise it won't work 🙄

src/rating/Rating.js
Monte9: Allowing the user to display up to 20 decimal places seems a bit much, don't you think? I think we should limit it to max 2. 

> I wanted a way to display Facebook ratings

Ratings on Facebook? hmm.. curious to know more about the use case.

src/rating/Rating.js
```diff
@@ -138,25 +151,57 @@ export default class Rating extends Component {
...
     }
 
     return currentRating;
   }
 
+  setCurrentRating(rating) {
+    const { imageSize, ratingCount } = this.props;
+    // `initialRating` corresponds to `startingValue` in the getter. Naming it
+    // differently here avoids confusion with `value` below.
+    const initialRating = ratingCount / 2;
```
Monte9: also I think the default `initialRating` should be set to 0

src/rating/Rating.js
Monte9: Right. Both seems like acceptable solutions, I like the second one personally.

docs/API/rating.md
```diff
@@ -44,14 +44,17 @@ const WATER_IMAGE = require('./water.png')
...
+| imageSize             | 50              | number         | The size of each rating image (optional)                                                                                     |
+| showRating            | none            | boolean        | Displays the Built-in Rating UI to show the rating value in real-time (optional)                                             |
+| readonly              | false           | boolean        | Whether the rating can be modiefied by the user (optional)                                                                   |
+| startingValue         | ratingCount / 2 | number         | The initial rating to render (optional)                                                                                      |
+| fractions             | undefined       | number         | The number of decimal places for the rating value; must be between 0 and 20 (optional)                                       |
+| style                 | none            | function       | Exposes style prop to add additonal styling to the container view (optional)                                                 |
```
Monte9: looks good. 

if you have time, can you add a couple sample use cases to show using the `showRating` and `readonly` props.. as well as using `fractions`.

If not, don't worry.. I can add some myself once I merge this in.

src/rating/Rating.js
Monte9: > I'm probably nitpicking here, but this small discrepancy may still confuse some users.

I love it. It's the minor details and polish that matters the most. 💯 

Also your gif editing skills are 🥇. Would be awesome to have these on the docs as well to show how fractions works.

src/rating/Rating.js
```diff
@@ -138,25 +151,57 @@ export default class Rating extends Component {
...
+    if (rating > ratingCount) {
+      value = ratingCount * imageSize / 2;
+    } else if (rating > ratingCount / 2) {
+      value = (rating - initialRating) * imageSize;
+    } else if (rating < 0) {
+      value = -ratingCount * imageSize / 2;
+    } else if (rating < ratingCount / 2) {
+      value = (rating - initialRating) * imageSize;
+    } else {
+      value = 0;
```
Monte9: also this entire function is really confusing to me.

For starters, this like should be:
`const initialRating = rating  || ratingCount / 2`
and then the rest of the logic should be refactored around it. 

But if this works, then we don't need to worry about it.

src/rating/Rating.js
Monte9: Great. I like it.

Monte9:
@leonfeng couple followup questions:
1. What are your thoughts on providing tap functionality alongside gestures. So basically, if a user taps on 4 stars, it will fill it up or they could drag from 0-> 4 to fill it up too. 
I think we can easily achive this with the `setCurrentRating` func that you wrote. Let me know what your thoughts are on this.
2. I am curious to know how you found out about Ratings.. and why you decided to send a PR to enhance it. Do you need it for your own project? Do you plan to use it any prod apps?

src/rating/Rating.js
leonfeng: Thanks. Will do :)

src/rating/Rating.js
leonfeng: Sounds fine to me. Can't think of a use case that needs more than 2 decimal places anyway.

src/rating/Rating.js
leonfeng: OK.

src/rating/Rating.js
leonfeng: I'll see what I can do. I basically "mirrored" the `getCurrentRating` function and didn't put much thought into refactoring…

leonfeng:
@Monte9 
1. I was actually trying to do that! But everything broke when I added in a `<TouchableHighlight>`. Will try again.

2. I found out about React Native Elements from a course I took on Udemy. And yes, I'm using it in an app I'm currently making. That's why I submitted this PR.

Also, apologies if I seem slow to respond. Been having network issues in the past 2 days. I'll push the changes as soon as I can.

leonfeng:
So I started to refactor `setCurrentRating`, and then realised if I changed one method I'd have to rework the whole class… Because pretty much every method has the "origin" set in the middle.

Sorry for leaving the job half done. Once I meet my deadline I should have more time to contribute.

Monte9:
@leonfeng yeah, that's my bad. I could have written it better to start with. haha

Don't be sorry, you have added a ton of functionality to `Ratings` component and it's awesome. Thanks for all your work so far. 😃 

OK, I think this PR is ready to be merged. Please let me know if you think it's ready as well and let's :shipit:

leonfeng:
@monte9 You bet 😎

PR #409, opened 2017-06-07
Fix cropped avatar on Android

xavier-villelegier:
Hey !

## Motivation
I noticed that Avatars are cropped on Android, but i can't find any issues about that.

## Solution
To solve this issue i just moved the `width` and `height` style props to the `container` instead of the `overlayContainer`.

|   |Before|After|
|---|---|---|
|Simple Avatar|screenshot_2017-06-07-11-39-42|screenshot_2017-06-07-11-49-33|
|ListItem Avatar|screenshot_2017-06-07-11-45-49|screenshot_2017-06-07-11-47-50|

Does someone know why is this issue android-specific ?
Please let me know if it breaks something that i didn't see.


Monte9:
Thanks for the quick fix @xavier-villelegier! 

Yeah, it's really strange that it's only android. iOS looks just fine. 🤔

Monte9:
also @xavier-villelegier, your app looks good. Are you using RNE in production? or plan to put it in production any time soon?

xavier-villelegier:
@Monte9 Yep, i'm working with @martinezguillaume for @optimdata, and it's planned to be in production for industrial clients. By the way, we switched from shoutem/ui to RNE, which is much more flexible for us, and the community seems to be more active.

Monte9:
That's awesome! Yeah, @martinezguillaume has submitted a few PRs too. So thanks to both of you for staying active and contributing. 

> we switched from shoutem/ui to RNE

That's great to hear! We love when someone is using RNE in beta/ prod apps. ❤️ We want to be able to provide awesome components that are not just easy to use by also fully customizable. 

So if anything doesn't fit your use case or you want to contribute back to RNE to improve existing component, please feel free to submit an issue/ PR. I'll try to make the release cycle faster.

PR #411, opened 2017-06-08
Fix SearchBar icons alignement on Android

martinezguillaume:
Yo

Just a quick fix for the SearchBar component on Android. I noticed that the clear button and the loading icon are not centered.

The loading icon was just 1 pixel far from the center (so close 😢), but the clear button was clearly not centered.

Before :
image

After : 😀
image

Here are the modifications:
image

And, I use the same style `icon` for the two icons `clear` and `search` to avoid duplication because they have the same layout configuration.

Tell me if there is anything to change, thanks ! 🌵💩

### DIFF
--- a/src/input/Search.js
+++ b/src/input/Search.js
@@ -159,7 +159,7 @@ const styles = StyleSheet.create({
     top: 13,
     ...Platform.select({
       android: {
-        top: 17,
+        top: 18,
       },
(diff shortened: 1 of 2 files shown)


Monte9:
Thanks for the fix @martinezguillaume!

💯 for adding UI polish.

PR #415, opened 2017-06-09
Fix leftIcon style in ListItem

xavier-villelegier:
Hey,

[This commit](https://github.com/react-native-training/react-native-elements/commit/467caa5fd4aa2ff6e14c66154eab21041133bec6) has broken `leftIcon` style when there is no `rightTitle`. I just quick-fixed this, but maybe the ListItem deserves a better style polish to better handle these cases.

|Before|After|
|---|---|
|capture d ecran 2017-06-09 a 11 09 19|capture d ecran 2017-06-09 a 11 09 50|


binoy14:
@xavier-villelegier Thanks for the PR. Yes your right, the listitem has become very brittle now and we were talking about cleaning it up. Feel free to send a PR if you some time.

PR #438, opened 2017-06-20
v0.14.0 Release

Monte9:
### Release notes:
https://github.com/react-native-training/react-native-elements/releases/tag/untagged-c5258711f1d5839f9efa

### Expo App:
https://expo.io/@monte9/react-native-elements-app


src/card/Card.js
```diff
@@ -0,0 +1,208 @@
+import PropTypes from 'prop-types';
```
binoy14: There is a duplicate file in the containers, if this is being moved than let's delete the container folder

Monte9:
Removed peerDependency of `react-native-vector-icons` on package.json. See [this commit](https://github.com/react-native-training/react-native-elements/pull/438/commits/681e9a344fe5f4889fb4f4931b6561ed9faf28c4)

The `example` app, for instance, uses expo and we don't explicitly need to install rn-vector-icons for an expo project.. and in such cases, it shows the following warning:

```npm WARN react-native-elements@0.14.0 requires a peer of react-native-vector-icons@^4.2.0 but none was installed.```

@binoy14 thoughts on this?

src/card/Card.js
Monte9: Done. ✅

binoy14:
@Monte9 It might be fine but it won't let the user know when they are not using CRNA or expo

PR #439, opened 2017-06-21
Refactor and allow for components to be passed in

RWOverdijk:
I also export the styles. Now I can use the default styles from the component.

```jsx
import {Card} from 'react-native-elements';
import {styles as CardStyles} from 'react-native-elements/src/containers/Card';

class MyComponent {
  renderTitle(title) {
    return <Text numberOfLines={2} style={[CardStyles.featuredSubtitle, styles.cardSubtitle]}>{title}</Text>;
  }
}
```

### DIFF
--- a/src/containers/Card.js
+++ b/src/containers/Card.js
@@ -106,7 +106,6 @@ const renderFeaturedTitle = (title, titleStyle, defaultStyle) => {
     return title;
   }
 
-  console.tron.pretty('yes', 'this is happening');
   return (
     <Text style={[defaultStyle, titleStyle && titleStyle]}>
       {title}
--- a/src/containers/Card.js
+++ b/src/containers/Card.js
@@ -1,12 +1,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import {
-  View,
-  StyleSheet,
-  Platform,
-  Image,
-  Text as NativeText,
-} from 'react-native';
+import {Image, Platform, StyleSheet, Text as NativeText, View} from 'react-native';
 import fonts from '../config/fonts';
 import colors from '../config/colors';
 import Text from '../text/Text';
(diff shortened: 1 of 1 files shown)


RWOverdijk:
Tests will follow upon approval of approach.

src/containers/Card.js
```diff
@@ -104,16 +66,63 @@ const Card = props => {
   );
 };
 
+const renderTitle = (title, image, titleStyle, fontFamily, dividerStyle) => {
+  if (!title) {
+    return null;
+  }
+
+  if (typeof title !== 'object') {
+    title = (
```
Monte9: this is confusing. So you are basically assigning an object to the prop itself?

src/containers/Card.js
```diff
@@ -74,24 +52,8 @@ const Card = props => {
...
-                  <Text
-                    style={[
-                      styles.featuredSubtitle,
-                      featuredSubtitleStyle && featuredSubtitleStyle,
-                    ]}
-                  >
-                    {featuredSubtitle}
-                  </Text>}
+                {renderFeaturedTitle(featuredTitle, featuredTitleStyle, styles.featuredTitle)}
+                {renderFeaturedTitle(featuredSubtitle, featuredSubtitleStyle, styles.featuredSubtitle)}
```
Monte9: this is a bit confusing. The function is named `renderFeaturedTitle` but you are also using it to render `featuredSubtitle`.

src/containers/Card.js
```diff
@@ -104,16 +66,63 @@ const Card = props => {
...
+      {title}
+      {!image &&
+      <Divider
+        style={[styles.divider, dividerStyle && dividerStyle]}
+      />}
+    </View>
+  );
+};
+
+const renderFeaturedTitle = (title, titleStyle, defaultStyle) => {
```
Monte9: Consider renaming this to something more generic such as `renderFeatureText` or something similar. It's doing more than just `renderFeaturedTitle`

src/containers/Card.js
```diff
@@ -122,7 +131,7 @@ Card.propTypes = {
   fontFamily: PropTypes.string,
 };
 
-const styles = StyleSheet.create({
+export const styles = StyleSheet.create({
```
Monte9: Not sure how many people will actually it especially since you need to import it like such `react-native-elements/src/containers/Card`, but if it's useful to you, I am OK with doing this.

Monte9:
@RWOverdijk PR looks good. Couple things to follow up on:
1. Please update the docs to reflect that `title`, `featuredTitle` and `featuredSubtitle` can all now we objects.
2. Add tests for the above props being an object or string

src/containers/Card.js
RWOverdijk: I'd say: featuredSubTitle is also a featured title. It's just a subtitle. Title nonetheless.

RWOverdijk:
Never mind. I'll fold my own.

PR #440, opened 2017-06-21
Add numberOfLines and ellipsizeMode to Text in Buttons

greew:
I've added the possibility to use the two following options in a button:
- textNumberOfLines
- textEllipsizeMode
My own use case is that I render buttons based on a posts title. If a
title is too long, I get a multiline button, which breaks the layout.
This way, it's possible to keep one at it one line.


Monte9:
@greew thanks for the PR. It looks good. 👍 

Couple things:
1. Please rebase against `next` as there are some extra commits on this PR
2. Please add the additional props to the docs to ensure the docs are up to date.

Ruzo:
What happened with this PR? Are these options cancelled?
Thanks.

xavier-villelegier:
@Ruzo This PR was merged and  if you check the [Button code](https://github.com/react-native-training/react-native-elements/blob/master/src/buttons/Button.js#L55) you can see that these props are still available

Ruzo:
@xavier-villelegier Ah, sorry. Thanks for letting me know.

Issue #426, opened 2017-06-15
SearchBar's clearIcon is too small. Provide prop to customize it? 


lsps9150414:
### Description
The default `clearIcon` is too small to be accurately pressed on devices with smaller screen (e.g. iPhone 5).

### Solution
Either provide a prop to customize the touchable size or simply make the touchable area bigger.

Currently the touchable area is on the icon itself not the container so the height and width of the touchable area is the same as the icon size. It would be better to make the container touchable so that we can make the touchable area bigger but have the icon remain small.

sdotson:
I'll take a crack at this.

iRoachie:
Any updates on this @sdotson. Just curious if you're already working on something.

sdotson:
@iRoachie , Feel free to go for it! I'd recommend taking a look at my closed PR for some background information.

Monte9:
@xavier-villelegier can you address this as part of your `SearcBar v1` refactor? Thanks.

PR #446, opened 2017-06-23
Add iconSize prop to SearchBar

sdotson:
Addresses #426 with a new iconSize prop.

### DIFF
--- a/src/input/Search.js
+++ b/src/input/Search.js
@@ -38,6 +38,7 @@ class Search extends Component {
       containerStyle,
       inputStyle,
       icon,
+      iconSize,
       noIcon,
       lightTheme,
       round,
@@ -77,14 +78,14 @@ class Search extends Component {
         />
         {!noIcon &&
           <Icon
-            size={16}
+            size={iconSize || 16}
             style={[styles.icon, styles.searchIcon, icon.style && icon.style]}
             name={icon.name || 'search'}
             color={icon.color || colors.grey3}
           />}
         {clearIcon &&
           <Icon
-            size={16}
+            size={iconSize || 16}
             style={[
               styles.icon,
               styles.clearIcon,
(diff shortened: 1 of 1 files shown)


filiphosko:
@sdotson Hi, it looks good but there are some things that can make it event better. In the issue you are addressing with this PR it is mentioned that maybe it would be good to provide some padding between the icon and the container so actually the touchable area could be bigger than the icon. There are several ways how to do it, since ```clearIcon``` uses the ```Icon``` component from ```react-native-vector-icons``` package directly, you could try to look at the API and create some padding, maybe by providing ```margin``` to ```iconStyle``` prop - https://github.com/oblador/react-native-vector-icons or just provide ```padding``` to ```clearIcon``` directly. Just have a look at it and maybe you'll find a better solution. The other thing I would suggest is instead of creating a new prop for the component, use the existing ```icon``` prop and since it's an object, you can add ```size``` as a property to it, I guess that way it would be cleaner.

sdotson:
Latest adds size property to icon and clearIcon. 

Looking around at the code, it looks like the size of the clickable area around an icon is tied to it's size: https://github.com/react-native-training/react-native-elements/blob/master/src/icons/Icon.js#L49-L50 .

I could play around with those proportions (e.g., `size * 2 + 8` instead of `size * 2 + 4`. What do you think about that idea? Also, apologies if this is a very beginner question, but is there an easy way of running this module locally and visually inspecting the different components in iphone or android emulator? It might be the easiest way to make the icon is still centered in the searchbar if the icon container dimensions change.

Monte9:
@sdotson thanks for PR! This is certainly a good start. 

You are right that playing with those number `size * 2 + 8` might result in a larger touch area for the icon.. but as it turns out the `Search.js` component is not using the RNE `Icon.js` but is instead directly using the RN-Vector Icons components like so: 

`import Icon from 'react-native-vector-icons/MaterialIcons';`

Hence if you are up for a challenge, start by replacing the Icon from RNVI to RNE. And then go from there. I am more than happy to help if you run into any problems. 😃

PR #448, opened 2017-06-26
Fix mix of import and module.exports

cpunion:
Build with webpack will raise `Cannot assign to read only property 'exports' of object '#<Object>' (mix require and export)`, see https://github.com/webpack/webpack/issues/4039#issuecomment-273804003.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -30,7 +30,7 @@ import Rating from './rating/Rating';
 import SwipeDeck from './swipedeck/SwipeDeck';
 import Header from './header/Header';
 
-const Elements = {
+export {
   Badge,
   Button,
   ButtonGroup,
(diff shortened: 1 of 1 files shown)


Monte9:
@binoy14 do you have thoughts on this?

Also @cpunion I am curious, why do you need to build it with webpack?

binoy14:
@cpunion Looks good, Thanks for the PR!

cpunion:
@Monte9 Sorry for late response. I used `react-native-elements` with https://github.com/taobaofed/react-web and react-navigation, run on web platform. :-)

Monte9:
@cpunion Ah that's really interesting. Curious to see what RNE on the web looks like. Would you be able to share a few screenshots?

Issue #444, opened 2017-06-22
ListItem's avatar uri not working


webraptor:
### Description
Upgraded from RNE 0.12.2 to RNE 0.13
## Expected
Expected the images used as avatar within the listItem elements to remain the same
## Actually
The images used as avatar within the listItem elements have sizing issues which destroys entire listItem's styling
Visual aids over here: https://github.com/react-native-training/react-native-elements/commit/ae9db39f745955ee09e8581fdf135bd831e9b7c8#commitcomment-22699038

### Reproduction Steps

Style and set width / height of the avatar element in a listItem , using RNE 0.12.2 
Upgrade to RNE 0.13 and see the changes.

### Issue triggered by

Commit [#7bc2e6f](https://github.com/react-native-training/react-native-elements/commit/7bc2e6f5f0874254b4c56fa27b9f61c3f699fc57)

### Solution
Since Avatar component has containerStyle, avatarStyle, overlayContainerStyle maybe it is best to expose all 3 within listItem.

roadev:
+1
Same issue here!

xavier-villelegier:
Hey @webraptor ,

I think it's more likely because of https://github.com/react-native-training/react-native-elements/commit/7bc2e6f5f0874254b4c56fa27b9f61c3f699fc57, because the `avatarStyle` is not passed to the `avatarStyle` of the Avatar component but to the `containerStyle`. (And this PR was not merged before 12.2 too.)

Does it work if you set your RNE version in your `package.json` to:
```
"react-native-elements": "git://github.com/react-native-training/react-native-elements.git#331e194"
```
(This was the commit just before the merge)

If it was working at this commit, can you give me a sample of code to clearly reproduce this issue, to be able to fix that properly ?

webraptor:
I'm out of the office. I'll be able to check it out on Monday only.

webraptor:
@xavier-villelegier Yeap, just tested with both #331e194 and #7bc2e6f and I can confirm that #7bc2e6f is the one that introduces the issue.

xavier-villelegier:
Yep that's what I was planning. However, I still think that the `ListItem` tries to handle too many things on its own, and I don't know if exposing the entire `Avatar` API through the `ListItem` is a good thing...
I'll open a PR to fix that because it's clearly a regression, thank's for your help @webraptor !

yash2code:
same issue . Is it resolved or is there any alternative ?

xavier-villelegier:
[...]

PR #449, opened 2017-06-26
Fix `ListItem` avatar styles

xavier-villelegier:
Hey,

Here is a fix for #444. For now I've exposed the `Avatar`'s style props through the `ListItem`, but as I said in #444, I don't know if it's the right solution. However, this needs to be fixed since it is a regression.

Please let me know if some changes are required !

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,8 @@ import { Button } from 'react-native-elements';
   icon={{name: 'home', size: 32}}
   buttonStyle={{backgroundColor: 'red', borderRadius: 10}}
   textStyle={{textAlign: 'center'}}
-  title={`Welcome to\nReact Native Elements`} />
+  title={`Welcome to\nReact Native Elements`}
+/>
 ```
 
 ## Components Included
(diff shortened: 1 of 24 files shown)


martinezguillaume:
This kind of stuff is very modular ! But the thing is that we need to not overstep boundary. For me, it has consistency and it didn't chock me to have these new props. And I think that it's very exceptionnal to use them, so it's a big plus.

binoy14:
@xavier-villelegier Thanks for the PR. The listitem is becoming a mess, I think we need to first stop exposing full component API's thru the listitem. If someone wants to add an avatar component I think it should be done by just passing the avatar component

```
<ListItem avatar={<Avatar {...avatarProps} />} />
```

I think this fix is fine for now till we refactor the list item. Please rebase ur branch since there seems to be extra commits.  Also update the docs please :)

xavier-villelegier:
@binoy14 Yup, I agree, that's what I was saying [here](https://github.com/react-native-training/react-native-elements/issues/444#issuecomment-311008137). I think delegating responsabilities to other components (`Avatar` in this case) is a good approach to keep things modular.
I've update the docs, but I messed up a bit with git, please tell me if my branch is not up to date

PR #455, opened 2017-06-28
fix autoresize for multiline text-inputs

A:
I found bug with autoresize multiline text-input. Here is small fix for it.

Just try to write few lines of text in this input:
```js
<FormInput multiline />
```
You can see that it isn't resizing, when RN `<TextInput />` is.


Monte9:
Thanks for the quick fix @A

This will be release in `v0.14.0` in a few days.

PR #475, opened 2017-07-09
Migrate View.propTypes to ViewPropTypes

### LINKED ISSUE #413: View.propTypes has been deprecated
I'm using 
- React Native 0.44
- React Native Elements 0.12.2 

and I'm getting the warning "View.propTypes has been deprecated and will be removed in a future version of ReactNative. Use ViewPropTypes instead." when using the Button component.

alphasp:
Fix #413


binoy14:
@alphasp this PR is awesome! Was thinking if we can create a new file with the view prop type so we can just import it instead of duplicating the conditional.

alphasp:
@binoy14 good idea, I updated the PR with new commit.

Monte9:
I'd prefer you just put the `ViewPropTypes.js` file under the `config/` folder.

alphasp:
@Monte9 done

Issue #418, opened 2017-06-11
SwipeDeck doesn't work correctly on Android


ahjorth:
### Description

1. Explain what you did
**Added Swipedeck to my application, also tried it on the react-native-elements-app on expo**
2. What you expected to happen
**I expected the top card to be swiped away and the card behind to replace it**
3. What actually happens
**A card in the stack BEHIND the first card is swiped away. Or it might be that the top card DOES get swiped, but the top card is shown further back in the stack than it's supposed to**

## More details to perhaps help debug

On the react-native-elements-app on expo, where the swipedeck has the cards of all the celebrities, the first celebrity is *Emma, 29*

When swiping, the card that moves is a card with the celebrity *Amanda, 28*, which I believe **should** be the first celebrity.

After *Amanda* is swiped away, on rerender, *Scarlett, 25* shows up

Monte9:
@ahjorth thanks for creating an issue about this. This is def a bug that we are aware of, but haven't looked into fixing it yet. 

The thing is it happens only on Android, and I suspect it has something to do with the `elevation` style property on Android. 

Do you have any ideas for how we could fix this?

Babazon:
Monte, Stephen Grider fixed this in the course where he built this component by reversing the array.

`    return Platform.OS === 'android' ? deck : deck.reverse();` in the SwipeDeck component's render return statement. 

But can you tell me how you got swiping to work on Android? Currently nothing happens when I try to swipe on Android, whereas it works smoothly on IOS.

wuyunqiang:
SwipeDeck/renderCards change code to 
 <Animated.View key={item.id} style={[styles.cardStyle, {zIndex: 0}]}>
            {this.props.renderCard(item)}
  </Animated.View>
will be OK.

Bersam:
There's another bug, when i set marginTop to swipedeck's cards, anything in top of wipedeck is untouchable. actually, if try swipe there, card will be change, it might be something with swipedeck container. I've used sample code on docs.

iRoachie:
Hey guys we deprecated the SwipeDeck component in the [last release](https://github.com/react-native-training/react-native-elements/releases/tag/v0.15.0) and now it's development is happening in another repo. I suggest you open the issue at that repo instead so it can be tracked and added in. https://github.com/Monte9/react-native-tinder-cards

flimawan:
just add a little elevation to styles.cardStyle, and the first card should be shown first

PR #477, opened 2017-07-09
adding zIndex, fix swipedeck elevation problem on android devices

Bersam:
fixes https://github.com/react-native-training/react-native-elements/issues/418

### DIFF
--- a/src/swipedeck/SwipeDeck.js
+++ b/src/swipedeck/SwipeDeck.js
@@ -118,7 +118,7 @@ export default class SwipeDeck extends Component {
         }
 
         return (
-          <Animated.View key={item.id} style={styles.cardStyle}>
+          <Animated.View key={item.id} style={[styles.cardStyle, {zIndex: 1}]}>
             {this.props.renderCard(item)}
           </Animated.View>
         );


Bersam:
I don't have access to iOS devices, so I've no idea about changes that this cause on iDevices.

wuyunqiang:
I change coe to style={[styles.cardStyle, {zIndex: 1}]} is not work.
but change to style={[styles.cardStyle, {zIndex: 0}]} will be ok on android devices.

Bersam:
I have nexus 5 with android 7.1.2, any zIndex works. but i thought zIndex 1 is more natural.
"react": "16.0.0-alpha.12",
"react-native": "0.46.1",
"react-native-elements": "^0.14.0",
"react-native-vector-icons": "^4.2.0"

Monte9:
@Bersam @wuyunqiang thanks for the fix. 1 thing you could do in the future is to fix the tests as well. All you do is run `npm test -- -u`

Also I'll update the `zIndex: 0` as that's the only case that works on both devices.

Manoj002:
{zIndex: 0} worked for me 
thanks guyss

Issue #377, opened 2017-05-16
Missing: sample for FormValidationMessage


mleduque:
The "forms" page contains code samples for FormInput and FormLabel but nothing showing how to use FormValidationMessage.
All there is is a screenshot.

For example, how do I set things up if I want to check the content is a number? How do I link the FormValidationMesage to the FormInput? How do I provide the validation function?

martinezguillaume:
The ```FormValidationMessage``` is just a styled text. You have to implement for now the logic of errors. Basically, if you have an error, display the ```FormValidationMesage```, if not, display nothing

Monte9:
@martinezguillaume thanks for pitching in. Could you by any chance provide a simple working example of what you have described above.. something that can help beginner users use as a reference to implement formValidation with the FormInput & FormLabel components?

martinezguillaume:
Yes no problem ! I put that in this issue ?

Monte9:
Sure. Let's put it here for now.. and then eventually we could move it to the docs as well.

martinezguillaume:
We can imagine this component :

```js
<View>
  <FormLabel>  
    {label}
  </FormLabel>

  <FormInput ...propsFromTextInput />

  <FormValidationMessage>
    {errorMessage}
  </FormValidationMessage>
</View>
```

When you have an error in your form, just fill the `errorMessage` with your error message.

#### Example
```js
errorMessage = null;
label = 'Email';
```
![image](https://cloud.githubusercontent.com/assets/17292331/26821863/4243dcbc-4aa8-11e7-9720-15faf9fc01df.png)

```
errorMessage = 'This field may not be blank';
label = 'Email';
```
![image](https://cloud.githubusercontent.com/assets/17292331/26821874/4fceccfc-4aa8-11e7-9194-5a60e62b45ea.png)

#### Enhancement

We can imagine a component wrapping `FormLabel`, `FormInput` and `FormValidationMessage` with the props `errorMessage`, `label` and all the props from `TextInput` from react-native. Like this :
```js
const Input = props => {
  const {
    errorMessage,
    label,
    labelStyle,
  } = props;
  return <View>
    // Display label if it's specified
    {label &&
      <FormLabel
        labelStyle={labelStyle}
      >  
        {label}
      </FormLabel>
    }
    <FormInput {...props} />
    // Display error message if it's specified (errorMessage is a simple string)
    {errorMessage &&
      <FormValidationMessage>
        {errorMessage}
[...]

PR #482, opened 2017-07-12
Complete FormValidationMessage docs

martinezguillaume:
Add some precision of the usage of `FormValidationMessage` component ([ref](https://github.com/react-native-training/react-native-elements/issues/377#issuecomment-313602329))


iRoachie:
Thanks @martinezguillaume for submitting this PR! In the future, I think we can add an example using state for multiple error messages.

PR #484, opened 2017-07-12
Allow containerBorderRadius to equal zero for ButtonGroup

JulianKingman:
Currently, a fix in ButtonGroup allows passing a containerBorderRadius prop. The current implementation uses this: `containerBorderRadius || 3`, which means that if containerBorderRadius is 0, then it will display 3, not 0.

This is important because displaying a border radius when the background is a different color causes a visual artifact (pictured below, notice the white triangles in the corner).

screen shot 2017-07-12 at 10 29 52 am

### DIFF
--- a/src/buttons/ButtonGroup.js
+++ b/src/buttons/ButtonGroup.js
@@ -40,6 +40,7 @@ const ButtonGroup = props => {
       {...attributes}
     >
       {buttons.map((button, i) => {
+        const containerRadius = !isNaN(containerBorderRadius) ? containerBorderRadius : 3;
         return (
           <Component
             activeOpacity={activeOpacity}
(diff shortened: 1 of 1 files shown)


src/buttons/ButtonGroup.js
```diff
@@ -68,12 +69,12 @@ const ButtonGroup = props => {
...
-                borderTopRightRadius: containerBorderRadius || 3,
-                borderBottomRightRadius: containerBorderRadius || 3,
+                borderTopRightRadius: containerRadius,
+                borderBottomRightRadius: containerRadius,
               },
               i === 0 && {
-                borderTopLeftRadius: containerBorderRadius || 3,
-                borderBottomLeftRadius: containerBorderRadius || 3,
+                borderTopLeftRadius: containerRadius,
+                borderBottomLeftRadius: ccontainerRadius,
```
iRoachie: @JulianKingman There's a a typo here. You have two 'c's

JulianKingman:
Thanks! Fixed

iRoachie:
Thanks @JulianKingman

PR #488, opened 2017-07-13
Allow Card overlay container styles to be overridden

SocialsuiteJesse:
Allow styling to be overridden, particularly so you can set the background color and opacity manually


iRoachie:
Hey @SocialsuiteJesse The PR is good but you have to update the snapshots(tests) since you're adding an additional value. `npm test -- -u`

Can you also add the property to the docs so others will know?

PR #492, opened 2017-07-14
Refactor Button Component 

Monte9:
## Major Breaking changes 

Please see the [`/example` app](https://github.com/react-native-training/react-native-elements/blob/button_v1/example/src/views/buttons_home.js) for the above implementation.

### Really simple and useful API:
```
Button.propTypes = {
  text: PropTypes.string,
  textStyle: PropTypes.object,
  textProps: PropTypes.object,

  buttonStyle: ViewPropTypes.style,

  loading: PropTypes.bool,
  loadingStyle: ViewPropTypes.style,
  loadingProps: PropTypes.object,

  onPress: PropTypes.any,
  containerStyle: ViewPropTypes.style,

  icon: PropTypes.object,
  iconContainerStyle: ViewPropTypes.style,
  iconRight: PropTypes.bool,
};
```


Monte9:
@binoy14 @iRoachie thoughts on above?

Basically, here's what I had in mind while refactoring:
1. Simplify the named Props handles by `Button.js`
2. Make the code more legible and modularized ( Icon, for instance, needs to be passed in as an object, because the Button doesn't care about how/where you get your icon from)
3. Simple and easy to understand button layout with full flexibility by using `containerStyle` & `style` props
4. Implement common use cases as seen in various designs

Here's an example of what a button looks like:
```
<Button
  text="Add to Cart"
  underlayColor="rgba(80,117,196,1)"
  activeOpacity={0.4}
  textStyle={{fontWeight: 'bold', fontSize: 18}}
  buttonStyle={{backgroundColor: 'rgba(90, 154, 230, 1)', height: 40, width: 200, borderWidth: 0, borderColor: 'transparent', borderRadius: 20}}
  containerStyle={{marginVertical: 10}}
  icon={
    <Icon
      name='arrow-right'
      size={15}
      color='white'
    />
  }
  iconRight
  iconContainerStyle={{marginLeft: 5}}
/>
```

This will breaking existing functionality, but I think I would rather have a cleaner and simpler API to work with for `1.0` than just patch and band-aid fixes for the current API.

Also here's some useful desings to implement in case anyone want to play with the `Button` API.
- http://www.designyourway.net/blog/inspiration/menus-and-buttons-in-mobile-design-45-examples/
- https://www.behance.net/gallery/47612471/Travel-Leisure-Concept-Re-Design

src/buttons/Button.js
```diff
@@ -1,281 +1,182 @@
 import PropTypes from 'prop-types';
```
Monte9: this is the file that needs to be review for the API changes

nonotest:
+1 for the icon change, hate passing an object for this.
also +1 for the style changes, current api can be confusing

Monte9:
Thanks for the feedback @nonotest!

xavier-villelegier:
This includes breaking changes but i think it is worth it. I really like the way you use `textProps` and `loadingProps` instead of exposing the entire API of those components through the `Button` component, I think this could be a good start point to simplify others ! 🔥 
+1 for the icon change too,  much more modular.

Monte9:
@nonotest @xavier-villelegier thanks for pitching in. I used this API myself to design buttons on the `example` app and I really like it. 

I am merging this PR now. I have added a few todo items [here](https://github.com/react-native-training/react-native-elements/projects/4).

PR #500, opened 2017-07-16
[SocialIcon] Fix boolean value for onLongPress

### LINKED ISSUE #499: [SocialIcon] Invalid prop `onLongPress` of type `boolean`
### Description

Fairly simple bug in code logic. When the `SocialIcon` is disabled, the `onLongPress` receives `false` value. 
I am running version 0.14.0 of elements.

I am not sure if the Snack has some different version or somehow it's blocking warnings, but I could not reproduce it there. However looking at the code it's kinda obvious error.

https://snack.expo.io/SydPdo_HW

### Solution

It should be probably [done same way](https://github.com/react-native-training/react-native-elements/blob/master/src/social/SocialIcon.js#L84) as `onPress` prop.

FredyC:
Fixes #499

### DIFF
--- a/src/social/SocialIcon.js
+++ b/src/social/SocialIcon.js
@@ -81,7 +81,7 @@ const SocialIcon = props => {
   return (
     <Component
       underlayColor={light ? 'white' : colors[type]}
-      onLongPress={!disabled && (onLongPress || log)}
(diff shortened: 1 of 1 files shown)


iRoachie:
Thanks for quick fix @FredyC! Works great. Will be live in the release Monday.

PR #514, opened 2017-07-22
TextButton - Uses Button component

Monte9:
It uses `Button.js` under the hood and takes in all the props of the Button component. 

screen shot 2017-07-22 at 10 58 59 am

Simple use case:
```<TextButton text="Create an Account" />```

### CODE REVIEW

File: src/buttons/TextButton.js [hunk 1]
```diff
@@ -0,0 +1,20 @@
...
+        text="Text Button"
+        buttonStyle={{ backgroundColor: 'transparent', width: 130, height: 40 }}
+        textStyle={{ color: 'rgba(78, 116, 289, 1)' }}
+        underlayColor="transparent"
+        activeOpacity={0}
+        {...attributes}
+      />
+    );
+  }
+}
```
Monte9: this is the file that needs to be reviewed.

Monte9:
@xavier-villelegier @iRoachie thoughts?

PR #524, opened 2017-07-24
Update: updated react-native-side-menu

formula1:


binoy14:
@formula1 Thanks for the PR. We are removing the side menu from react-native-elements so I am not sure how much value it will provide to update it right now. I'll leave the PR open and @iRoachie or @Monte9 can decide if they want to merge it

Monte9:
That's a good point @binoy14, although we are probably going to have a few releases before `1.0` is released. 

So in the meantime, it's prob a good idea to keep the packages up to date. Thanks for the update @formula1 👍

Monte9:
This upgrade looks ok to me as there don't seem to be any (documented) breaking changes in the latest version. Here are the [release notes](https://github.com/react-native-community/react-native-side-menu/releases/tag/v1.0.0)

PR #529, opened 2017-07-29
0.16

iRoachie:
### Release notes:
https://github.com/react-native-training/react-native-elements/releases/tag/untagged-ba7d8ba3138c470498c3


src/form/FormInput.js
```diff
@@ -149,13 +146,12 @@ const styles = StyleSheet.create({
...
       android: {
         minHeight: 46,
-        width: width - 30,
       },
       ios: {
         minHeight: 36,
-        width: width,
       },
     }),
+    width: '100%',
```
binoy14: @iRoachie This will only work on RN 0.43 or 0.42+ It will break for anyone not on that version. Is that something that was decided?

iRoachie: @binoy14 Not necessarily, but what other options do we have? Right now the width of the formInputs are calculated based on the screen width. So if they're nested inside a parent thats smaller then it will overflow the parent. If you have another way we could go around it, I'm all for it

src/header/Header.js
```diff
@@ -68,6 +70,17 @@ const Header = props => {
...
+Header.propTypes = {
+  leftComponent: PropTypes.object,
+  centerComponent: PropTypes.object,
+  rightComponent: PropTypes.object,
+  backgroundColor: PropTypes.string,
+  outerContainerStyles: ViewPropTypes.style,
+  innerContainerStyles: ViewPropTypes.style,
+  children: PropTypes.element,
+  statusBarProps: PropTypes.object,
+};
```
Monte9: thanks for adding this @iRoachie 👍

iRoachie: Sure np!

PR #530, opened 2017-07-29
Add shake animation to Input v1

xavier-villelegier:
# Purpose

Hey !
Here is the shake input animation that @martinezguillaume did, but on the v1 input. 🔥 

I replicated the shake props logic, because I think using redux and stateless component it can be useful to pass directly the `error` object returned from an API call to the input, and it will shake itself if the `error` is `!==` from the previous one (and not falsy obviously).

I updated the `screen3`, as it is already animated it will fit perfectly.

Let me know if some changes are required ✌️ 

# Example

juil -29-2017 10-38-57

(It is quite laggy on a gif tho...)

### DIFF
--- a/example/src/views/login/screen3.js
+++ b/example/src/views/login/screen3.js
@@ -94,8 +94,8 @@ export default class LoginScreen2 extends Component {
       LayoutAnimation.easeInEaseOut();
       this.setState({
         isLoading: false,
-        isEmailValid: this.validateEmail(email),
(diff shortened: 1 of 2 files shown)


Monte9:
@xavier-villelegier this is great! 💯 Merging it in now. 

p.s - don't forget to cash the [payout](https://github.com/react-native-training/react-native-elements/projects/4#card-3823160) for this ticket. 😉

PR #532, opened 2017-08-01
allow passing react element to Card title

chilijung:
right now, I'm getting this error while passing React element to Card component

```
ExceptionsManager.js:71 Warning: Failed prop type: Invalid prop `title` of type `object` supplied to `Card`, expected `string`.
```

allow passing element in Card title

### DIFF
--- a/src/card/Card.js
+++ b/src/card/Card.js
@@ -111,7 +111,7 @@ Card.propTypes = {
   flexDirection: PropTypes.string,
   containerStyle: ViewPropTypes.style,
   wrapperStyle: ViewPropTypes.style,
-  title: PropTypes.string,
+  title: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
   titleStyle: NativeText.propTypes.style,
   featuredTitle: PropTypes.string,
   featuredTitleStyle: Text.propTypes.style,
--- a/src/card/__tests__/Card.js
+++ b/src/card/__tests__/Card.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { Text } from 'react-native';
 import { shallow } from 'enzyme';
 import toJson from 'enzyme-to-json';
 import Card from '../Card';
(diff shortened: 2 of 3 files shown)


iRoachie:
Hey @chilijung thanks for PR. Currently right now the library only displays titles as strings, and then the title is styled using the `titleStyle` prop as well as other styles. Passing in an element as a title would nest that element in the `<Text>` element that's already there, see [here](https://github.com/Canner/react-native-elements/blob/2cafc860fd4f426184c1be4d9958efe1be83cb6a/src/card/Card.js#L57)

Nesting Text elements inside each other sometimes throws off the styling and might produce unwanted styling effects. Might need some opinions from the others @Monte9 @xavier-villelegier @binoy14

chilijung:
hi @iRoachie I think allow users to pass elements to title, can help many use cases, and allow more control over title.

xavier-villelegier:
@iRoachie It seems legit to me. But maybe it implies more changes in the PR. To my mind, the `title` should be wrapped in a `View` (which is already the case) that ensures that the title will always behave correctly (at the top of the `Card`). And then the user can pass whatever he wants (an `Icon`, a nested `Text`, even a `Button`, who knows !), it will be displayed at the top of the `Card`.

Moreover, in the `v1` it will probably be this way. The v1 `Button` takes an Icon component [for example](https://github.com/react-native-training/react-native-elements/blob/v1/src/buttons/Button.js#L78). It is just wrapped to be sure that it will behave as expected, no matter what is passed.

But as he has planned to look into the v1 `Card`, it could be good to know what @binoy14 thinks about this PR ! 👌

Monte9:
@chilijung Thanks for submitting the PR! I def understand the need for allowing the title to be more than just text. In fact, this is something that's going to be handled in `v1` that's coming out soon. 

But as far as this PR is concerned, just changing the PropType validation to allow for it to render a React Element is not a good idea as @iRoachie pointed out. For a full solution, you should prob check where the title props is a text/element and render a view the elment around it as @xavier-villelegier was saying. 

Closing this PR now as its only half the solution, but if this is something you need ASAP, we are experimenting with "Paid feature requests" on this repo where you could ask someone to make changes for you as you need them. Feel free to ask me more if you are interested.

PR #533, opened 2017-08-01
Update "Badge props" to "Divider props"

SCasarotto:
The title was mislabeled for the props.


iRoachie:
Thanks @SCasarotto good find 💯

PR #534, opened 2017-08-02
Support right icon and left icon at the same time.

yeknava:
i kept iconRight's prop for compatibility but no longer needed.

### DIFF
--- a/src/buttons/Button.js
+++ b/src/buttons/Button.js
@@ -57,26 +57,54 @@ const Button = props => {
     allowFontScaling,
     ...attributes
   } = props;
-  let { Component } = props;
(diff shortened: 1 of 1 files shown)


Monte9:
@yeknava Thanks for the PR. Basically this PR allows you to 
- pass `icon` to render it on the left
- pass `rightIcon` to render it on the right
- pass `rightIcon` and `leftIcon` to render icons on the left & right?

Is that accurate? If so, seems like `icon` and `leftIcon` can be consolidated as 1 prop. 

I'll come back to this after `v0.16.0` release.

yeknava:
@Monte9 welcome, exactly, the reason i kept icon and left icon props is to save backward compatibility.

PR #539, opened 2017-08-04
Update package.json

puhfista:


iRoachie:
What’s the reason for this ?

puhfista:
The latest version of react-native includes a version of react that does not include PropTypes inside the react project itself. As of version .47 of react native, any projects that try to use React.PropTypes.[type] will throw a run time error :( The author of the changed library in question fixed the problem in 1.0.2. 

Also, the same problem is happening for react-native-tab-navigator .... it is trying to use React.Proptypes.[type] =\

iRoachie:
I’m aware of the PropTyoes change and I’ll accept a PR that upgrades those libraries to a version that has the PropTypes fix. However we can’t remove the tab navigator completely as yet as that would be a breaking change and people are still using it in production. We’ve already removed it and the other deprecated components in the v1 branch.

Monte9:
@puhfista Thanks for the PR! 

Couple things:
1. the version of `side-menu` is already the latest on the next branch:
```"react-native-side-menu": "^1.0.2",```
2. As @iRoachie pointed out, we cannot yet remove the Tab Navigator yet as some users might still be depending on it. All users see a deprecated warning on `v0.15.0` and it will be complely removed on `v1` coming soon.

PR #541, opened 2017-08-06
Profile v1

Monte9:
simulator screen shot aug 14 2017 8 15 53 pm
simulator screen shot aug 14 2017 9 00 01 pm
simulator screen shot aug 14 2017 9 00 09 pm


example/src/views/profile/screen1.js
```diff
@@ -0,0 +1,246 @@
...
+              <View style={{flex: 1, marginTop: 30}}>
+                <Text style={{flex: 1, fontSize: 15, color: 'rgba(216, 121, 112, 1)', fontFamily: 'regular', marginLeft: 40}}>
+                  INTERESTS
+                </Text>
+                <View style={{flex: 1, width: SCREEN_WIDTH, marginTop: 20}}>
+                  <ScrollView
+                    style={{flex: 1}}
+                    horizontal
+                    showsHorizontalScrollIndicator={false}
+                  >
```
Monte9: here's the scrollview that works on ios but not on Android.

PR #544, opened 2017-08-08
Update Divider Props subheading

dustykeyboard:


Monte9:
Good catch. Thanks for the PR @dustykeyboard!

PR #546, opened 2017-08-09
Register custom icon types

cranberyxl:
`react-native-vector-icons` allows you to use [custom icon fonts](https://github.com/oblador/react-native-vector-icons#custom-fonts), but `react-native-elements` does not. I added the ability to register your own custom fonts and use them anywhere an icon type is used.

```js
import { registerCustomIconType } from 'react-native-elements';
import { createIconSetFromFontello } from 'react-native-vector-icons';
import fontelloConfig from './fonts/custom-icons/config.json';
registerCustomIconType('custom', createIconSetFromFontello(fontelloConfig));
```


src/helpers/getIconType.js
```diff
@@ -1,15 +1,20 @@
 import ZocialIcon from 'react-native-vector-icons/Zocial';
 import OcticonIcon from 'react-native-vector-icons/Octicons';
 import MaterialIcon from 'react-native-vector-icons/MaterialIcons';
-import MaterialCommunityIcon
-  from 'react-native-vector-icons/MaterialCommunityIcons';
+import MaterialCommunityIcon from 'react-native-vector-icons/MaterialCommunityIcons';
```
cranberyxl: The precommit hook is changing this whitespace, even though it isn't part of my change

Monte9: Noted.

Monte9:
@cranberyxl thanks for the PR! The is pretty useful. 

Before I merge this PR, could you please add a demo for using custom Fonts into the `/example` expo app? It's useful to point users to an implementation in case they are wondering how to use it. Let me know if you have any q's about the `/example` app. Thanks

iRoachie:
@Monte9 no response for a week? Can we just merge it in?

cranberyxl:
I'm on holiday. I'll be able to look into the demo next week.

cranberyxl:
Here is an example of it in use. Because of Expo I had to dance around what would be the normal asset behavior you would use to load a custom font. Note: you have to change the `example/package.json` to point to this branch to see it in action.

Monte9:
This is great! Thanks for adding an example for it @cranberyxl! 👍 
I'll test it out once I merge it into `next`.

Just fyi, this will be part of the `v0.17` and will be release within the next day to two. Again, thanks for the PR! 👏

PR #553, opened 2017-08-15
Initial trials for ThemeProvider

iRoachie:
Right now this is proof of concept for a ThemeProvider. Right now it only accepts overrides for the base colours. 

In the future, we'll need to add other fields as well as better naming. e.g(checkboxColor, badgeBackgroundColor)

#### Pending
- [ ] Example


iRoachie:
Updated description for better gist.

src/config/getTheme.js
```diff
@@ -0,0 +1,421 @@
...
+    checkbox: StyleSheet.create({
+      wrapper: {
+        flexDirection: 'row',
+        alignItems: 'center',
+      },
+      container: {
+        margin: 5,
+        marginLeft: 10,
+        marginRight: 10,
+        backgroundColor: '#fafafa',
```
binoy14: Should this be using the colors object?

iRoachie: Not sure, it wasn’t using it before

binoy14: Not sure what it affects but probably a good idea to take a look what it's doing

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+  "editor.formatOnSave": true,
+  "prettier.eslintIntegration": true
+}
```
Monte9: this should not be committed. Please add this file to `.gitignore`

iRoachie: Ooops

iRoachie:
Taking a step back and going to make an easier way to show how it works, and an upgrade path for making it available for all components

Monte9:
Ok, that's good to know @iRoachie! Let's plan to swing back to this after we do the v1 release. That way we will already have the new API changes & components out and being used. 

So this theme PR will be easier to implement & release.

iRoachie:
Sounds good to me!

PR #555, opened 2017-08-17
Add LinearGradient from Expo to Button v1

martinezguillaume:
First implementation of a LinearGradient button for the v1.

LinearGradient from expo has different props like colors, start, end and locations to play with colors. Do we have to ignore them and just put the colors props as gradientColors ? or implement a props like linearGradientProps where the user can inject the props from LinearGradient that he wants.


martinezguillaume:
Wrong base

PR #557, opened 2017-08-18
Add LinearGradient from Expo to Button

martinezguillaume:
First implementation of a LinearGradient button for the v1.

LinearGradient from expo has different props like colors, start, end and locations to play with colors. 
You can add this props to `linearGradientProps` prop.
Of course, this feature is only available for expo user to make it simple to have gradient button.

img_1568


example/v1/buttons/Button.js
```diff
@@ -24,57 +24,76 @@ class Button extends Component {
...
       ...attributes
     } = this.props;
 
     // this is what RN Button does by default
     // https://github.com/facebook/react-native/blob/master/Libraries/Components/Button.js#L118
-    const Touchable = Platform.OS === 'android' ? TouchableNativeFeedback : TouchableOpacity;
+    const Touchable =
+      Platform.OS === 'android' ? TouchableNativeFeedback : TouchableOpacity;
+    const ButtonContainer = linearGradientProps
+      ? require('expo').LinearGradient
```
Monte9: I'd be curious to see what happens when I pass `linearGradientProps` when not using expo. 

Since there is no fallback, I would think this would throw an error. But I am willing to merge it and try it out.

Monte9:
I am curious to try this out. Merging this into `v1` and will add an demo to the `/example` app as well.

Also we need documentation for this. But since the Button component on `v1` has a new API, we can write the docs for Button later on. Def a TODO.

Thanks for the PR @martinezguillaume! 💯

Monte9:
Beautiful! 😍 I love it!
```
<Button
  buttonStyle={{ height: 55, width: SCREEN_WIDTH - 40, borderRadius: 30, justifyContent: 'center', alignItems: 'center' }}
  linearGradientProps = {{
    colors: ['rgba(214,116,112,1)', 'rgba(233,174,87,1)'],
    start: [1, 0],
    end: [0.2, 0]
  }}
  text="Message Theresa"
  textStyle={{ fontFamily: 'regular', fontSize: 20, color: 'white', textAlign: 'center' }}
/>
```
screen shot 2017-08-26 at 12 59 54 am

Monte9:
If someone can test it's behaviour on non-expo apps, that would be great. 
/cc @iRoachie @xavier-villelegier

Issue #538, opened 2017-08-04
ListItem textInput doesn't show up


anicoa:
### Description

I want to use ListItem with `textInput={true}` in my project but i can't get the input-field to show up. Not sure if it's my lack of knowledge or some other problem. Title is shown as expected. Chevron ist also there. Doesn't work with `hideChevron={true}` also.

### Reproduction Steps and Sample Code

Initialize new project
`react-native init AwesomeProject`

Install react-native-elements and react-native-vector-icons
```
npm --save install react-native-elements
npm install --save react-native-vector-icons
react-native link
```

Insert List/ListItem 

```
import React, { Component } from 'react';
import {
  AppRegistry,
  StyleSheet,
  Text,
  View
} from 'react-native';

import { List, ListItem } from 'react-native-elements';

export default class AwesomeProject extends Component {
  render() {
    return (
      <View>
        <List>
          <ListItem title='ListItem with Placeholder' textInput={true} />
        </List>
      </View>
    );
  }
}

AppRegistry.registerComponent('AwesomeProject', () => AwesomeProject);
```

Used versions
```
    "react": "16.0.0-alpha.12",
    "react-native": "0.47.1",
    "react-native-elements": "^0.15.0",
    "react-native-vector-icons": "^4.3.0"
```

### Solution

none so far. Would be very kind if someone could test/confirm or point me into the right direction.

uffou:
Having the same problem ListItem textInput is not working :( 
```
<List>
				<ListItem
					title='Title'
					textInput={true}
					textInputValue={this.state.title}
					textInputOnChangeText={(title) => this.setState({title})}
					textInputReturnKeyType={'done'}
				/>
		  	</List>
```

Monte9:
@anicoa @uffou hmm.. sorry about that. Someone from RNE can help you resolve this soon.

Smashbandit:
I'd love to see this get resolved as well :) I'm having the same issue.

Monte9:
kk. I'll look into this tonight! Will update you guys soon!

martinezguillaume:
Are you on Android or iOS or both ? For me, the textinput shows up in iOS in the future version of react-native-elements

martinezguillaume:
In v16, it works on iOS. This [pr](https://github.com/react-native-training/react-native-elements/pull/565) fix it on Android

iRoachie:
Thanks @martinezguillaume will review it later

iRoachie:
[...]

PR #565, opened 2017-08-22
Fix ListItem TextInput on Android

martinezguillaume:
ref: https://github.com/react-native-training/react-native-elements/issues/538

Before:
screenshot_2017-08-22-14-38-49

After:
screenshot_2017-08-22-14-39-46


src/list/ListItem.js
```diff
@@ -190,9 +190,16 @@ const ListItem = props => {
             </Text>
           </View>}
         {textInput &&
-          <View style={[styles.rightTitleContainer, textInputContainerStyle]}>
+          <View
+            style={[
+              styles.rightTitleContainer,
+              { alignItems: null },
```
iRoachie: Instead of inlining this here, can you make add it to the styles as textInputContainerStyle?

martinezguillaume: My gosh, yes of course !

iRoachie: Thank you!

PR #573, opened 2017-08-24
Update Tile.js: adding 'numberOfLines' prop

HOUCe:
Adding 'numberOfLines' prop for Text component (used to display title) in Tile.js.
So users can set numberOfLines for their 'title' text.

image


yanhaijing:
i need this

iRoachie:
Sweet PR @HOUCe, I will definitely allow this. Can you add it to the docs as well?

HOUCe:
Sure, with great pleasure.

PR #576, opened 2017-08-25
fix(FormInput): follow the official doc with keyboardType

flexbox:
According to the spec
https://facebook.github.io/react-native/docs/textinput.html#keyboardtype

Here is an example with a `keyboardType={'email-address'}`

__Before:__
screen shot 2017-08-25 at 16 16 35

__After:__
screen shot 2017-08-25 at 16 16 46

The email keyboard appears 🎉


iRoachie:
Good catch @flexbox. I think instead of just adding all of the props, we could do one prop called `textInputProps`, then spread it into the TextInput. So we wouldn't have to keep adding more props.

@Monte9 @xavier-villelegier

Monte9:
Thanks for the PR @flexbox! Keeping things consistent with RN is important. 

@iRoachie this PR just renames the prop to be consistent with RN. Not adding any new props. But I agree with you that all the props should just be spread on the component as opposed to explicitly adding them to the component. 

In fact, I wrote a really clean implementaion of the `Input` component on `v1`, just fyi.

PR #579, opened 2017-08-30
Update badge.md

artdias90:
Caused some confusion when there is a <Text> element inside the Badge Component.


iRoachie:
Thanks @artdias90, good catch

Issue #535, opened 2017-08-03
Raised Button With Border Radius - Android API 23


KyleScharnhorst:
Wanted to have a raised button that also had a border radius.

Here is the result:
![radiusbuttonraised](https://user-images.githubusercontent.com/9146614/28920112-48c724c4-7805-11e7-9d41-de274a2eaac9.PNG)

[Snack Expo Link](https://snack.expo.io/SyIlMcePZ)

Version of `react-native-elements`: 0.15.0

Is there something that I'm missing or is this a bug, can anyone confirm?

xavier-villelegier:
Hey @KyleScharnhorst 
You can add `containerViewStyle={{borderRadius: 20}}` to have a clean button.
This is not really a bug, but it's true that the borderRadius should also be applied to the view wrapping the button [here](https://github.com/react-native-training/react-native-elements/blob/master/src/buttons/Button.js#L130).

sujit13666:
Hi @xavier-villelegier , I am facing same problem with **0.15.0**. `containerViewStyle` seems not to work.

`<Button containerViewStyle={{borderRadius:25}} buttonStyle={{width:145,height:45,borderRadius:25}} textStyle={style.buttonTextStyle} onPress={onPress} title={text}/>`

![signin](https://user-images.githubusercontent.com/5456641/28961336-c522377e-7923-11e7-9b31-8a6a4680c1b8.jpg)

xavier-villelegier:
@sujit13666 Mmh that's not really the same issue. The border radius of the `TouchableHighlight` on Android is a bit more complicated to handle. Try the following:

```jsx
<Button
  borderRadius={25}
  containerViewStyle={{borderRadius:25}}
  buttonStyle={{width:145,height:45,borderRadius:25}}
  textStyle={style.buttonTextStyle}
  onPress={onPress}
  title={text}/>
```

sujit13666:
That's just working like a charm. @xavier-villelegier 

> The border radius of the `TouchableHighlight `on Android is a bit more complicated to handle

Totally agree with you. Otherwise `borderRadius` would not be needed twice (as `borderRadius` itself and in `containerViewStyle`)

xavier-villelegier:
Yeah for sure. The `borderRadius` props should be applied to the `containerViewStyle` and `buttonStyle` automatically. This will probably be handled in a better way in the `v1` ! But feel free
 to open a PR if you want this to be more intuitive in the `v0.16`✌️

[...]

PR #581, opened 2017-08-31
Fix(Button): Border Radius was not applied to view container.

KyleScharnhorst:
[issue #535](https://github.com/react-native-training/react-native-elements/issues/535), where to have a clean button you would also need to apply the
border radius value to the containerViewStyle prop so that the containing
view would wrap the rounded button. Test snapshots were updated to reflect
the change.

### DIFF
--- a/src/buttons/Button.js
+++ b/src/buttons/Button.js
@@ -111,7 +111,8 @@ const Button = props => {
 
   const baseFont = {
     color: (textStyle && textStyle.color) || color || stylesObject.text.color,
-    size: (textStyle && textStyle.fontSize) ||
+    size:
+      (textStyle && textStyle.fontSize) ||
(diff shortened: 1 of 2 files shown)


src/buttons/Button.js
```diff
@@ -127,7 +128,12 @@ const Button = props => {
 
   return (
     <View
-      style={[styles.container, raised && styles.raised, containerViewStyle]}
+      style={[
+        styles.container,
+        raised && styles.raised,
+        containerViewStyle,
+        borderRadius && { borderRadius },
```
Monte9: 👍

Monte9:
Thanks for the quick fix @KyleScharnhorst! 
This will be a part of `v0.17` and will be released in the next day or two.

PR #607, opened 2017-09-18
fix #597

### LINKED ISSUE #597: Button with borderRadius prop crash on API < 21
Buttons with a border radius like so only works on Lollipop and above because of the ripple drawable is not available.

`<Button
              title="Login"
              borderRadius={5}
              onPress={() => this._login()}/>`

https://github.com/react-native-training/react-native-elements/blob/32bf779930f928bf2a22658dcb9ca729535610d1/src/buttons/Button.js#L106

We should add an if API > 21 on that and add a proper borderRadius @Monte9

jjdp:


iRoachie:
@jjdp Can you rebase this with next and submit it again?

jjdp:
sure ill submit when i come home. closing this

PR #610, opened 2017-09-20
Rename iconLeft to leftIcon in button docs api

chengsieuly:
`iconLeft` is not a valid prop for Button. This should be `leftIcon`.


iRoachie:
Thank you @chengsieuly! Great first contribution.

chengsieuly:
Thanks @iRoachie ! Hopefully, more to come =)

PR #616, opened 2017-09-24
0.18

iRoachie:
Let's try to release week of October 8


docs/API/header.md
```diff
@@ -62,17 +62,19 @@ Components defined through props take precedence over components passed in as ch
 
 ### Header customisability
 
-We wanted the Header to be as customisable as possible, so you are free to try different combinations of props.
+We wanted the Header to be as customisable as possible, so you are free to try different combinations of props.For example, if you want to change the left, center, or right component's layout, you can adjust the ` innerContainerStyles `
```
Monte9: nitpicky. missing space between `props.For`

src/header/__tests__/Header.js
```diff
@@ -28,6 +28,17 @@ describe('Header Component', () => {
...
+    const component = shallow(
+      <Header>
+        <Button title="Test button 1" onPress={() => {}} />
+        <Button title="Test button 2" onPress={() => {}} />
+      </Header>
+    );
+
+    expect(component.find('Button').length).toBe(2);
+  });
+
```
Monte9: thanks for writing the test 🙏

src/index.d.ts
```diff
@@ -0,0 +1,2248 @@
+// Type definitions for react-native-elements 0.16
```
Monte9: @iRoachie is this something that will have to be maintained and kept up to date with each new version? 

What's the process of generating a new typed `index.d.ts` file?

src/index.d.ts
```diff
@@ -0,0 +1,2248 @@
...
+// Project: https://github.com/react-native-training/react-native-elements#readme
+// Definitions by: Kyle Roach <https://github.com/iRoachie>
+//                 Ifiok Jr. <https://github.com/ifiokjr>
```
Monte9: also we should consider adding this to the docs to let users know that it exists and how they can use it if they are beginners in typescript. 

How about we do a fast follow after v0.18 release to add docs for it? thanks.

src/index.d.ts
iRoachie: Yes! It's not generated automatically. Has to be maintained manually. I don't mind being in charge  for that

src/index.d.ts
iRoachie: Previously this definition was on DefinitelyTyped which required an extra step after `npm i react-native-elements`. By moving the declaration here, no extra setup is needed

docs/API/header.md
Monte9: fixed.

src/index.d.ts
Monte9: Great. That works for me. If it gets too cumbersome, esp with the changes on v1.. I don't mind helping out. lmk.

src/index.d.ts
Monte9: >  add docs for it

docs/Installation/default_installation.md
```diff
@@ -4,19 +4,19 @@ If your project is a standard React Native project created using `react-native i
 
 ## Step 1: Install react-native-vector-icons
 
-If you already have this installed, or are using create-react-native-app, this isn't necessary.
+If you already have intalled **react-native-vector-icon** as a dependency for your project you can skip this step. Otherwise run the following command:
```
Monte9: fix typo `intalled`

src/index.d.ts
iRoachie: Oooo alright then guess I'll do that

PR #609, opened 2017-09-20
I have a problem understanding how to modify header. So I fix the document

pacozaa:
- Default of `outerContainerStyles`  and `innerContainerStyles` is not `none` as the document state.

-  I give and example how to manage `innerContainerStyles` which is important

PR #619, opened 2017-09-25
Add more detail for Header document

pacozaa:
I edit my pull request regarding #609


iRoachie:
Sweeet! Love it! Thanks again @pacozaa

PR #635, opened 2017-10-03
resizeMethod (Android) added to images

dale-french:
**First of all, awesome work you guys are doing. Thank you!**

I have added a _resizeMethod_ prop to images on the Avatar and Card components as I don't think it is always necessary to handle image resizing straight away. My case is the user presses on the avatar which then lets the user choose a profile image. On success I replace the image on the avatar component.

If this image is quite large it causes a lag and then the app crashes. That is why i have added the _resizeMethod_ prop.

More details can be found here: https://facebook.github.io/react-native/docs/image.html#resizemethod 

Please note that this is **Android only** and it defaults to 'auto'.

### DIFF
--- a/src/avatar/Avatar.js
+++ b/src/avatar/Avatar.js
@@ -40,6 +40,7 @@ const Avatar = props => {
     showEditButton,
     editButton,
     onEditPress,
+    resizeMethod,
     ...attributes
   } = props;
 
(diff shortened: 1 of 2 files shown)


iRoachie:
Thanks, @dale-french for the PR. Gonna close this in favour of https://github.com/react-native-training/react-native-elements/commit/2eb6806782d662e03f51793960b3182d7b74bdbf which has a more generic way of passing props down to the images.

imageprops

PR #643, opened 2017-10-06
feat(Icon): added Stroke ability

kosiakMD:
feat(Icon): added Stroke ability

Added next props:
 - stroke: boolean, default = false;
 - strokeWidth: number, via borderWidth, default = 1;
 - strokeColor: string, via borderColor, default = this.props.color || 'black';

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -1,0 +1,6 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
+    <inspection_tool class="Eslint" enabled="true" level="ERROR" enabled_by_default="true" />
+  </profile>
+</component>
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -1,6 +1,0 @@
-<component name="InspectionProjectProfileManager">
-  <profile version="1.0">
-    <option name="myName" value="Project Default" />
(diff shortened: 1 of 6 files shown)


kosiakMD:
@iRoachie, deleted

iRoachie:
Hey guys @martinezguillaume @Monte9 @xavier-villelegier can I get your thoughts on this? This is a utility over passing a style object over to the Icon. It’s not apart of the react-native-vector-icons Icon props.

xavier-villelegier:
@iRoachie @kosiakMD 
If I understand correctly, this:
```jsx
<Icon
  …
  stroke
  strokeWidth={1}
  strokeColor={'red'}
/>
```
is equivalent to:
```jsx
<Icon
  …
  containerStyle={{
    borderColor: 'red',
    borderWidth: 1,
  }}
/>
```

? If so, I don't know if it is really more intuitive to use a different props to apply styles on the component 🤔

Moreover, this prevents people from factorizing by using the same style across the file by doing
```jsx
<Icon
  …
  containerStyle={styles.icon}
/>
…
  icon: {
    borderColor: 'red',
    borderWidth: 1,
  },
})
```

Monte9:
@kosiakMD thanks for the PR! This is a good idea and I can see people using it by passing the stroke props explicitly.

Although I think @xavier-villelegier brings up a good point, i.e by creating explicit props for adding borderWidth & borderColor, it can confuse the user and also make the style less reusable.

Moreover, it adds more lines to the source code instead of keeping the Icon component as lean as possible. 

Again, thanks for the PR, but I think it's better to use containerStyle in order to achieve what you are doing.

PR #661, opened 2017-10-19
Update to React@16

FredyC:
I am not too familiar with this project, using just a couple components of it. However, trying to use it with React@16 ended up with error of using `PropTypes` in the "old-way" in `react-native-side-menu`. I've noticed they have already released an update with that fixed. So far it seems this is only blocker to be React@16 compatible.


iRoachie:
Thanks @FredyC ! Been meaning to do this

Issue #355, opened 2017-05-05
Disable ListItem


RWOverdijk:
Hey! For the ListItem, it would be nice if it could be disabled. I can do it myself, by checking (and I do). I just thought I'd put this out there.

Monte9:
@RWOverdijk that's a good idea. Should be fairly straightforward especially since you are already doing it for your use case. Would you like a submit a PR for this?

RWOverdijk:
@Monte9 Would that include some sort of style change? `disabledStyle` or something?

Monte9:
yup. So one possible way would be to have a `disabled` prop that would apply some defauly disabled styling to the `ListItem`. In addition you would expose the `disabledStyle` prop to allow the user to add custom styles when it's disabled.

iRoachie:
Implemented in v18.0.0

PR #662, opened 2017-10-19
feat #355 Allow disable on list item

krose422:
Allows disabling of listItem and exposes disabledStyle prop for custom disabled style to listItem.

### DIFF
--- a/src/list/ListItem.js
+++ b/src/list/ListItem.js
@@ -104,8 +104,8 @@ const ListItem = props => {
       style={[
         styles.container,
         containerStyle && containerStyle,
-        disabled && { opacity: .5 },
-        disabled && disabledStyle && disabledStyle
+        disabled && styles.disabled,
+        disabled && disabledStyle && disabledStyle,
       ]}
       {...attributes}
     >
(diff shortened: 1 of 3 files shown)


kyler-hyuna:
@krose422 You have been on fire lately :D Been looking at this for a while. Can you happen to do the same for some of the other components? I guess button would be an easy one to do

iRoachie:
@krose422 You have been on fire lately :D Been looking at this for a while. Can you happen to do the same for some of the other components? I guess button would be an easy one to do

src/list/ListItem.js
```diff
@@ -97,8 +99,14 @@ const ListItem = props => {
     <Component
       onLongPress={onLongPress}
       onPress={onPress}
+      disabled={disabled}
       underlayColor={underlayColor}
-      style={[styles.container, containerStyle && containerStyle]}
+      style={[
+        styles.container,
+        containerStyle && containerStyle,
+        disabled && { opacity: .5 },
```
iRoachie: Could you move this opacity into a style object at the bottom?

krose422:
@iRoachie - Moved that style object. And yeah, I'd love to do the same for others! I thought button already had a disabled option; does it not? Let me know what other components should allow disabling though and I can certainly jump on them. :) Thanks!

iRoachie:
Alright one more thing to pass the tests. Run `npm test -- -u` @krose422

kyler-hyuna:
Sweet thanks again! I think the FormInput doesn't have one

PR #671, opened 2017-10-25
removed avatar's view wrapper

cordosvictor:
The reason avatar's view-wrapper should be removed is because, as you can see in the picture below, the avatar is out of bounds (the height of avatar's wrapper being constrained to 34, property set in styles.avatar). This problem occurs only if you choose to render a custom component (Avatar in my case) whose height is bigger (e.g medium, large, xlarge). 

before

After I removed that wrapper, everything looks nice. (see picture below)

after


Monte9:
@cordosvictor good catch. 

Looking at the code, I see that that the avatar view style is the following:
```
avatar: {
  width: 34,
  height: 34,
}
```

I agree! We should remove this outer view. Also as part of this PR, can you delete the `styles.avatar` as well. thanks.

/cc @iRoachie

cordosvictor:
@iRoachie I made the changes @Monte9 suggested and also ran ```npm test --u```, but fails when commiting.

snapshot_result

cordosvictor:
@iRoachie I made it work with ```npm test -- -u```

src/list/ListItem.js
```diff
@@ -110,49 +110,41 @@ const ListItem = props => {
...
+                <View>
+                  <Icon
+                    type={leftIcon.type}
+                    iconStyle={[styles.icon, leftIcon.style && leftIcon.style]}
+                    name={leftIcon.name}
+                    color={leftIcon.color || colors.grey4}
+                    size={leftIcon.size || 24}
+                  />
+                </View>
+              </LeftIconWrapper>
```
iRoachie: Forgot  to close the brackets

cordosvictor: @iRoachie, resolved.

iRoachie:
Thanks again @cordosvictor. Good first PR :)

cordosvictor:
Glad to be part of this, @iRoachie. You guys are doing a great job here.

PR #676, opened 2017-10-26
update bundle.js to not include references of view.proptypes

jayeszee:
-Fixes a critical bug that crashes all production application running react@16.0.0-beta.5 or higher due to references of View.propTypes


xavier-villelegier:
Hey @jayeszee, just to understand, how are you using the `bundle.js`? I mean, instead of the `npm` package?  🤔 I see that the build hasn't been updated for 5 months

jjzazuet:
@xavier-villelegier Since version `0.49`, React Native now forces everybody to access to reuse `View` proptype references under a new export object. I see that your sources already apply this fix, with a fallback for RN pre-0.49:

```
import { View, ViewPropTypes as RNViewPropTypes } from 'react-native';

const ViewPropTypes = RNViewPropTypes || View.propTypes;

export default ViewPropTypes;
```

For some reason that I still don't understand, every time an app requests a production bundle, the React Native packager opts to use your `bundle.js` file, which as you correctly pointed out has not been updated in a very long time, and thus results in incorrect references to `View.propTypes`.

I would suggest not including the bundle at all, instead suggesting the users to run the React Native Elements sources through the RN Metro bundler instead when building their apps. How exactly? I'm not sure because I do not have enough experience in configuring this part of the build process.

But this is just my gut feeling. Hope it helps. Thanks!

jjzazuet:
@xavier-villelegier Hey, sorry man, my bad. I got confused ad wrote incorrect statements. I'm just too tired to going through a bunch of separate RN components having t apply the same patch. I need some rest :P.

Anyway, the metro bundler (React Native's packager) is not looking at `bundle.js` at all. So please disregard what I wrote above. Still, looking further into React Native element's sources, I see you guys are using Webpack to produce `bundle.js`. Not exactly sure why this is being done. Have you guys found a way to use RN together with Webpack?

Thanks again!!

xavier-villelegier:
To be honest, @binoy14 did the `bundle.js` and I don't feel like I have enough experience on that too… Maybe we should get rid of this `bundle.js`, as this seems to create more issue than it solves.

Monte9:
@xavier-villelegier I think I would be ok with that. I was pretty sure we were generating new bundle each release as part of Travis CI, but surprising it's on in our Ci. weird. 

Here's our Travis CI config if you are interested to check it out: https://travis-ci.org/react-native-training/react-native-elements/jobs/295222389/config

So in the case that we have to manually generate the bundles, we should consider getting rif of it altogether as it doesn't seem much of a purpose now.

Monte9:
@jayeszee we have decided to remove the `bundle.js` file as it doesn't serve a specific purpose right now. 

here's the PR that removes it. https://github.com/react-native-training/react-native-elements/pull/687 Thanks @xavier-villelegier! 👍

PR #682, opened 2017-10-31
fix(*): Spread passed props first to prevent computed props rewrite

### LINKED ISSUE #679: Issues with styled-components
This issue seems to affect a lot of the components provided by `react-native-elements`. Here's an example with `<ListItem />`:

If I use styled-components with a `<ListItem/>`, the default container style isn't used anymore.

Here's how the styled component is created:
```js
const StyledListItem = styled(ListItem).attrs({
  titleStyle: { /* ... */ },
})``;
```

Here's the result when using `<StyledListItem>` for my first element in the list:
screen shot 2017-10-27 at 6 15 01 pm

As you may see, the bottom border defined as default is lost.

The problem is located here: https://github.com/react-native-training/react-native-elements/blob/master/src/list/ListItem.js#L97-L103

When using styled-components, `attributes` gets automatically passed this value:
```
{ style: [ 540, undefined ],
  innerRef: [Function],
  children: undefined }
```

so the default style gets overrided.

### Solution

An obvious solution would be to spread `attributes` first, then define your own props, as you definitely don't want them to be overridden (or do you?)

```js
    <Component
      {...attributes}
      onLongPress={onLongPress}
      onPress={onPress}
      underlayColor={underlayColor}
      style={[styles.container, containerStyle && containerStyle]}
    >
```

I'd happily work on a PR for this, but i'd like to make sure that it's not done on purpose on the first place.

machour:
Fix #679

### DIFF
--- a/src/avatar/Avatar.js
+++ b/src/avatar/Avatar.js
@@ -195,6 +195,7 @@ const Avatar = props => {
 
   return (
     <Component
+      {...attributes}
       onPress={onPress}
       onLongPress={onLongPress}
       activeOpacity={activeOpacity}
@@ -203,7 +204,6 @@ const Avatar = props => {
         rounded && { borderRadius: width / 2 },
         containerStyle && containerStyle,
       ]}
-      {...attributes}
     >
       <View
         style={[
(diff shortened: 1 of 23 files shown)


machour:
Made an oopsy with prettier, will re-open new PR for clarity :)

PR #687, opened 2017-11-01
Get rid of bundle.js

xavier-villelegier:
Remove `bundle.js` to avoid weird issues in production


Monte9:
Just for the record, if we ever need to generate the `bundle.js` file, we could just run `yarn build` and commit those files.

PR #689, opened 2017-11-01
Remove deprecated components - v0.18

Monte9:
### Remove:
- [x] Side Menu
- [x] Grid
- [x] Row
- [x] Column
- [x] SwipeDeck
- [x] Tabs


Monte9:
I guess just skim the PR and see if you find anything out of place. The only way really "test" this is to merge into `next` and then see if anything in the example app breaks.

Since it's mostly components that's not used in the example app (except swipedeck) as long as the example app compiles and renders fine we are good aka removing these components doesn't break any config stuff for RNE. 

@iRoachie feel free to merge this PR once you have review it. thanks

Issue #659, opened 2017-10-17
Potential issue with font size normalization on Google Pixel


Nickersoft:
### Description

**1. Explain what you did**
I ported over the font size normalization code for this library to my own app.

**2. What you expected to happen**
Correctly normalize my fonts on iOS and Android

**3. What actually happens**
iOS is fine, running on my device (2016 Google Pixel) results in impossibly small font sizes.

**Also please include the version of `react-native-elements` that you are using.**
I'm not.. just your code :)

### Solution

So I actually found the cause of the problem. The PixelRatio output on my device is 2.6, and the conditionals found in `normalizeText.js` only include whole numbers. As a result, the original fontSize was just being returned every time. I changed the code to this and it works:

```javascript
if (pixelRatio >= 2 && pixelRatio < 3) {
        // iphone 5s and older Androids
        if (deviceWidth < 360) {
            return size * 0.95;
        }
        // iphone 5
        if (deviceHeight < 667) {
            return size;
            // iphone 6-6s
        } else if (deviceHeight >= 667 && deviceHeight <= 735) {
            return size * 1.15;
        }
        // older phablets
        return size * 1.25;
    }
    else if (pixelRatio >= 3 && pixelRatio < 3.5) {
        // catch Android font scaling on small machines
        // where pixel ratio / font scale ratio => 3:3
        if (deviceWidth <= 360) {
            return size;
        }
        // Catch other weird android width sizings
        if (deviceHeight < 667) {
            return size * 1.15;
            // catch in-between size Androids and scale font up
            // a tad but not too much
        }
        if (deviceHeight >= 667 && deviceHeight <= 735) {
            return size * 1.2;
        }
        // catch larger devices
        // ie iphone 6s plus / 7 plus / mi note 等等
        return size * 1.27;
    }
    else if (pixelRatio >= 3.5) {
        // catch Android font scaling on small machines
        // where pixel ratio / font scale ratio => 3:3
        if (deviceWidth <= 360) {
            return size;
            // Catch other smaller android height sizings
        }
        if (deviceHeight < 667) {
            return size * 1.2;
            // catch in-between size Androids and scale font up
            // a tad but not too much
        }
        if (deviceHeight >= 667 && deviceHeight <= 735) {
            return size * 1.25;
        }
        // catch larger phablet devices
        return size * 1.4;
    }
    // if older device ie pixelRatio !== 2 || 3 || 3.5
    else return size;
```

[...]

PR #690, opened 2017-11-01
Corrected normalizeText for certain Android devices

Nickersoft:
As per [this issue](https://github.com/react-native-training/react-native-elements/issues/659), a bug was found in which certain devices will return the pixel ratio as a float, breaking the font size normalization function. This PR fixes that.


Monte9:
Thanks for the quick turnaround on this PR @Nickersoft! 💯

PR #694, opened 2017-11-03
feat(Search): add 'onClearText' callback

wilomgfx:
I found i needed to do some specific stuff to my app's state when clearing the Search input.
Figured it would be useful to have it done via a callback, just like the with onChangeText callback.


Monte9:
@wilomgfx thanks for taking the time out to submit this PR! If this is something that is useful to you, chances are it might be useful to others as well.

This will be released in `v0.18.1`. Stay tuned.

PR #697, opened 2017-11-04
ListItem v1

martinezguillaume:
This is a first shot of the new ListItem component for the v1.0.
Refactored and simplified.
## Props:
```js
  containerStyle: ViewPropTypes.style,
  component: PropTypes.element,
  onPress: PropTypes.func,
  title: PropTypes.node,
  subtitle: PropTypes.node,
  titleProps: PropTypes.object,
  subtitleProps: PropTypes.object,
  icon: PropTypes.object,
  avatar: PropTypes.object,
  leftElement: PropTypes.element,
  rightElement: PropTypes.element,
  rightTitle: PropTypes.node,
  rightTitleProps: PropTypes.object,
  textInputProps: PropTypes.object,
  switchProps: PropTypes.object,
  checkBoxProps: PropTypes.object,
  badgeProps: PropTypes.object,
  disclosure: PropTypes.bool,
  checkmark: PropTypes.bool,
```

### before
```js
            <ListItem
              roundAvatar
              avatar={{ uri: l.avatar_url }}
              onPress={log}
              title={l.name}
              subtitle={l.subtitle}
            />
```
### after
```js
           <ListItem
              avatar={<Avatar rounded source={{ uri: l.avatar_url }} />}
              onPress={log}
              title={l.name}
              subtitle={l.subtitle}
            />
```
or
```js
           <ListItem
              avatar={{ rounded source={ uri: l.avatar_url }}
              onPress={log}
              title={l.name}
              subtitle={l.subtitle}
            />
```

### before
```js
              <ListItem
                leftIcon={{ name: 'user-circle-o', type: 'font-awesome', style: {color: 'blue'} }}
                title={l.name}
                titleStyle={{color: 'red'}}
                subtitle={l.subtitle}
                rightTitle='11:00am'
                rightTitleStyle={{color: 'green'}}
              />
```
### after
```js
            <ListItem
              leftElement={
                <Icon
                  name="user-circle-o"
                  type="font-awesome"
                  color="blue"
                  size={40}
                />
              }
              title={l.name}
              titleProps={{ style: { color: 'red' } }}
              subtitle={l.subtitle}
              rightTitle="11:00am"
              rightTitleProps={{ style: { color: 'green' } }}
            />
```

### before
```js
          <ListItem
            title={'Name w/ placeholder'}
            textInput={true}
            textInputPlaceholder={'John Doe'}
          />
```
### after
```js
          <ListItem
            onPress={log}
            title={'Name w/ placeholder'}
            textInputProps={{ placeholder: 'John Doe' }}
          />
```

### before
```js
          <ListItem
            title="RNE is the best"
            hideChevron={false}
          />
```

### after
```js
          <ListItem
            title="RNE is the best"
            disclosure
          />
```

## Things missing

- [x] Badge support
- [x] Switch button support
- [ ] CheckBox support


iRoachie:
Maybe slightly confusing to represent the leftElement as avatar since it can hold any element the user chooses

martinezguillaume:
So leftElement is a better name ?

Monte9:
@martinezguillaume this is a good start. But it think for `listitem v1` it's important to make sure that we are supporting basically anything the user might want to do with a ListItem and the API is very intuitive. 

Check this screenshot on iOS for example. 
screen shot 2017-11-04 at 11 34 27 am

I think the it makes sense to go with `leftElement` which is an object (like you have above with icon) so it's up to the user to render a Icon or Avatar or something else.
And on the right, the user should be able to render multiple components just by passing `Badge` component, `Switch` component, `Chevron` etc.. and RNE should contruct the `rightContainer` and add these component to it by default with `justifyContent: "flex-end"`

Does that make sense? I am def open to feedback. @johot what are your thoughts on this?

Btw, great clean up of the ListItem component @martinezguillaume! 👍

martinezguillaume:
We can do both ! Because there are component that need default style to work properly (TextInput for example that need flex: 1 to show properly). So maybe we can make basic component available by props (textInputProps, switchProps, sliderProps) in a classic layout (like the screenshot above). And if the user want something really custom, he can do it anyway with a rightElement.
Same thing with the leftElement.
And the code will be understandable and simple, I'll take care of that ahah

johot:
@martinezguillaume @Monte9 If you check my suggested design document I also suggested that we have a number of basic "prebuilt" inner components and then a left, middle and right slot for custom content.

In my design doc I suggested the following:

"Included components"
---------------------
* text
* textInput
* subtitleText
* switch (true/false) + style props
* accessory (with prebuilt icons like disclosure arrow, checkmark etc)

Right left and middle "slot"
---------------------
iconComponent (left slot)
contentComponent (middle slot)
accessoryComponent (right slot)

I just used names similair to those used by Apple. But maybe names such as left, right etc makes more sense for existing RNE users.

I know I haven't had time to create a PR for this yet, but here is my latest code when doing my POC. If you want to incorporate this code right away and don't wait for my PR you could use this gist I just made:

https://gist.github.com/johot/436176963d9a5394af4c522464cd3a2a.js

Monte9:
This PR will need to be recreated since the /example app has been deleted from the master branch and now lives in it's own repo here: https://github.com/react-native-training/react-native-elements-app#get-started

I suggest cloning that repo and using the submodule to create a PR on the `react-native-elements-app` repo as well as create a new branch and submit this PR from within the submodule for `react-naive-elements`.

Hope this is not too confusing. Let me know if you have any questions.

src/list/ListItem.js
```diff
@@ -158,4 +130,13 @@ const Disclosure =
     <MaterialIcons name="keyboard-arrow-right" style={styles.disclosure} />
   );
 
+const renderNode = (content, props, style) =>
```
iRoachie: Brilliant!

martinezguillaume:
Heyyy, I close this one in favour of https://github.com/react-native-training/react-native-elements/pull/866

This branch was a mess, it was not updating with upstream as expected 😭 I recreated it !

PR #703, opened 2017-11-06
Fix Header width on rotation

### LINKED ISSUE #701: Header doesn't render properly on screen rotation
Since update 0.18.0 the header doesn't render properly on rotation from portrait to landscape and vice versa. This was working fine on 0.17.0.

### 0.17.0 
- - -
**Portrait**

screenshot_1509957908

**Landscape**

screenshot_1509957912

### 0.18.0 
- - - 
**Portrait**

screenshot_1509957982

**Landscape**

screenshot_1509957985

### Current fix

 ```javascript
outerContainerStyles={{ width: '100%' }}
```

### P.S. If you intentionally made it this way, feel free to close the issue.

### PR DESCRIPTION
**Pull request by denisdimitrov**
- Header had strange behavior on rotation and it wasn't being rendered properly.

### Bug preview
screenshot_1509957985

### Fix preview
screenshot_1509963618

Fixes #701

### DIFF
--- a/src/header/Header.js
+++ b/src/header/Header.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { StatusBar, StyleSheet, View, Dimensions } from 'react-native';
+import { StatusBar, StyleSheet, View } from 'react-native';
 import isEmpty from 'lodash.isempty';
 import DummyNavButton from './DummyNavButton';
 import NavButton from './NavButton';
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by iRoachie** [comment]
How would this fix it? Seems like the width would be determined by the styles on the container this is placed in?

**Comment by xavier-villelegier** [comment]
Ah yes did you let ``containerStyle={{ width: '100%' }}`` for your screenshots?

**Comment by denisdimitrov** [comment]
The problem is that you are using react-native 0.41.2 and 'width' accepts only numbers.
Since 0.42.0 you can pass strings in the width style. Here is the issue:
https://github.com/styled-components/styled-components/issues/456

In this case if I pass 'width: 100%' your tests will fail and I'm not sure if you will be able to merge this at all. So my only option is as @iRoachie already said is to remove the width completely and let the parent decide the width which by default should be 100%.

**P.S.** This is still working and I already tested it on my personal project. You can checkout the PR and try it yourself.

**Optional** - You can update your react-native to at least 0.42.0 and then you can apply the fix by adding ```width: '100%'```

**Comment by xavier-villelegier** [comment]
@denisdimitrov Yes we're trying to avoid as long as possible to use percentage, to be compatible with RN < 0.42. Today, we do not have any percentage dimensions.

@iRoachie I can confirm that it works with a code as simple as this:
```jsx
  render() {
    return (
        <Header
          leftComponent={{ icon: 'menu' }}
          centerComponent={{ text: 'MY TITLE' }}
          rightComponent={{ icon: 'home' }}
        />
      </View>
    )
  }
}

  container: {
    paddingTop: Constants.statusBarHeight,
    flex: 1,
    backgroundColor: '#fff',
  },
})
```

|On @denisdimitrov PR|On RNE `0.18.2`|
|----------------------|------------------|
|simulator screen shot - iphone 7 - 2017-11-09 at 14 29 48|simulator screen shot - iphone 7 - 2017-11-09 at 14 32 14|

PR #760, opened 2017-12-06
style and theme example

Gregor1971:
This PR adds a second example project

This example demonstrates a method to theme and style RNE components much like styled-components does.  I took this from @necolas : (https://github.com/necolas/react-native-web/blob/master/docs/guides/advanced.md#use-as-a-library-framework) and changed it to allow all props to be manipulated (not just style), hence the name “proppedComponents”

In my opinion, the best aspect of this technique is that the RNE components do not require any modification.

The example is very basic, but it shows how an RNE component can be wrapped, via “proppedComponent”, to build reusable components with set styles.

A ThemeProvider is included, which provides on-the-fly switching of themes.  It does use React context, which most providers do; but this one puts a function (getTheme) in the context, instead of the theme itself.  This makes switching possible and safe.  (again, stolen from @necolas).

When wrapped in “themedProppedComponent”, a wrapped component will be passed a theme object as a prop, which it then uses to set dynamic styles.

I see that the alternative to this approach is already underway in the “theming_v1” branch.  These changes expect a “context.theme” object in each component, requiring a certain provider and tying RNE to this decision.  It's already headed down a path which prohibits switching themes on-the-fly.

Reasons I recommend this approach:
- The theme and styling mechanism is separate from the component source.
- Can be used to change all props via theme, not just style.  It makes sense to continue to provide props like “color” for the Icon component, for example.  Who knows, maybe somebody will need a function in their theme - this would allow it.
- The utilities included handle context safely, according to articles like this:   https://medium.com/react-ecosystem/how-to-handle-react-context-a7592dfdcbc  They also address the problems discussed in this: https://medium.com/@mweststrate/how-to-safely-use-react-context-b7e343eff076 but in a simpler manner.
- Little issues along the way, like “that isn’t part of the theme”, won’t require changes to the component source, as long as “that” is exposed via component props.  The components already do a good job of exposing the styles of inner elements.
- RNE won’t be locked into a choice.  Since it’s separate, users could use styled-components or glamourous-native right now.  If RNE bakes the themes in, they’re stuck with what RNE gives them.  A new styling library could pop up, and users could take advantage.
- react-native StyleSheets would still be used for most styles.  It’s likely that styling and themes will only ever touch a small portion of the styles in there now.  This maintains good performance.  Benchmarks for web:  https://github.com/necolas/react-native-web/blob/master/benchmarks/README.md#performance 

A possible plan forward:

1. this basic example
2. a more comprehensive example
  - really show off what can be done
3. fully-themed example
  - match common colors / sizes / etc to build a small theme object schema
  - wrap each component, applying the theme
  - bundle in a separate package, with named exports that match RNE’s.  this makes switching over to themes just a matter of changing imports
4. pre-defined themes
  - provide popular themes like:
    - flat
    - material design
    - native, based on current platform
  - light & dark variations of each
  - again, in a separate package

Happy to discuss


xavier-villelegier:
@Gregor1971 I really like this approach!! 
We could make each RNE component a `themedProppedComponent` and handle the theme in the context. This way, the users could just use the `ThemeProvider` with their big object theme to customize all the RNE components just once. Each component could use in priority the styles in the props, and the context theme as a fallback, to allow the user to override the theme if needed.

But if I get it correctly, your point is to provide theme without modifying the base components, so without making them a `themedProppedComponent`, as I proposed above. How would you do that? By creating a whole new set of `themedProppedComponent`s for each theme?

Gregor1971:
@xavier-villelegier  My goal is to try to minimize changes to existing source while providing the most flexibility.

I’m just guessing here, but let’s say the user base breaks down like this (groups named to make easy references):

- 80% (the Mob): want a simple theme implementation where they decide colors and the list of variants
- 10% (the Choosy): want the simple implementation above, but with “just a few small additions”
- 5% (the Artisans): want to do their own thing with styles and themes, like use glamourous or sass or whatever.  Or they’re afraid of react context and want to supply theme differently, say via Redux or MobX
- 3% (the Pickers): just want to grab one component from RNE, and would rather not be forced to use our ThemeProvider
- 2% (the Hasty) aren’t going to style or theme at all

To show what I’m suggesting, let’s use the SearchBar component as an example:

The vanilla component retains its current StyleSheet:
```js
 container: {
   borderTopWidth: 1,
   borderBottomWidth: 1,
   borderBottomColor: '#000',
   borderTopColor: '#000',
   backgroundColor: colors.grey0,
 },
 containerLight: {
   backgroundColor: colors.grey5,
   borderTopColor: '#e1e1e1',
   borderBottomColor: '#e1e1e1',
 },
 icon: {
   backgroundColor: 'transparent',
   position: 'absolute',
   top: 15.5,
   ...Platform.select({
     android: {
       top: 20,
     },
     web: {
       top: 19,
     },
   }),
 },
 loadingIcon: {
   backgroundColor: 'transparent',
   position: 'absolute',
   right: 16,
   top: 13,
   ...Platform.select({
     android: {
       top: 18,
     },
     web: {
       top: 18,
     },
   }),
 },
 input: {
   paddingLeft: 26,
   paddingRight: 19,
   margin: 8,
   borderRadius: 3,
   overflow: 'hidden',
   backgroundColor: colors.searchBg,
   fontSize: normalize(14),
   color: colors.grey3,
   height: 40,
   ...Platform.select({
     ios: {
       height: 30,
     },
     android: {
       borderWidth: 0,
     },
   }),
 },
 inputLight: {
   backgroundColor: colors.grey4,
 },
 searchIcon: {
   left: 16,
 },
 clearIcon: {
   right: 16,
 },
});
```
To make the Masses happy, we add wrapped components to a different package that makes RNE a dependency or peerDependency

The wrapped SearchBar would look like this:
```js
const ThemedSearchBar = themedProppedComponent(SearchBar, props => {
 return {
   containerStyle: {
       backgroundColor: props.theme.headerBackground
   },
   inputStyle: {
       backgroundColor: props.theme.secondaryColor,
       color: props.theme.secondaryTextColor,
       borderRadius: props.theme.smallRadius
   }
 };
});
export { ThemedSearchBar as SearchBar };
```
Each wrapper would be small and simple.

The theme object would just look like this:
```js
const simpleThemes = {
 light: {
   pageColor: "c",
   primaryColor: "c",
   highlightColor: "c",
   secondaryhighlightColor: "c",
   headerBackground: "c",
   textOverPageColor: "c",
   textOverPrimaryColor: "c",
   textOverHighlightColor: "c",
   textOverSecondaryhighlightColor: "c",
   textOverHeaderBackground: "c",
   largeRadius: "n",
   smallRadius: "n"
 },
 dark: {...theDarkVersion}
};
```
Using the basic themes would simply require a change from:
```js
import { SearchBar } from "react-native-elements";
```
to:
```js
import { SearchBar } from "react-native-elements-simpleTheme";
```
(in this example, users can’t even override the theme with props)

We wouldn’t need to pass ugly stuff in the theme, like “overflow: hidden”; it’s in the StyleSheet.

The Choosy would take this, fork it, and add an extra assignment or two.

The rest would just use the vanilla, un-themed components.

Eventually, we (or even the community) could add more complicated themes and wrappers, as separate packages.  Since the themes can change styles and props, we could do some complex stuff, like adding transition animations - without touching vanilla component code.  (I have to verify that, but I think supplying an Animated.Value to a prop which sets a nested element’s style and animating that would work)

sorry, this got long... and I didn't need to name the groups :smile:

Monte9:
Hey @Gregor1971! Firstly thanks for the PR. 👍 I apologize for my delayed response to it.

I love the level of detail & explanation on this PR. 👏 And thanks for taking the lead on themeing. This def would be a great addition to RNE.

Here are my thoughts and a few suggestions:

> In my opinion, the best aspect of this technique is that the RNE components do not require any modification.

💯 I especially like it. How it's not directly dependent on the RNE source and yet can be easily wrapped around it as an additional package/layer. I see you suggested using it through another package such as `import { SearchBar } from "react-native-elements-simpleTheme";`.

Although can I propose that we instead keep it within RNE and just export it as a style component? like `import { StyledSearchBar } from "react-native-elements";`. That way the user doesn't have to install an additional package. 


> pre-defined themes
provide popular themes like:
flat
material design
native, based on the current platform
light & dark variations of each
again, in a separate package

This sounds like a great idea but seems like a lot of work. I think a good POC/ useful MVP would just be to allow the RNE theme to support light/dark themes as this is prob the most common use case for using themeing. So let's aim for that, to begin with, and hopefully within the same RNE package as I described above. 

Lastly, I really like how you ended up creating a simple example app to demo this. Just fyi, for v1 we ended up moving the example app out of this repo and into it's own repo here: https://github.com/react-native-training/react-native-elements-app. So I was wondering if instead of creating a new example app for themes, if we can just add to the existing app so all the demo & components are in 1 app for the users to see.

So just to recap:
- support only light/dark theme to begin with
- create styled component within RNE and just export it as `StyledSearchBar`
- move this app example into the main expo example app

Let me know what you think of the above points ^^ and we can figure out a time to ship this PR. :shipit:. Thanks for taking the time to create this PR @Gregor1971 👏 💯 

/cc @iRoachie @xavier-villelegier would you guys agree with my suggestions? Feel free to pitch any other ideas you might have.

Gregor1971:
@Monte9,

Totally fine with adding styled versions to the core RNE package.  How about a PR after web support is in?  I'll also submit a PR for adding the example to the react-native-elements-app.  How about an additional page?

Agreed, light/dark to begin with.  My suggestions for the very-specific high-effort themes was just to argue the case for leaving the base components untied to the theme - "unopinionated", in the parlance of our times.

Monte9:
@Gregor1971 that's sounds great. 

Let's leave this PR open and follow-up on these things once RNW PR is released in `beta-3` this evening.

- How about a PR after web support is in? 
Sounds good.

- I'll also submit a PR for adding the example to the react-native-elements-app. How about an additional page?
That would be perfect! Thanks in advance.

PR #763, opened 2017-12-06
Fixes failed prop type/style #740

### LINKED ISSUE #740: [Same as 668] Tile: Warning: Failed prop type: Invalid props.style key 'resizeMode' supplied to 'View'
### Description
I get the same warning as listed on this issue: https://github.com/react-native-training/react-native-elements/issues/668

```
Warning: Failed prop type: Invalid props.style key resizeMode supplied to View.
Bad object: {
"alignItems": "center",
"justifyContent": "center",
"resizeMode": "cover",
"backgroundColor": "#ffffff",
"flex": 2
}
...
```

Versions of the dependencies
```
"dependencies": {
	"react": "16.0.0",
	"react-native": "^0.50.4",
	"react-native-elements": "^0.18.4",
	"react-native-vector-icons": "^4.4.2",
	"react-navigation": "^1.0.0-beta.21"
},
```

You can clone my practice repo under that branch. I see the warning when I run it on an android simulator. 
https://github.com/Jun711/ReactNativePractice/tree/resizeModeWarning

josephch405:


src/tile/Tile.js
```diff
@@ -107,7 +104,7 @@ const Tile = props => {
           styles.imageContainer,
           imageContainerStyle && imageContainerStyle,
         ]}
-        imageStyle={styles.imageStyle}
+        imageStyle={{ resizeMode: 'cover' }}
```
xavier-villelegier: This doesn't change the result haha 😅 The issue here is that `resizeMode` needs to be a props of the `BackgroundImage`, and not a style anymore
```jsx
<BackgroundImage
  …
  resizeMode="cover"
/>
```

josephch405: Oops - my bad. Silly mistake. 😅

francisrod01: This modifications don't work to me.

Piropa:
Correction should not be made in Tile.js, since imageStyle is a correct property for the ImageBackground react-native's element. The problem is loceted in FeaturedTile.js, where in styles, the imageContainer has the resizeMode proerty when it shouldn't, Currently:

    ...
    imageContainer: {
      alignItems: 'center',
      justifyContent: 'center',
      resizeMode: 'cover',
      backgroundColor: '#ffffff',
      width,
      height,
    },
    ...

The proper solution would be to split this style in two, as follows:

    ...
    imageContainer: {
      alignItems: 'center',
      justifyContent: 'center',
      resizeMode: 'cover',
      backgroundColor: '#ffffff',
      width,
      height,
    },
    image: {
      resizeMode: 'cover'
    },
    ...

and then apply it to the BackgroundImage element:
    ...
      <BackgroundImage
        source={imageSrc}
        style={[
          styles.imageContainer,
          imageContainerStyle && imageContainerStyle,
        ]}
        imageStyle={[
          styles.image,
          imageStyle && imageStyle,
        ]}
      >
    ...

We should also capture the imageStyle from props at the begining:

const FeaturedTile = props => {
  const {
    ...
    imageContainerStyle,
    imageStyle,
    ...
  } = props;

and that's it!.

xavier-villelegier:
Hey @Piropa, as you can see, `FeaturedTile.js` was already fixed before in #787.
`resizeMode` is now a props of the React Native `Image` component, check out [the official doc](https://facebook.github.io/react-native/docs/image.html#resizemode). So we simply move out `resizeMode: cover` from style to props 👍

PR #773, opened 2017-12-10
Added to SocialIcon

alburdette619:
I added several icons to the `SocialIcon` element including `envelope` for email, `reddit-alien` for reddit, `stack-overflow`, `etsy`, and `steam`.  Also added to the docs to reflect the changes.

Additionally, I alphabetized the `colors`, props deconstruction, and docs type listings.  If this needs to be reverted for the merge I understand but I believe it makes it a better U/DX when lists get this long.


iRoachie:
This is such an awesome PR. Can you also add them into the example app? The file is here https://github.com/react-native-training/react-native-elements/blob/master/example/src/views/buttons_home.js

alburdette619:
Currently the `SocialIcon`s featured in the example app only consist of 10 of the possibilities.  Would you like me to add everything that's missing or just the 5 I added on another row?

alburdette619:
I went ahead and added them all.

iRoachie:
Fantastic job man! 🎉 Thanks again

PR #782, opened 2017-12-15
Adjusted Avatars Card and children Views

francisrod01:
The `height` prop in `Card` component were hiding the second View, limiting the View.
Style props in each child was also spared. This can be chest in the parent View.
This is more simple. I based of [Layout with flexbox](https://facebook.github.io/react-native/docs/flexbox.html)

emulator: `Genymotion`

before:

image

I just removed the `height` props and...

image

So.. I improved the adjustments and... (I use background props to see this modifications better)

image

### DIFF
--- a/example/src/views/lists_home.js
+++ b/example/src/views/lists_home.js
@@ -303,13 +303,19 @@ class Icons extends Component {
             containerStyle={{
               marginTop: 15,
               marginBottom: 15,
-              height: 230,
               paddingLeft: 10,
             }}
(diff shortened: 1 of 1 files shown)


xavier-villelegier:
Arrg, didn't see that `master` was the base branch. Please open your next PR's on the `next` branch, for this time I'll just revert on `master` and cherry-pick your commit of `next`

francisrod01:
Thanks, @xavier-villelegier ! :)

PR #783, opened 2017-12-16
imageProps declaration for Card component

budiadiono:
Added missing Typescript `imageProps` declaration for `Card` component.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -10,6 +10,7 @@ import {
     ViewStyle,
     TextStyle,
     Image,
+    ImageProperties as RNImageProperties,
     ImageStyle,
     ImageURISource,
     TouchableWithoutFeedbackProps,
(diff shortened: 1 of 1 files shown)


src/index.d.ts
```diff
@@ -506,6 +514,11 @@ export interface CardProps {
      * Add an image as the heading with the image prop
      */
     image?: ImageURISource;
+  
+    /**
+     * Optional properties to pass to the image if provided e.g "resizeMode"
+     */
+    imageProps?: ImageProperties;
```
iRoachie: Why not just `Partial<ImageProperties>`. Why extend a new interface?

budiadiono: Maybe some other components will/might need this partial `ImageProperties` it too? What do you think?

budiadiono: Ah, I must be dreaming :) You're right -- this could lead some mistake when someone thinks that this new `ImageProperties` is come from `react-native` itself -- I'll fix it now. Thanks for the review.

iRoachie: Sure thing 🙂

iRoachie:
Thanks again! 🎉

budiadiono:
You're welcome :)

PR #786, opened 2017-12-17
Adjusted Avatars Card and children Views

iRoachie:
The `height` prop in `Card` component were hiding the second View, limiting the View.
Style props in each child was also spared. This can be chest in the parent View.
This is more simple. I based of [Layout in flexbox](https://facebook.github.io/react-native/docs/flexbox.html)


iRoachie:
Correct base for https://github.com/react-native-training/react-native-elements/pull/782

PR #790, opened 2017-12-19
Allow specifying of selectedButtonStyle in button group

ceckobecko:
Greetings from Bulgaria,
Firstly - Thanks for the great framework.

I've found an issue with the button group and styling.

Since we want to allow the overwriting of styles for selected
button without the need of custom logic. This commit
allows that behavior by introducing a new property -
selectedButtonStyle.

Also include the change in the docs

### DIFF
--- a/docs/API/button_group.md
+++ b/docs/API/button_group.md
@@ -78,7 +78,6 @@ render () {
 | containerStyle | inherited styling | object (style) | specify styling for main button container (optional) |
 | buttonStyle | inherited styling | object (style) | specify styling for button (optional) |
 | selectedButtonStyle | inherited styling | object (style) | specify styling for selected button (optional) |
-| selectedBackgroundColor | white | string | specify color for selected state of button (optional) |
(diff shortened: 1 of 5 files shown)


xavier-villelegier:
Hey @ceckobecko, greetings from RNE 👋😄
Thanks for the PR!
Maybe we can avoid having `selectedBackgroundColor` + `selectedButtonStyle` by factoring a bit. @iRoachie what do you think?

iRoachie:
2nd that lol. Think we can drop selected background color and just use selected button style

ceckobecko:
Will gladly drop it. with your permission :)

iRoachie:
Sure. I think `selectedButtonStyle` is a lot more flexible, although this will be a breaking change. Cool with you @xavier-villelegier ?

xavier-villelegier:
Sounds good to me! 🤙

iRoachie:
Thanks again!

PR #793, opened 2017-12-21
Add disabledTextStyle property to Button

arturowczarek:


arturowczarek:
@xavier-villelegier Now should be better.

iRoachie:
Oh my you even updated the typescript definitions ❤️

iRoachie:
Thank you so much!

PR #808, opened 2018-01-07
[TS] Adjust TextProps to extend RN TextProperties

### LINKED ISSUE #807: [TS] Text missing RN Text props
Text component is missing default RN Text props such as `numberOfLines`

raduflp:
Fixes #807


iRoachie:
Thanks! Good catch

Issue #815, opened 2018-01-13
Raised styles on Button are applied to wrong View


hhg2288:
### Description

1. Explain what you did:
When you have a button `raised` and with some type of spacing styling (margin or padding), the shadow created by the `raised` prop is attached to the wrong view inside the button.

2. What you expected to happen
The border generated from the `raised` prop should be in the "colored" button, not in that View that happens to be the Button Implementation wrapper.

3. What actually happens
example:
```
<Button
            title="button"
            raised
            buttonStyle={{ margin: 16 }}
          />
```
result:
![img_43c046e71b79-1](https://user-images.githubusercontent.com/725120/34900531-27875f82-f801-11e7-92ce-2cf5fbeb3b07.jpeg)

Also please include the version of `react-native-elements` that you are using.
**version: 0.19.0**

### Reproduction Steps and Sample Code

https://snack.expo.io/ryrxN0LVf

### Solution

I know how to fix it but I need some guide dealing with snapshots because the fix actually change the snapshots, so the fix did not pass..

I actually update the snapshot tests in my branch's fork. you can see it here: https://github.com/hhg2288/react-native-elements/commit/c8655012f0b70b3436bd564af0a581cd6c5e0fbf

I test it locally & It works correctly. If soeone can help reviewing this I can create the PR to fix it :)

UPDATE: PR -> https://github.com/react-native-training/react-native-elements/pull/816

iRoachie:
All components in react-native-elements have a container element around them. In your example by adding a margin to the `buttonStyle` prop, you are essentially creating a gap between the container and the element itself. What you probably actually want is the `containerStyle` prop, in which then applying a margin would make more since as it's the container that interacts with other elements.

steamedcotton:
I was having a similar issue and I think the prop name is `containerViewStyle` (not `containerStyle`).  Hope that helps.

iRoachie:
Thanks made the typo update

hhg2288:
@iRoachie oh ok!, so there is no fix then?
I'm assuming that if I'm using a `Button`, to change its styles, I need to add something to the `buttonStyle` prop, not to the `containerViewStyle` prop. In any case, those that shadow still needs to be applied to the container?

just let me know if I need to delete the issue & the PR :)

thanks for the answer :)

hhg2288:
this is my humble opinion...

the fact that all the elements in this library has a container view around them, is an implementation detail. users may or may not need to know or care about it.

I see two solutions here:
[...]

PR #816, opened 2018-01-13
fix(Button): add raised styles into the correct View

hhg2288:
now the Styles added but the `raised` prop are added to the properly View inside Button. this avoids to have weird visuals when you apply some styles with `buttonStyle`.

solves #815 

here you can see the error: https://snack.expo.io/ryrxN0LVf

Visual Representation of the Solution:
button change

### DIFF
--- a/src/buttons/Button.js
+++ b/src/buttons/Button.js
@@ -157,7 +157,6 @@ const Button = props => {
     <View
       style={[
         styles.container,
-        raised && styles.raised,
         containerViewStyle,
         borderRadius && { borderRadius },
       ]}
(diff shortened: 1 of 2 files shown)


iRoachie:
Replied to your original issue, https://github.com/react-native-training/react-native-elements/issues/815#issuecomment-357480088

Monte9:
Hey @hhg2288! Thanks for the PR. 

As you might know (or might not), for our v1 release we ended up re-writing the button component from scratch. And we just release `v1.0.0-beta` hence there are merge conflicts on this PR. 

I would suggest that you give the new Button API a try and let us know if you still have trouble with it. For now use this as a reference for how to use the new API: https://github.com/react-native-training/react-native-elements-app/blob/master/src/views/buttons_home.js#L13

If you are curious, you can read more about the v1 release on our readme [here](https://github.com/react-native-training/react-native-elements#v10-roadmap---)

hhg2288:
hey @Monte9 !
awesome, will check the beta version then... 

thanks!

PR #834, opened 2018-01-23
Export FormLabel, FormInput and FormValidationMessage

sobrinho:


Nickman87:
Yes please! maybe also add the FormValidationMessage component while you're at it?

sobrinho:
Sure, just a sec.

sobrinho:
Done.

xavier-villelegier:
@Monte9 is there a reason you removed `FormLabel` and `FormValidationMessage` exports ?

Nickman87:
They were removed in this commit: https://github.com/react-native-training/react-native-elements/commit/8da9a07eb5a75961d232ba48b8fdd1cacb569745#diff-1fdf421c05c1140f6d71444ea2b27638

Probably because the input was refactored into a dedicated component? But no alternatives for the labels were added there.

Monte9:
Yes, I did end up removing it in `v1` in favor of the new Input component. As far as I know, you should be able to do almost everything with the Input component. The only difference is that the `FormValidationMessage` is not baked into the Input component as it is the most common use case.

Let me know if there is something missing from the Input component and we can talk about how best to go about achieving it.

/cc @xavier-villelegier we should consider deleting the FormInput folder from v1 as it's now deprecated.

Nickman87:
@Monte9 What about the FormLabel component? Is there an alternative to that or do we just style a Text component?

Nickman87:
@Monte9 Thanks for the explanation 👍

PR #836, opened 2018-01-23
Added the ability to use the git repo as npm package directly

npinochet:
Adding .git to .npmignore makes it so that npm install doesn't error when using the git repo in npm's package.json.


iRoachie:
Hey @npinochet, there's no need to add `.git` to the npmignore to use it directly from GitHub. Using `npm install --save https://github.com/react-native-training/react-native-elements.git` works as expected.

npinochet:
Yeah, but doing that error when using the latest versions of npm. Related Issues:
https://github.com/npm/npm/issues/18683
https://github.com/npm/npm/issues/10372
https://stackoverflow.com/questions/27356915/how-to-update-npm-modules-ignoring-a-git-repo

iRoachie:
I haven't encountered any problems using it @npinochet. I'm on the latest version on npm `v5.6.0`

PR #839, opened 2018-01-24
Pass activeOpacity to TouchableOpacity inside Tile

### LINKED ISSUE #479: <Tile> prop "activeOpacity " doesn't work
### Description
I use the <Tile> in my Demo, But I want to set activeOpacity to "0.95",but doesn't work.

There is my code:

``` 
<Tile imageSrc={require('../image/sea.jpg')}
                 activeOpacity={0.95}
                 onPress={() => alert('onPress')}
            />
```

I use the latest version of react-native-elements

lex111:
Fixes #479.

### DIFF
--- a/src/tile/Tile.js
+++ b/src/tile/Tile.js
@@ -96,6 +96,7 @@ const Tile = props => {
     <TouchableOpacity
       {...attributes}
       onPress={onPress}
+      activeOpacity={activeOpacity}
       style={[styles.container, containerStyle && containerStyle]}
     >
       <BackgroundImage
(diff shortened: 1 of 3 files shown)


xavier-villelegier:
Thanks for the fix @lex111 ! 👍 Could you add a test in `tile/tests/Tile.js` with a Tile with `activeOpacity` and update the snapshots by running `npm test -- -u` ? This way we are sure that it won't break on future updates 🙏

lex111:
@xavier-villelegier done :heavy_check_mark:

PR #848, opened 2018-01-26
Fix ListItem

MrFoxPro:
Fix error on empty rightTitle prop by adding `!!`


xavier-villelegier:
@MrFoxPro Did you test it ? I was saying that it might do the job but it was hypothetic ahah

MrFoxPro:
@xavier-villelegier, Do not worry, I found an error before I saw your comment! Ive just tested and found that everything works

PR #850, opened 2018-01-26
Separate checkbox single and checkbox with text styling

iRoachie:
Currently right now making a checkbox without text produces an ugly spacing and border around the component. This is caused because the styling for the text style is being applied.

screen shot 2018-01-26 at 12 44 03 pm

This PR allows the checkbox to be ~an independent grown man~ able to display differently depending on if there's a title.

screen shot 2018-01-26 at 12 44 56 pm


Monte9:
I like it! It's such details that make a library ~into a living legend~ great! Thanks for doing it @iRoachie 🎉

Issue #826, opened 2018-01-21
Beta version throws error `Module does not exists 'expo'`


deadcoder0904:
Tried running `react-native init demo && yarn add react-native-elements@beta` & the default application did not work. It throws the following error -

```
Unable to resolve module `expo` from `~/demo`: Module does not exist in the module map
This might be related to https://github.com/facebook/react-native/issues/4968
To resolve try the following:
1. Clear watchman watches: `watchman watch-del-all`.
2. Delete the `node_modules` folder: `rm -rf node_modules && npm install`.
3. Reset Metro Bundler cache: `rm -rf $TMPDIR/react-*` or `npm start -- --reset-cache`.  
4. Remove haste cache: `rm -rf $TMPDIR/haste-map-react-native-packager-*`.
```

It throws error because of [this line in V1 branch](https://github.com/react-native-training/react-native-elements/blob/v1/src/buttons/Button.js#L45) as it uses Expo & I'm not using it.

Monte9:
Ah shit. I knew that would cause a problem. I am going to copy @xavier-villelegier & @martinezguillaume on this. 

What we are trying to do there is to make liner gradient available to the user by using Expo's LinearGradient. But unfortunately `require('expo')` throws an error if not found instead of just returning undefined. 

If you or anyone else has suggestions on how to check if a module/library exists without breaking the app, please let me know.

@deadcoder0904 Let's leave this issue open for a bit, and if we don't find a fix we can get rid of that line and figure out a workaround so `@beta` works with `react native init`

deadcoder0904:
Naah, checkout https://github.com/facebook/react-native/issues/4968 for that issue

martinezguillaume:
Hey ! I can’t see anything related to the issue in your link 😕

What we are trying to do is to detect if the node modules (expo here) exists in your project. @xavier-villelegier runs into a lot of issues, maybe à possible workaround exists now, we have to check

deadcoder0904:
Ohh I thought the error which I've shown has that above link so I thought those 2 might be related.

nimo007:
Any update? I have the same problem

Hiro5409:
+1

xavier-villelegier:
@deadcoder0904 @nimo007 @Hiro5409 
I've tested on a `react-native init` project from scratch, and **yes, it crashes** without passing any `linearGradientProps`.

[...]

PR #854, opened 2018-01-28
Beta version throws error Module does not exists 'expo'

Ariel-Rodriguez:
## Issue
https://github.com/react-native-training/react-native-elements/issues/826

## Solution (proposal)
  As far i know, react-native-elements is main oriented for Expo projects. Having said that i propose a fallback for projects being created by `react-native init` and expo packages are not mandatory.

 The fallback consist to transpile /src into /lib folder, this process will yank out all dead code.
 With the power of removing dead code on prepublish, the final users will get the opportunity to install react-native-elemnets and choose the sources at their convenient case.

## Changes
- babelrc was updated in order to inject environment variables and remove dead code
- Add packages.json script "prepublish" in order to transpile /src into /lib and pack it all to NPM when publishing it.
- Added to git ignore lib and build folders.

## Test
- Checkout to fix/expo repository
- Perform a clean install
 ```
rm -rf $TMPDIR/react-* && rm -rf $TMPDIR/haste-map-react-native-packager-* && watchman watch-del-all && rm -rf node_modules && yarn
```
- Then you have tow ways to try it out, (A) from NPM (desired) or Local files (B)

A)
- You may change package.json version and publish to NPM in order to verify if prepublish task does what is expected. So then you will be able to run 'yarn add react-native-elements@canary-revision-fix' and fetch the content.
 
B)
 - `npm run prepublish`
 - `cp -R ./lib   <PATH TO YOUR PROJECT>/node_modules/react-native-elements `

- Import the elements as follow

```
import { Avatar } from 'react-native-elements/lib'
```

- yarn start

**expected** Bundler compiles without error

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,12 @@
 {
-  "presets": ["react-native"]
+  "presets": ["react-native"],
+  "plugins": [
+    "transform-inline-environment-variables",
+    ["minify-dead-code-elimination", {
+      "optimizeRawSize": false,
+      "keepFnName": true,
+      "keepFnArgs": true,
+      "keepClassName": true
+    }]
+  ]
 }
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ site
 coverage
 jsconfig.json
 .vscode/
+lib
+build
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,1 @@
+node_modules
--- a/docs/Installation/default_installation.md
+++ b/docs/Installation/default_installation.md
@@ -21,3 +21,11 @@ yarn add react-native-elements
 ```
 npm i react-native-elements --save
 ```
+
+## Step 3: Implementation
+
+If your project does not use "expo" dependency, then you should import from /lib
+
+```javascript
+import { Button } from 'react-native-elements/lib'
+```
--- a/package.json
+++ b/package.json
@@ -16,9 +16,11 @@
   ],
   "scripts": {
     "build": "rm -rf build; webpack",
+    "noexpo": "rm -rf lib && EXPO_IS_DISABLED=true babel src --out-dir lib --copy-files",
     "test": "jest",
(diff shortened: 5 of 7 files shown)


xavier-villelegier:
Here are my thoughts:

This is a nice solution, I wouldn't have thought about it. However, maybe it's a bit overkill to precompile everything and have each file twice, with and without dead code, just for one feature that could be added in other ways (thinking about [this](https://github.com/react-native-training/react-native-elements/issues/826#issuecomment-360495235)) 🤔

Finally, I don't think RNE is Expo oriented. We're just trying to avoid any `react-native link` to keep Expo users, but not exclusively.

@Monte9 @iRoachie Let me know your thoughts about that

iRoachie:
It’s a solution, but i really don’t like the precompile thing. I’ve came across it on other libs outside react-native and never liked it. It makes using a lib directly from github basically impossible, since a precompile step is required. 

Because we use a `next` for features/fixes introduced since the last release, users have the ability to just change their version in the package.json and use the link to GitHub branch if they want to use a fix before it's released. This is a __very often use case in open source__, especially when you have a fix that hasn't been merged as yet and you want to just be able to use your fork, until the PR's been merged. 

With the above solution for the above use case we'd introduce having users to:
1. `cd to node_modules`
2. `npm i`
3. `Run compile script`

And then finally delete `node_modules/react-native-elements/node_modules` folder since if we leave it there then haste (the webpack of react native) will detect duplicate versions of  react-native and report errors to the user.

Really think it's too much overhead for how we currently use react-native-elements. I much more prefer [the other solution])https://github.com/react-native-training/react-native-elements/issues/826#issuecomment-360495235) by @xavier-villelegier

xavier-villelegier:
@Ariel-Rodriguez Thanks a lot for digging into this. We did this in a more simple way, but it's always good to know that this approach is possible ! Have a look at #857 for the actual fix.

PR #857, opened 2018-01-28
Add ButtonComponent props and linear gradient props

xavier-villelegier:


iRoachie:
Looks good, can you add the fields into the docs?

PR #858, opened 2018-01-28
feat(button-group): Adds selectMultiple feature

### LINKED ISSUE #713: Select multiple when using ButtonGroup
It seems that when I use the `ButtonGroup` component I can always only select one of the buttons.

Wouldn't a boolean property like `selectMultiple` make sense for cases where I don't want only a simple switch?

iRoachie:
Closes #713

Adds the functionality to select multiple buttons on the button group.

Example:
```jsx
state = {
  selected: [0]
};

render() {
  return (
      <ButtonGroup
        buttons={['Hey', 'Bop', 'Ha']}
        selectMultiple
        selectedIndexes={this.state.selected}
        onPress={selected => {
          this.setState({ selected });
        }}
      />
    </View>
  );
}
```

select-multiple

### DIFF
--- a/docs/API/button_group.md
+++ b/docs/API/button_group.md
@@ -72,6 +72,8 @@ render () {
 | prop | default | type | description |
 | ---- | ---- | ----| ---- |
 | selectedIndex | none | number | current selected index of array of buttons (required) |
+| selectMultiple | false | boolean | allows the user to select multiple buttons |
+| selectedIndexes | [] | array (number) | current selected indexes from the array of buttons |
 | onPress | none | function | method to update Button Group Index (required) |
 | buttons | none | array | array of buttons for component (required), if returning a component, must be an object with { element: componentName } |
 | component | TouchableHighlight | React Native Component | Choose other button component such as TouchableOpacity (optional) |
--- a/src/buttons/ButtonGroup.js
+++ b/src/buttons/ButtonGroup.js
@@ -18,6 +18,8 @@ const ButtonGroup = props => {
     buttons,
     onPress,
     selectedIndex,
+    selectedIndexes,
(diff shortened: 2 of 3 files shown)


Monte9:
Looks great! 💯 Total rockstar for adding an example & working gif in the PR!!  @iRoachie 🎉

NYamaguchi415:
Very new to javascript, so forgive me, but in the example code does it overwrite the state to equal the last selected index instead of creating an array of selected indexes?

iRoachie:
Here's the same code you made, but I pulled out the setState to a method. You can see the `alert` prints out an array of numbers https://snack.expo.io/Hy7rh_U8m

NYamaguchi415:
Ohhh thanks @iRoachie , I think it might be a version issue. I just realized prod v1.0.0 wasn't released, so if I want v1.0.0-beta do I need to switch over to yarn?

iRoachie:
Nope you don't have to use yarn! `npm install --save react-native-elements@beta`

NYamaguchi415:
Thanks so much for your help!

omrihaim:
Hey @iRoachie, thanks you and appreciate all your work in adding this useful feature. 

1. I think the feature is currently not working, as the code still require selectedIndex, and the current code don't really update the selection correctly:
const isSelected = selectedIndex === i || selectedIndexes.includes(i);
require selectedIndex and will get the value of selectIndex. Please correct me if I am wrong.
I suggest switching between statements and adding ():
isSelected = (selectedIndexes.includes(i) || selectedIndex === i);

2. @NYamaguchi415 was correct, as the onPress is always called with just one selected index and not the updated list. I've tried to locate the issue in the code, and haven't found it yet.

PR #895, opened 2018-02-09
[docs] Fix Button docs wrong title property

Pl4gue:
Previously, the button usage examples in the docs used the `text` property which led me personally to big confusion as using `text` doesn't show any button text.

The right property name is `title` instead of `text`.

I fixed this naming in this PR.

### DIFF
--- a/docs/API/buttons.md
+++ b/docs/API/buttons.md
@@ -7,10 +7,10 @@ import { Button } from 'react-native-elements';
 import Icon from 'react-native-vector-icons/FontAwesome';
 
 <Button
-  text='BUTTON'
+  title='BUTTON'
 />
 
-<Button
+<Button/
   icon={
     <Icon
       name='arrow-right'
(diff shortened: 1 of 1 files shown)


xavier-villelegier:
@Pl4gue The documentation is for `v1.0.0-beta2`, not for `v0.19.0`. In `v1.0.0-beta2` the `Button` component use a `text` props. This should be changed back to `title` in future betas by the way

Pl4gue:
But how does the user recognize this? @xavier-villelegier 

I personally got stuck for quite some time because of this...

xavier-villelegier:
@Pl4gue Yea sorry, that's not really clear, I agree. Unfortunately we don't have versioning for the documentation, so the docs are the latest version in development. If you have a doubt you can browse the code of a specific release [here](https://github.com/react-native-training/react-native-elements/tags), to check the components or the docs

Pl4gue:
Hey @xavier-villelegier <https://github.com/xavier-villelegier>,
does that mean that in future, title will be renamed text?

2018-02-09 19:13 GMT+01:00 Xavier Villelégier <notifications@github.com>:

xavier-villelegier:
@Pl4gue Not exactly, `title` was renamed to `text` in the `v1.0.0-beta` (don't know exactly why) but will likely be renamed back to `title` to stay consistent with RN conventions.

PR #935, opened 2018-02-22
Added Password Toggle View Option

Prashant-Pandey:
showPassword: Icon
hidePassword: Icon
It'll automatically hide and unhide the password
It's all drawn to Input Component
function works only if user setted the secureTextEntry
only Input File is editted


xavier-villelegier:
Hey @Prashant-Pandey, you need to base your branch on `next`, not `master`, as you can see there is a lot of files changed

If I understand correctly, your point is to be able to toggle the visibility of the password. I would personally use the API of the `Input` by doing something [like this snack example](https://snack.expo.io/BkoLOS2wz). I don't think it is necessary to add props to handle that internally, as it can be handle pretty easily by the user, using the simple API of the `Input`.
Moreover, people may want to handle the password visibility by their own (in redux for example), as shown in my snack example.

Thoughts about that @iRoachie @Monte9 ?

iRoachie:
Hey @Prashant-Pandey, it seems you based this PR off an old `next` branch since it includes content from before the new docs. 

You've got two options:
1. Rebase your PR from the current `next` branch to get rid of the merge conflicts. 
2 Create a new PR from a new branch. Make sure you create a branch from the `next` branch.

The second one is a lot easier 💯

iRoachie:
Definitely leaning on letting the user handle the state themselves. The default TextInput component from react-native handles secure fields by default so I don't think there's much benefit for adding an additional API over this

PR #945, opened 2018-02-25
ListItem v1 (Android & iOS)

martinezguillaume:


src/divider/Divider.js
```diff
@@ -1,21 +1,27 @@
...
   style: ViewPropTypes.style,
 };
 
 styles = StyleSheet.create({
   container: {
-    height: 1,
-    backgroundColor: colors.grey5,
+    ...Platform.select({
+      ios: {
+        height: 0.5,
```
ocarreterom: Should not the `height` be `StyleSheet.hairlineWidth`?
https://facebook.github.io/react-native/docs/stylesheet.html#hairlinewidth

ocarreterom: Should not the `height` be `StyleSheet.hairlineWidth`?
https://facebook.github.io/react-native/docs/stylesheet.html#hairlinewidth

martinezguillaume: Good point ! I'll use it !

Monte9:
here we goo.. 🤞

PR #915, opened 2018-02-17
v1.0.0-beta3

xavier-villelegier:
## Release notes

https://github.com/react-native-training/react-native-elements/releases/tag/untagged-f2563a88ab94e38617bc

PR #950, opened 2018-02-28
docs: Allow viewing `next` docs in version switch

iRoachie:
Allows to view the latest changes in docs.

screen shot 2018-02-28 at 7 08 20 am

### DIFF
--- a/website/pages/en/versions.js
+++ b/website/pages/en/versions.js
@@ -15,6 +15,45 @@ const CWD = process.cwd();
 const siteConfig = require(CWD + '/siteConfig.js');
 const versions = require(CWD + '/versions.json');
 
+class VersionItem extends React.Component {
+  render() {
+    const version = this.props.version;
+    const versionName = version === 'next' ? 'Master' : version;
+
+    const isCurrentVersion = this.props.currentVersion === version;
+    const isNext = version === 'next';
+
+    const documentationLink = (
+      <a
+        href={
+          this.props.baseUrl +
+          'docs/' +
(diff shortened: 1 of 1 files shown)


website/pages/en/versions.js
```diff
@@ -15,6 +15,45 @@ const CWD = process.cwd();
 const siteConfig = require(CWD + '/siteConfig.js');
 const versions = require(CWD + '/versions.json');
 
+class VersionItem extends React.Component {
+  render() {
+    const version = this.props.version;
+    const versionName = version === 'next' ? 'Master' : version;
```
xavier-villelegier: Did you use the "Master" name for simplicity or is it a typo ? 😄

iRoachie: Yea since they’re not sssociated to any version you need to explicitly set a name.

xavier-villelegier: I meant "Master" instead of "Next", my bad*

iRoachie: Lol what should we name it? I just took the pattern from the react-native docs

Monte9: @iRoachie I think it's better to name it `next` for consistency sake.

## Header version name
screen shot 2018-03-07 at 12 13 13 am

##Display version name
screen shot 2018-03-07 at 12 13 18 am

This is kinda confusing. Thoughts?

iRoachie: Changed in https://github.com/react-native-training/react-native-elements/pull/915/commits/f319ad3f2a93d7f2ff9107d0027acfbdaf39e04f

PR #951, opened 2018-02-28
add beta version to getting started

hhg2288:
Add `@beta` version in the instructions when you install react-native-elements, needed to follow the latest API.

Without this, users will install the latest "pre-1.0.0" version & some APIs change.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -114,9 +114,9 @@ If your project is a standard React Native project created using `react-native i
 ### Step 1: Install react-native-elements
 
 ```
-yarn add react-native-elements
+yarn add react-native-elements@beta
 # or with npm
-npm i react-native-elements --save
+npm i react-native-elements@beta --save
 ```
 
 ### Step 2: Install react-native-vector-icons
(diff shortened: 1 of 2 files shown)


iRoachie:
Hey there @hhg2288! Glad to see you managed to get the PR done 👍 There's one more change you have to make. Since `1.0.0-beta2` is already released, the changes you made will only appear for future versions. However we want them see these when they are viewing the docs for `1.0.0-beta2` as well.

You can get a clearer understanding here https://react-native-training.github.io/react-native-elements/docs/contributing.html#rewording-adding-missed-info-or-making-a-typo-for-a-released-version

After you read it, let me know if you understand what you need to do. If not I'll walk you through the process :)

hhg2288:
@iRoachie on it! :)

hhg2288:
let me see if I understand :)

- I need to create a new file copying the content of `getting-started.md` inside `/website/versioned_docs/version-1.0.0.-beta2` right?
- do I need to keep the change in the original `getting-started.md` file ?

also, do I need to PR to `master` or to `docs`?

cc @iRoachie

iRoachie:
Yes that's correct! You don't have to create a new PR, any changes you make to your existing branch (`patch-1`), will be added to this PR.

The only thing that will be different for the file in `versioned_docs/version-1.0.0-beta2` is that the metadata will be slightly different.

Instead it would be:
```md
---
id: version-1.0.0-beta2-getting-started
title: Getting Started
original_id: getting-started
---
```

hhg2288:
done!

hope now its OK :)

iRoachie:
Thanks again @hhg2288

PR #962, opened 2018-03-04
added onLongPress property to FeaturedTile.js

Bobwu1:
added onLongPress property

PR #963, opened 2018-03-04
added onLongPress properties to Tile.js

Bobwu1:
It's useful to have a longpress callback if we want to allow people to treat tile as a button. Usually, longpress pops up a menu or selects/deselect the tile. I think this is a very useful functionality...

### DIFF
--- a/src/tile/Tile.js
+++ b/src/tile/Tile.js
@@ -24,6 +24,7 @@ const Tile = props => {
     caption,
     titleStyle,
     onPress,
+    onLongPress,
     activeOpacity,
     overlayContainerStyle,
     captionStyle,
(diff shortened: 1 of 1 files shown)


iRoachie:
Hey @Bobwu1, any attributes that are passed to `Tile` are passed down to the underlying `Touchable` as you can see https://github.com/react-native-training/react-native-elements/blob/master/src/tile/Tile.js#L97.

So you don't need to specifically define `onLongPress` as a prop. @xavier-villelegier correct me if I'm wrong

iRoachie:
This is the same for https://github.com/react-native-training/react-native-elements/pull/962#issuecomment-370205781

xavier-villelegier:
@iRoachie Yep I was thinking about that, I think the docs should specify that it inherits every props from `TouchableOpacity`

iRoachie:
Alright then. Closing in favor of https://github.com/react-native-training/react-native-elements/issues/964

Issue #918, opened 2018-02-20
Allow title for `Card` to be a React Component


meganlwood:
Is there a way to change the Title at the top of a Card to be a TextInput?

xavier-villelegier:
Hey @meganlwood, not for now ! But I'll let this opened as a possible enhancement

iRoachie:
If anyone wants to work on this, here are the following requirements:
1. Card title should be able to pass in a string or a react element
2. If title is a react element - display it. If not display the current `Text` component with `titleStyle` and `titleNumberOfLines` https://github.com/react-native-training/react-native-elements/blob/master/src/card/Card.js#L60-L70
You can check if a variable is a react element by using `React.isValidElement(<element>)`.
Be sure to checkout the [Checkbox component](https://github.com/react-native-training/react-native-elements/blob/3437699b608ab1fda2bde54d6096bc3c9714abd6/src/checkbox/CheckBox.js#L80-L92) which does this check.

3. Update the propTypes to allow for a component or string https://github.com/react-native-training/react-native-elements/blob/master/src/card/Card.js#L121. 
It'll become `title: PropTypes.oneOfType([PropTypes.string, PropTypes.node])`
4. Update the typescript definitions to allow for a component or string
https://github.com/react-native-training/react-native-elements/blob/master/src/card/Card.js#L60-L70. 
It'll become `title?: string | JSX.Element;`

msevestre:
@iRoachie yep count me in

iRoachie:
Merged in https://github.com/react-native-training/react-native-elements/commit/2308b1b02519fe90a26bd03a024b00a9c8924525. Will be available in 1.0.0-beta3

iRoachie:
Live in [1.0.0-beta3](https://github.com/react-native-training/react-native-elements/releases/tag/v1.0.0-beta3).

PR #968, opened 2018-03-06
Allow title for Card to be a React Component (#918)

vintage:
Ref: https://github.com/react-native-training/react-native-elements/issues/918

Extend the `Card` to accept custom `React` component as title (similar to `CheckBox`).


iRoachie:
Thanks @vintage Super sweet PR

PR #978, opened 2018-03-10
Update index.d.ts

lifeiscontent:
Make optional `selectedIndexes` property optional


iRoachie:
Thanks! @lifeiscontent Good catch.

Issue #942, opened 2018-02-25
Can't set icon type for Avatar


shubhamrathi97:
### Description
`<Avatar
                        large
                        rounded
                        icon={{name:"user-female", type:"simple-line-icon", size: 25}}
                        // title="CR"
                    >`

Above code should render SimpleLineIcon of react-native-vector-icon, but instead, it is only showing icons of font awesome. 

Version
`"react-native-elements": "^1.0.0-beta2"`
`"react-native": "0.52.1"`

### Solution
As I can read in Avatar.js file, We have not handle icon type. It is only rendering font-awesome icons. 
For example : 
[Reference to Avatar.js](https://github.com/react-native-training/react-native-elements/blob/master/src/avatar/Avatar.js#L104)

It is taking the 'type' prop and passing props to Icon of react-native-vector-icon. As per vector-icons doc, We need to import different types of Icons separately. It wouldn't do anything if we pass icon prop 'type' like this.

iksent:
I think it is not "Enhancement", but a real bug. Please, fix it as soon as possible. Very important thing, to use custom icon types in Avatar.

iRoachie:
Sorry @iksent, I misunderstood the issue at first. Will take a look at it this evening

vintage:
@iRoachie If I got it right, the only implementation issue here is that `Avatar` is importing direct version of `Icon` - https://github.com/react-native-training/react-native-elements/blob/master/src/avatar/Avatar.js#L17 - instead of "proxied" one - https://github.com/react-native-training/react-native-elements/blob/99f80cd8b0937e076ce17c7f99daba2ee9b0b997/src/icons/Icon.js which correctly handles the type?

Monte9:
@vintage ah yeah. that's a good catch. 

Yup, all we need to do is to change the `Icon` import from using `react-native-vector-icons` to using our own custom `Icon.js` component. 

Feel free to submit a PR if anyone gets a chance. Also update the `Avatar` demo on the example app to show how to pass multiple icons to the Avatar component.

vintage:
@Monte9 Fixed. I've submitted two PRs:

- First for `react-native-elements` itself to support icon type for `Avatar` - https://github.com/react-native-training/react-native-elements/pull/980
- Second for `react-native-elements-app` to show the usage of the `Avatar` (the section has been empty before my changes, strange)

iRoachie:
[...]

PR #980, opened 2018-03-10
Add support for different icon type in Avatar component #942

vintage:
Ref: https://github.com/react-native-training/react-native-elements/issues/942
Related PR in `react-native-elements-app`: https://github.com/react-native-training/react-native-elements-app/pull/46

### DIFF
--- a/src/avatar/Avatar.js
+++ b/src/avatar/Avatar.js
@@ -12,9 +12,8 @@ import {
   TouchableWithoutFeedback,
 } from 'react-native';
 
+import Icon from '../icons/Icon';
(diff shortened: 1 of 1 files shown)


src/avatar/Avatar.js
```diff
@@ -124,9 +123,7 @@ const Avatar = props => {
       );
     } else if (title) {
       return (
-        <Text style={[styles.title, titleStyle && titleStyle]}>
```
vintage: Strange, I've didn't touch these lines manually. Some crazy git hook? Will look into it.

PR #983, opened 2018-03-11
docs(website): Better syntax colouring for code blocks

iRoachie:
The syntax coloring for code blocks on our website don't look great.

This is because the default syntax highlighter from Docusaurus it uses [highlight.js](https://highlightjs.org) which doesn't support JSX so the code blocks aren't so nice :\. 

This PR uses a commit by @brentvatne to use [prism.js](http://prismjs.com) as the highlighter instead of highlight.js until [custom syntax highlighters](https://github.com/facebook/Docusaurus/issues/438) are supported natively in Docusaurus.

**Before:**
screen shot 2018-03-10 at 9 55 02 pm

**After:**
screen shot 2018-03-10 at 10 39 25 pm

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -9,6 +9,6 @@
     "rename-version": "docusaurus-rename-version"
   },
   "devDependencies": {
-    "docusaurus": "^1.0.7"
+    "docusaurus": "https://github.com/iRoachie/Docusaurus.git#prism"
   }
 }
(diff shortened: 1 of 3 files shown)


Monte9:
It's the details that matter. Looks much nicer! 

Thanks @iRoachie! 💯

PR #989, opened 2018-03-12
docs(1.0.0-beta2): Correct prop on Button to be text

iRoachie:
Soooo what happened is that the original docs from when we were using MKdocs had `title` on the button where it should've been `text`. Same for `textProps` and `textStyle`. 

See https://github.com/react-native-training/react-native-elements/blob/v1.0.0-beta2/docs/API/buttons.md.

This PR corrects the props to be `text` like they should've been.


iRoachie:
Thanks @xavier-villelegier

Issue #892, opened 2018-02-08
Add a `label` feature to the v1 `Input`


mateen-hussain:
### Description
I'm playing with FormLabel and FormINput and this is what I have:
```
import React from 'react';
import PropTypes from 'prop-types';
import { View, Text } from 'react-native';
import { FormLabel, FormInput, FormValidationMessage } from 'react-native-elements';

const DetailsScreen = () => (
  <View>
      <FormLabel>Name</FormLabel>
      <FormInput
        placeholder = 'AB12DEF'
      />
      <FormValidationMessage>Error message</FormValidationMessage>
  </View>
);
export default DetailsScreen;
```
The above code renders to below error
<img width="276" alt="screen shot 2018-02-08 at 17 34 37" src="https://user-images.githubusercontent.com/26531653/35988500-7170854c-0cf6-11e8-9108-3b5888cd4a0d.png">

When I replace FormLabel, FormInput and replace with <Text> then I see the text message. 

My package.json

```
    "expo": "25.0.0",
    "react": "16.2.0",
    "react-native": "0.52.0",
    "react-native-elements": "1.0.0-beta2",
```

kristofferkarlsson93:
I have the same issue

xavier-villelegier:
`FormInput` was replaced by the new refactored `Input` by @Monte9, and `FormLabel` was removed since the `v1.0.0-beta1`, as it is basically just a text with a styling. We're trying to focus on a modular `Input` than multiple form components, but we're open to suggestion about that 👍

iRoachie:
This is actually a great idea. This is how I make my custom inputs for react on the web. Would definitely appreciate I you could make a PR for this.

api182:
I'm getting this error when trying to use 'Input' also though..?

xavier-villelegier:
@api182 If you use `v1.0.0-beta2` this shouldn't be an issue
```jsx
import { Input } from 'react-native-elements'

…
render() {
  return <Input placeholder="Test…" />
}
```

Check your `package.json` to know which version you're using

jacklinton:
Input is not on the list of exports.
`// UI references
// https://ionicframework.com/docs/components/#buttons
// https://material.io/guidelines/components/buttons.html#buttons-raised-buttons
// https://material.angularjs.org/latest/demo/button

// Core
import Button from './buttons/Button';
import Input from './input/Input';

import Icon from './icons/Icon';
import List from './list/List';
import ListItem from './list/ListItem';
[...]

PR #990, opened 2018-03-12
Add `label` for the Input component

xavier-villelegier:
Refs #892 
Associated PR for the example app: https://github.com/react-native-training/react-native-elements-app/pull/48

## Props changes
* `label` adds a label on top of the Input
* `labelStyle` adds style to the label
* Remove useless `displayError` props. If the `errorMessage` is not null, I suppose there is an error to display…

I basically took back the style of the old `FormLabel`

## Result
simulator screen shot - iphone x - 2018-03-12 at 23 26 28

### DIFF
--- a/src/input/Input.js
+++ b/src/input/Input.js
@@ -74,10 +74,11 @@ class Input extends Component {
     return (
       <View>
         {
-          label != null &&
-          <Text style={[styles.label, labelStyle]}>
-            {label}
-          </Text>
+          label && (
+            <Text style={[styles.label, labelStyle]}>
+              {label}
+            </Text>
+          )
         }
         <Animated.View
           style={[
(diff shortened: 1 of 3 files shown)


src/input/Input.js
```diff
@@ -67,6 +71,12 @@ class Input extends Component {
 
     return (
       <View>
+        {
+          label != null &&
+          <Text style={[styles.label, labelStyle]}>
+            {label}
+          </Text>
```
iRoachie: Should be the same as `label && element` not so? Any reason why explicitly checking for `null`?

src/input/Input.js
```diff
@@ -101,9 +111,9 @@ class Input extends Component {
             </View>
           )}
         </Animated.View>
-        {displayError && (
+        {errorMessage != null && (
```
iRoachie: Same as above

src/input/Input.js
xavier-villelegier: Yea I was thinking about the case where label is an empty string, but I can't think about any use case for that

PR #971, opened 2018-03-07
Text style default width

Gregjarvez:
ref: #959

Issue #959, opened 2018-03-01
Button text not centered with "%" width


alfredkam:
###Description
Using `react-native-elements@^1.0.0-beta2`, passing a width in percentage into the Button Component caused the text align to not center.

![screen shot 2018-03-01 at 2 14 26 pm](https://user-images.githubusercontent.com/1495567/36873268-d353cc94-1d5c-11e8-876e-75cb93f42298.png)

Sample Code
```
import { StyleSheet, View } from 'react-native';
import { Button } from 'react-native-elements';

const styles = StyleSheet.create({
  buttonTest: {
    backgroundColor: "#402293",
    width: "80%",
    height: 50,
  }
});
const ButtonTest = () => (
  <View>
    <Button
       text="LOGIN"
       buttonStyle={styles.buttonTest} />
  </View>
);
```
###Solution
include
`textStyle={{width: "100%"}}`

xavier-villelegier:
Thanks for reporting this. [Here is](https://snack.expo.io/rkYLucLuM) a snack example of the bug.
Maybe the best approach is to add `width: '100%'` by default if a `width` is defined in `buttonStyle`

martinezguillaume:
That's weird, I tested the snack in an android phone and it works, it's just an iOS bug 🤨

Gregjarvez:
@xavier-villelegier 
I found this fix to work for me too and it only happens on IOS hence there might be a need to make this platform specific. Leaving this  solution in doesn't affect text position on android.

```javascript
    let { textStyle = {} } = this.props;
    textStyle = (buttonStyle && buttonStyle.width) && !textStyle.width ?
             Object.assign({}, textStyle, {width: '100%'}) : textStyle;
```
what do you think ?

xavier-villelegier:
@Gregjarvez Yep that's what I was proposing ! Feel free to open a PR, you seem to be on fire recently 🔥

Gregjarvez:
@xavier-villelegier sure !!!!!!!!

iRoachie:
Released in [`1.0.0-beta4`](https://github.com/react-native-training/react-native-elements/releases/tag/v1.0.0-beta4)!

miller365:
I am having this same issue except its now with Android only and not affecting iOS.

iRoachie:
Hey @miller365 would prefer if you created a new issue

PR #995, opened 2018-03-13
Text default width

Gregjarvez:
I checked out a new branched and cherry picked only the needed commit. seems cleaner and has no conflicts too . #971 
Refs #959


Gregjarvez:
@xavier-villelegier thanks for the Patience !

PR #1001, opened 2018-03-15
Update typings to match button title prop

viggyfresh:
* One liner - seems the `title` prop became `text` and was then reverted?


chitswe:
We also need to change `textStyle` to `titleStyle`. I think.

chitswe:
@xavier-villelegier Already done! #1007

PR #1016, opened 2018-03-18
Update input.md

peakyDicers:
'icon' is not a prop.


xavier-villelegier:
Hey @peakyDicers, thanks for the PR ! This change is already included [here](https://github.com/react-native-training/react-native-elements/pull/990/files#diff-861c79f8024bbde6ed781eaea1f172c6L18) and should be deployed in like a week or so

PR #1030, opened 2018-03-22
Update button.md

lsiden:
The documentation for embedding an icon in a button is wrong. The icon property takes an object, not an element.

### DIFF
--- a/docs/button.md
+++ b/docs/button.md
@@ -9,7 +9,6 @@ Buttons can be used to interact with the screen. It takes the following props.
 
 ```js
 import { Button } from 'react-native-elements';
-import Icon from 'react-native-vector-icons/FontAwesome';
 
 <Button
   title='BUTTON'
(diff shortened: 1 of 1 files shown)


iRoachie:
It's probably better to create an issue to discuss usually before making a PR.

The `Button` component in `1.0.0-beta*` takes in an Icon as a component as you can see here
https://react-native-training.github.io/react-native-elements/docs/button.html.

The `Button` component in `0.19.0` takes in an Icon as an object as seen here
https://react-native-training.github.io/react-native-elements/docs/0.19.0/button.html.

The version of docs is indicated in the top left of the header. You can click on the version number to change which docs you'd like to view.
screen shot 2018-03-21 at 9 57 12 pm

Here's the versions page https://react-native-training.github.io/react-native-elements/versions.html

lsiden:
Okay, thanks.

Issue #1022, opened 2018-03-19
PR App Previews


iRoachie:
Generate a preview version of the changes in PR to verify if it breaks the RNE app

xavier-villelegier:
(Cf [`react-navigation` PRs](https://github.com/react-navigation/react-navigation/pulls), WIP on [these PRs](https://github.com/xavier-villelegier/react-native-elements/pulls))

xavier-villelegier:
Merged and live on `next`. Still need to enhance this by editing the comment instead of commenting. (I'll work on that this week-end, should be ready for `v1.0.0-beta4` release)


TD;DR
@Monte9 @iRoachie 
We also need to find a solution for external PRs. The fact that the example app is in another repo and people cannot update it by adding example of what they've done in the PR makes the system a bit less effective. If there are breaking changes, then the app will be broken, people should be able to update it and add some examples (and we should see them on the deployed app).

I think we'll need to find another way than the submodule to have live-development + example app in the RNE repo. In this case, people will be able to add examples in their PRs, and we'll see them in the automatically deployed app 🎉

PR #1033, opened 2018-03-24
Deploy automatically an example app on PRs

xavier-villelegier:
Refs #1022


iRoachie:
Example app for 98de54a2277a9d1150e1b154055d4bf785531b04:

QR Code

https://exp.host/@xavier-villelegier/react-native-elements-app?release-channel=98de54a2277a9d1150e1b154055d4bf785531b04

xavier-villelegier:
>  xavier-villelegier deleted a comment from iRoachie 11 minutes ago

cc @iRoachie There was still your token Travis, my bad 😆

xavier-villelegier:
@Monte9 @iRoachie Okay nice that works fine 👆
(At least we can see that the buttons are broken on the `next` branch lol. We need to review #1000 )

.travis.yml
```diff
@@ -17,11 +17,10 @@ install:
 - yarn
 script:
 - yarn run test && codecov
+after_script:
+- ./.travis/expo-ci.sh
```
iRoachie: might wanna change here too :P

xavier-villelegier: Oh, this one is rude

iRoachie:
@xavier-villelegier is there anyway for the bot to update the comment instead of creating a new comment each time? This could get pretty noisy on long PRs

xavier-villelegier:
@iRoachie Yea had added that to my backlog, i've been thinking about that as an enhancement. This implies reading the comments first to check if there is already one, which becomes a bit tricky in shell script. I'll take some time later to improve that with a node script

iRoachie:
Okay great. The logic around it could be checking the author username in the comments

xavier-villelegier:
@iRoachie Yes sure, in js this can be done pretty easily (cf [Github PR comments API](https://developer.github.com/v3/pulls/comments/#response-1))

iRoachie:
Would it be easier to write a node script then?

xavier-villelegier:
@iRoachie Yeaa, that's basic API calls, really not a big deal. I'll do it next week end

PR #1004, opened 2018-03-15
Vertical orientation button group

k3ithl1m:
#869 
Implemented the feature of vertical. When vertical is inserted, it will look like this. 
screen shot 2018-03-15 at 3 12 56 pm

Issue #869, opened 2018-01-30
Vertical orientation for ButtonGroup


Donhv:
As tittle. Pls help me.

xavier-villelegier:
Just putting `containerStyle={{ flexDirection: 'column' }}` will give you a weird result, because the `ButtonGroup` doesn't support vertical orientation for now.

It could be a nice feature to have tho, I'll let that open if someone wants to take the lead on this.

k3ithl1m:
I'm interested in taking a lead on this.

iRoachie:
Go for it!

k3ithl1m:
Would it look like this? Would it be like.
1. Adding a bool prop to determine whether it is vertical or not.
2. If it is, then flexDirection: 'column'.
3. Are there other more specific details that I should consider?
<img width="118" alt="screen shot 2018-02-16 at 5 04 35 am" src="https://user-images.githubusercontent.com/14284173/36302564-ecbc3dfa-12d6-11e8-9f31-127e50037476.png">

iRoachie:
Will need @Donhv to confirm, but it seems like that's it

iRoachie:
@k3ithl1m Did you get anywhere with implementing this?

k3ithl1m:
I didn't because I thought we were waiting for confirmation. :X

k3ithl1m:
Hey @iRoachie Should I just implement it?

iRoachie:
Closing this until we have more information on the requirements

iRoachie:
Example by @EskelCz https://github.com/react-native-training/react-native-elements/pull/1060#issuecomment-427677997

antgel:
@iRoachie Any ideas what the status of this is? Various issues and PRs seem to have come and gone.

iRoachie:
There's no one actively working on it. Feel free to submit a PR if you have an idea of how it should behave and look

antgel:
@iRoachie Realistically, unfortunately I doubt I have the time to work on it. I see it's in the 1.1.0 milestone, which is cool. I see there is a PR https://github.com/react-native-training/react-native-elements/pull/1037, @k3ithl1m any news on this?

k3ithl1m:
Hey @antgel, feel free to work on it as I've been pretty busy as well

antgel:
I'm not sure I have the time and knowhow at the moment. Anyone willing to take it on a professional (funded) basis? Ping me offline if so. :)

[...]

PR #1037, opened 2018-03-26
Add vertical layout for ButtonGroup

k3ithl1m:
#869 #1004 @xavier-villelegier I think this is it


src/buttons/ButtonGroup.js
```diff
@@ -42,7 +43,11 @@ const ButtonGroup = props => {
   return (
     <View
       {...attributes}
-      style={[styles.container, containerStyle && containerStyle]}
+      style={[
+        !vertical && styles.container,
+        containerStyle && containerStyle,
+        vertical && styles.verticalContainer,
```
xavier-villelegier: @k3ithl1m Always put `containerStyle` on the last line, the user should be able to override any default style.

Moreover, `styles.container` and `styles.verticalContainer` are almost the same.
In `styles.verticalContainer`, you should just override what changes in the vertical case (as far as I can see in the `styles` object in added, the `flexDirection` and the `height`)

k3ithl1m:
modified

iRoachie:
Hey @k3ithl1m we just released `v1.0.0-beta4` so there are a lot of changes. Would love if you could do over this PR.

PR #1039, opened 2018-03-26
Added Possibility of Custom Component in Button.

FlaviooLima:
Added Prop to Button named "customComponent" that replace the typical Title/Icon.
This way you can configure the content of a button in any way that you want.
This prop receives a view component.
I don't know if anyone as the same problem as I do but I need to be able to customize the way my designer want the buttons and in the current button I was limited to title and icon.


iRoachie:
What do you guys think about this? Seems like a pretty good use case @martinezguillaume @xavier-villelegier @Monte9 .

Only thing is that I would probably use the standard react `children` as the prop name

martinezguillaume:
I think something like is more clearer, working like a normal children component
<Button>
  <You custom view />
<\Button>

Any thoughts ?

iRoachie:
Yea that’s what I meant

FlaviooLima:
Hi, thanks for the feedback, I'm on it  ;)

Monte9:
I don't think this PR serves any real value. If your use case it to pass children to the button component.. you might as well use the `TouchableOpacity` or `Hightlight` component.

FlaviooLima:
@Monte9  Hi, before anything thanks for the feedback.
Ok, you can use TouchableOpacity or Highlight as you said, and in Android, you can use TouchableNativeFeedback. 
But if I want to use this in a correct way I have to choose a component depending on the platform, like you do in this library.
But since I'm already working with this library, and if this possibility of customizing the component doesn't exist in some way, I will be force to create a new component and if I create a new component then I have no more reason to use your library.
Without this PR you will force some user's to go search for other components or create their own and getting out of this library.

But hey don't take this message in a wrong way, this is only my two cents.   :)

FlaviooLima:
@xavier-villelegier  Sorry can't reopen :/

iRoachie:
@FlaviooLima  Seems like this PR changes lines in files that shouldn't have been changed. It's probably because your branch is out of sync with `next`. Could you recreate this PR? That would be the easiest way to solve the problem.

FlaviooLima:
@iRoachie 
Sure.

PR #1041, opened 2018-03-27
Wrong prop 'icon' defined for Input component instead of correct prop 'leftIcon'

themakerman:
updated the wrong prop icon passed to Input component to correct one ---> leftIcon

### DIFF
--- a/docs/input.md
+++ b/docs/input.md
@@ -15,7 +15,7 @@ import { Input } from 'react-native-elements';
 
 <Input
   placeholder='INPUT WITH ICON'
-  icon={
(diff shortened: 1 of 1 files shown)


themakerman:
@iRoachie Sorry budd. Actually out for business whole week. I don't know if would be able to access my laptop.

iRoachie:
No prob man!

xavier-villelegier:
Fixed in https://github.com/react-native-training/react-native-elements/commit/ae31613347cb2ad1a9169f553f818e8748965310

PR #1043, opened 2018-03-27
Fix custom icon fonts example link

anandprabhu:
Fixes the broken URL link in the documentation for adding custom icon fonts [here](https://react-native-training.github.io/react-native-elements/docs/icon.html#custom-icon-fonts)


iRoachie:
Thanks @anandprabhu! Good catch.

PR #1080, opened 2018-04-06
remove checkmarkIcon style

vishrut:
This styling is preventing the checkmark from rendering correctly


xavier-villelegier:
@vishrut You're right there is a bug on Android because of this style. But removing it will cause `ListItem` to have big margin and looks bigger than others:

capture d ecran 2018-04-06 a 13 34 28

I'll open an issue about that

PR #1082, opened 2018-04-06
Added Possibility of Custom Component in Button.

FlaviooLima:
Added possibility of Custom Component in Button that replace the typical Title/Icon.
This way you can configure the content of a button in any way that you want.

Usage:
Just Add Children to the Button  Like the example:
```
<Button>
	<View style={{ width: 200, height: 40, backgroundColor: "rgba(92, 99,216, 1)", justifyContent: 'center' }} >
		<Text style={{ color: 'white', textAlign: 'center', fontWeight: '800' }}>CUSTOM COMPONENT</Text>
	</View>
</Button>
```

Reason:
I don't know if anyone as the same problem as I do but I need to be able to customize the way my designer want the buttons and in the current button I was limited to title and icon.

@iRoachie  @martinezguillaume @xavier-villelegier @Monte9

### DIFF
--- a/docs/button.md
+++ b/docs/button.md
@@ -53,6 +53,13 @@ import Icon from 'react-native-vector-icons/FontAwesome';
   }}
   containerStyle={{ marginTop: 20 }}
 />
+
+
+<Button>
+	<View style={{ width: 200, height: 40, backgroundColor: "rgba(92, 99,216, 1)", justifyContent: 'center' }} >
+		<Text style={{ color: 'white', textAlign: 'center', fontWeight: '800' }}>CUSTOM COMPONENT</Text>
+	</View>
+</Button>
 ```
 
 ---
--- a/docs/listitem.md
+++ b/docs/listitem.md
@@ -281,6 +281,8 @@ import LinearGradient from 'react-native-linear-gradient' // Only if no expo
 
 ### `component`
 
+replace element with custom element (optional)
+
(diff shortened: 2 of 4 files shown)


FlaviooLima:
@iRoachie  Can you check it now?
For some reason, my Editor wasn't rendering White Spaces. :(

Monte9:
@FlaviooLima I understand your use case of not being able to customize the RNE button per your style.. but I don't think this is the right way to do it. 

With this PR, since you are only using the `<Button />` component as a wrapper all you get is a Touch target based on the platform from RNE.

Logic: `Platform.OS === 'ios' ? TouchableOpacity : TouchableNativeFeedback,`

But by doing this.. you lose out on so many of the other utilities that the RNE button provides you such as:
- display loading indicator with `loading` prop
- display icon
- add LinearGradient background
- platform based Touchable component 

Hence, in your case, if all you need is a touch target.. I would recommend just using `TouchableOpacity` or `TouchableNativeFeedback` directly instead of using the RNE button component. Hope this makes sense.

FlaviooLima:
@Monte9  Ok.
If in a near future if you need this to work, "give me a call".  :) 

But be aware that:
- display loading indicator with loading prop (which I think I can solve it quickly )
- display icon (but this is intended)
- add LinearGradient background (which I think I can solve it quickly )
- platform based Touchable component (you don't lose this)

PR #1086, opened 2018-04-07
Update avatar.md to display how to change background color

lucaszanella:
I had to go deep and  find that `overlayContainerStyle` meant the background thing. Maybe someone can help and update the image or maybe I'll do in the future

### DIFF
--- a/docs/avatar.md
+++ b/docs/avatar.md
@@ -57,6 +57,7 @@ title: Avatar
 <Avatar
   large
   title="LW"
+  overlayContainerStyle={{backgroundColor: 'orange'}}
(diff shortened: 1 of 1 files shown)


xavier-villelegier:
Hey @lucaszanella, sorry you wasted some time on that ! 😕
There is already an example in the docs though, [check here](https://react-native-training.github.io/react-native-elements/docs/avatar.html#avatar-with-icons)

lucaszanella:
@xavier-villelegier  I saw the example but in my head it was only for the image. Maybe something with language since I don't speak native english

PR #1092, opened 2018-04-10
Add WeChat color to SocialIcon

limianwang:
Hi, I would like to add WeChat theme to the SocialIcon list as well.  

Color HEX was found via: https://brandcolors.net/b/wechat

Thanks!


Monte9:
For sure. Let us know if you do end up putting it on the App Store. We are planning to start featuring apps that use RNE.

PR #1098, opened 2018-04-11
Refactor Avatar Size props

### LINKED ISSUE #1095: Rewrite Avatar sizes props to use one size prop
Right now avatar has `small`, `medium`, `large`,  and `xlarge` props for setting the size.

This doesn't make sense because the user shouldn't be able to set both `small` and `medium` at the same time.

Instead we should have one `size` prop that can take in either "small" | "medium" | "large" | "xlarge"

Gregjarvez:
fixes #1095

### DIFF
--- a/docs/avatar.md
+++ b/docs/avatar.md
@@ -9,26 +9,26 @@ title: Avatar
 
 ```js
 <Avatar
-  small
+  size="small"
   rounded
   source={{uri: "https://s3.amazonaws.com/uifaces/faces/twitter/ladylexy/128.jpg"}}
   onPress={() => console.log("Works!")}
   activeOpacity={0.7}
 />
 <Avatar
-  medium
+  size="medium"
   source={{uri: "https://s3.amazonaws.com/uifaces/faces/twitter/kfriedson/128.jpg"}}
   onPress={() => console.log("Works!")}
   activeOpacity={0.7}
 />
 <Avatar
-  large
+  size="large"
   source={{uri: "https://s3.amazonaws.com/uifaces/faces/twitter/brynn/128.jpg"}}
   onPress={() => console.log("Works!")}
   activeOpacity={0.7}
 />
 <Avatar
-  xlarge
+  size="xlarge"
   rounded
   source={{uri: "https://s3.amazonaws.com/uifaces/faces/twitter/adhamdannaway/128.jpg"}}
   onPress={() => console.log("Works!")}
@@ -42,26 +42,26 @@ title: Avatar
 
 ```js
 <Avatar
-  small
+  size="small"
   rounded
   title="MT"
   onPress={() => console.log("Works!")}
   activeOpacity={0.7}
 />
 <Avatar
-  medium
+  size="medium"
   title="BP"
   onPress={() => console.log("Works!")}
   activeOpacity={0.7}
 />
 <Avatar
-  large
+  size="large"
   title="LW"
   onPress={() => console.log("Works!")}
   activeOpacity={0.7}
 />
 <Avatar
-  xlarge
+  size="xlarge"
   rounded
   title="CR"
   onPress={() => console.log("Works!")}
@@ -75,7 +75,7 @@ title: Avatar
 
 ```js
 <Avatar
-  small
+  size="small"
   rounded
   icon={{name: 'user'}}
   onPress={() => console.log("Works!")}
@@ -83,7 +83,7 @@ title: Avatar
   containerStyle={{flex: 2, marginLeft: 20, marginTop: 115}}
 />
 <Avatar
-  medium
+  size="medium"
   overlayContainerStyle={{backgroundColor: 'blue'}}
   icon={{name: 'meetup', color: 'red'}}
   onPress={() => console.log("Works!")}
@@ -91,7 +91,7 @@ title: Avatar
   containerStyle={{flex: 3, marginTop: 100}}
 />
 <Avatar
-  large
+  size="large"
   icon={{name: 'rocket', color: 'orange'}}
   overlayContainerStyle={{backgroundColor: 'white'}}
   onPress={() => console.log("Works!")}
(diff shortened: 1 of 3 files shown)


src/avatar/Avatar.js
```diff
@@ -264,6 +236,19 @@ Avatar.propTypes = {
   }),
 };
 
-Avatar.defaultProps = defaultProps;
+Avatar.defaultProps = {
+  showEditButton: false,
+  onEditPress: null,
+  width: 34,
+  height: 34,
```
xavier-villelegier: Maybe use your `DEFAULT_SIZES` object here, to have only one source for the sizes

Gregjarvez: I will do that @xavier-villelegier 😀typescript 😍

src/avatar/Avatar.js
```diff
@@ -45,25 +49,10 @@ const Avatar = props => {
...
-    width = 34;
-    height = 34;
-  } else if (!width) {
-    width = height;
-  } else if (!height) {
-    height = width;
+  const iconDimension = DEFAULT_SIZES[size];
+
+  if (iconDimension) {
+    [width, height] = iconDimension;
```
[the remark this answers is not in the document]
iRoachie: I was just about to type this on the way to work :D Good job. Think we can set `small` as true by default.

Is there any reason we explicitly have `height` and `width` as props. If the avatar is either a circle or rectangle that means that width and height will always be the same no? If so `size` could either be `'small', 'medium', 'large', 'xlarge'` or a number.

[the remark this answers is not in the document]
xavier-villelegier: @Gregjarvez 
I think the idea of @iRoachie to have either a string or a number is cool.
You can just have something like `width = typeof props.size === 'number' ? props.size : DEFAULT_SIZES[size][0] || DEFAULT_SIZES.small[0]` (doesn't have to be exactly this but you got the idea), same goes for the `height`. Be sure you also set `small` as the `defaultProps` for `size`.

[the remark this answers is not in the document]
Gregjarvez: Fantastic. so which props do we get rid of apart from `width and height`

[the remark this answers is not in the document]
xavier-villelegier: Oh was just talking about `'small', 'medium', 'large', 'xlarge', 'width', 'height'`. 6 become 1 props 🔥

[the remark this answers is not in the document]
Gregjarvez: oh ok. I will get that done soon !!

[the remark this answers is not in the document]
Gregjarvez: @iRoachie @xavier-villelegier I did the changes required seem to work 
https://snack.expo.io/rJ_bq9isz
changed code 
```javascript 
// iRoachie mentioned avatar is either a circle or a square. one number will work fine
const DEFAULT_SIZES = {
  small: 34,
  medium: 50,
  large: 75,
  xlarge: 150,
};
//size is `small` by default in defaultProps
 let { size } = props;

const iconDimension = typeof size === 'number' ? size : DEFAULT_SIZES[size];

let height;
let width = (height = iconDimension);

```

docs/avatar.md
```diff
@@ -273,23 +250,22 @@ Makes the avatar circular
...
 
-### `source`
-
-Image source
+Size of the avatar
 
 |      Type      | Default |
 | :------------: | :-----: |
-| object (style) |  none   |
+| string(`small`, `medium`, `large`, `xlarge`) |  `small`  |
```
iRoachie: type would be string(`small`, `medium`, `large`, `xlarge`) OR number

src/index.d.ts
```diff
@@ -185,24 +185,10 @@ export interface AvatarProps {
...
-  /**
-   * Large sized icon
-   */
-  large?: boolean;
-
-  /**
-   * Extra-large sized icon
-   */
-  xlarge?: boolean;
+  size?: 'small' | 'medium' | 'large' | 'xlarge '
```
iRoachie: Also needs to update here

docs/avatar.md
Gregjarvez: yeah I changed that for the next pul request

src/avatar/Avatar.js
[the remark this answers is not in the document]
xavier-villelegier: @Gregjarvez Just add a fallback on the small size for `iconDimension` (`DEFAULT_SIZES[size] || DEFAULT_SIZES.small`) in case you typed an invalid string size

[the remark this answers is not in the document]
xavier-villelegier: Otherwise it looks good to me

src/avatar/Avatar.js
```diff
@@ -42,29 +44,12 @@ const Avatar = props => {
     imageProps,
     ...attributes
   } = props;
+  let { size } = props;
 
-  let { width, height } = props;
+  const iconDimension = typeof size === 'number' ? size : DEFAULT_SIZES[size];
```
iRoachie: You forgot the change from @xavier-villelegier. This will still fail if I pass in `size="bob""`

Gregjarvez: oh yeh sorry about that. I was thinking propTypes will fail it 
`PropTypes.Oneof(['small', '....'])

iRoachie: Yea but propTypes is just a warning, it doesn't stop the app from trying to load. If we don't provide a fallback it'll red screen (crash)

Gregjarvez: that is done

src/index.d.ts
```diff
@@ -185,24 +171,10 @@ export interface AvatarProps {
   iconStyle?: StyleProp<TextStyle>;
 
   /**
-   * Small sized icon
+   * Size of Avatar
```
iRoachie: Just one small change!

Add a new line here and put `@default "small"`

Gregjarvez: sure

PR #1101, opened 2018-04-12
Use v0.19.0 as default website version

xavier-villelegier:
We'll need to make sure `0.19.0` is the first index everytime we generate a new versioned doc, but this is the only way waiting for this to be fixed https://github.com/facebook/Docusaurus/issues/346


iRoachie:
Ha once we don't forget this should be fine. I'm just wondering how hard it is to implement this in docusaurus.

PR #1103, opened 2018-04-12
Revert website default version

xavier-villelegier:


Monte9:
Why did we revert these changes? Did it not produce the expected behaviour with indexing the default version for the docs?

Also I assume we are still waiting to test this PR https://github.com/facebook/Docusaurus/pull/554 out

xavier-villelegier:
@Monte9 This isn't that simple actually. Each generated doc depends on the previous ones, so you can't rearrange the versions like that, this was crashing…

Monte9:
ah gotcha. Ok, np. We'll just have to wait it out.

PR #1104, opened 2018-04-12
Fix Icon style when using touchableNativeFeedback

### LINKED ISSUE #1084: Icon styles break when using TouchableNativeFeedback
### Description

1. Use Icon with `TouchableNativeFeedback` using the `component` prop
2. Icon to render and styles show correctly
3. No styles are applied to the Icon

Also please include the version of `react-native-elements` that you are using.
> 1.0.0-beta4

```jsx
<Icon
  component={TouchableNativeFeedback}
  name="check"
  color="#fff"
  size={24}
  raised
  iconStyle={{ fontSize: 20 }}
  containerStyle={{ backgroundColor: Theme.accent, marginRight: 0 }}
  onPress={this.addReview}
/>
```

### Solution
The reason this breaks is that we assume that `component` is always a component that is `View` based. https://github.com/react-native-training/react-native-elements/blob/master/src/icons/Icon.js#L32
`TouchableOpacity` and `TouchableHighlight` both have views attached to them natively. 

However, this is not the case with `TouchableNativeFeedback` - which requires an encasing `View`.
So when we try to apply `containerStyle` it will fail. https://github.com/react-native-training/react-native-elements/blob/master/src/icons/Icon.js#L60.

The solution is to place a `View` inside the Touchable and apply the container styles to it.

Gregjarvez:
fixes #1084

### DIFF
--- a/src/icons/Icon.js
+++ b/src/icons/Icon.js
@@ -17,57 +17,46 @@ const Icon = props => {
     size,
     color,
     iconStyle,
-    component,
     underlayColor,
     reverse,
     raised,
     containerStyle,
     reverseColor,
     onPress,
+    component: Component = onPress ? TouchableHighlight : View,
     ...attributes
   } = props;
 
-  let Component = View;
-  if (onPress) {
-    Component = TouchableHighlight;
-  }
-  if (component) {
-    Component = component;
-  }
-  let Icon;
-  if (!type) {
-    Icon = getIconType('material');
-  } else {
-    Icon = getIconType(type);
-  }
+  let Icon = getIconType(type || 'material');
   return (
-    <Component
-      {...attributes}
-      underlayColor={reverse ? color : underlayColor || color}
-      style={[
-        (reverse || raised) && styles.button,
-        (reverse || raised) && {
-          borderRadius: size + 4,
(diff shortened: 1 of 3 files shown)


Gregjarvez:
I took the liberty to cleanup the conditions alittle bit 😀

src/icons/Icon.js
```diff
@@ -17,57 +17,46 @@ const Icon = props => {
...
-      onPress={onPress}
-    >
-      <Icon
-        style={[{ backgroundColor: 'transparent' }, iconStyle && iconStyle]}
-        size={size}
-        name={name}
-        color={reverse ? reverseColor : color}
-      />
-    </Component>
+    <View style={containerStyle || {}}>
```
xavier-villelegier: Is the `|| {}` necessary ?

Gregjarvez: most probably not but it used to `containerStyle && containerStyle` so i thought i'd provide a fallback too

iRoachie: Don't really like it. Prefer either use the `&&` like the others or add the `{}` in the default props

Gregjarvez: sure !!

Issue #1117, opened 2018-04-15
what does property "chevron" mean in ListItem


xiechao06:
### Description

1. Explain what you did
in 
https://react-native-training.github.io/react-native-elements/docs/listitem.html#chevron,
I find nothing useful. I guess it controls whether the right arrow appears, but it doesn't.

version: 1.0.0-beta4

xavier-villelegier:
@xiechao06 Yes it controls the right arrow.

|Code|Result|
|-----|-------|
|`<ListItem title="Hey" chevron />`|![image](https://user-images.githubusercontent.com/17592779/38778127-74a303ec-40b4-11e8-8d73-da35bf68a206.png)|
|`<ListItem title="Hey" chevron />`|![image](https://user-images.githubusercontent.com/17592779/38778129-7e3f3196-40b4-11e8-98bb-c3dfee8c8221.png)|

3ace:
I'm also experiencing this, I'm using version 0.19. By default Chevron is displayed, even setting chevron={false} wont hide it.

3ace:
@iRoachie ah thanks, I've been looking at 1.0.0 beta 4 documentation all this time without realizing that it's different :(

iRoachie:
Sorry about that mate! In the next version the website will show 0.19 as the default version

PR #1118, opened 2018-04-15
Add descriptions for ListItem props in v1.0.0-beta4 docs

xavier-villelegier:


iRoachie:
They’re there for the next branch, is this a rush fix for live?

xavier-villelegier:
@iRoachie Yep to avoid future issues like #1117

Issue #1125, opened 2018-04-17
Ability to set color and accessibility label for cancel button in SearchBar


cdalton615:
### Description

1. Explain what you did: I've added the SearchBar component to my navigation bar in my ios app. I've set the component's "cancelButtonTitle" and "platform" props so that the "Cancel" button is displayed.
2. What you expected to happen: I expect to be able to set the color of the "Cancel" button, as well as the accessibility label. 
3. What actually happens: Cancel button's text color is set to the default blue, and accessibility label is not set.

`react-native-elements` version: 1.0.0-beta4

### Solution
Add props to SearchBar component for the cancel button's color and accessibility label. If these values are set, use them when rendering the button.

M-Jas:
#1159

M-Jas:
#1159

xavier-villelegier:
Fixed in `v1.0.0-beta5`, you now have a `cancelButtonProps` props

PR #1126, opened 2018-04-17
ios: set color and accessibility label for cancel button in SearchBar

cdalton615:
Ability to set color and accessibility label for cancel button in SearchBar #1125


xavier-villelegier:
Closing in favor of #1159

PR #1127, opened 2018-04-19
Must be 'showLoadingIcon' not 'showLoading'

jennish:
The prop to show a loading icon on the SearchBar should be 'showLoadingIcon' not 'showLoading'


iRoachie:
The name of the prop in the source is [`showLoading`](https://github.com/react-native-training/react-native-elements/blob/master/src/searchbar/SearchBar-ios.js#L78). Is this PR suggesting a name change to the prop?

jennish:
@iRoachie  Sorry my mistake. I am using version 0.19.1.

Issue #1069, opened 2018-04-04
Avatar Image in ListItem beta4 not working


neogeno:
### Description

Avatar Image not appearing in Listitem after upgrading to beta4. Code used to be  avatar={{uri:l.avatar_url}} but even after changing to new syntax  avatar={{ source: { uri: l.avatar_url } }} still nothing shows up

1. Explain what you did

Upgraded fro beta3 to beta4 and left code unchanged. Found my ListItems no longer show Avatars.

2. What you expected to happen

Round Avatar to show up on the left side of each List Item

3. What actually happens

Text Content is rendered but not Avatar

Also please include the version of `react-native-elements` that you are using.

1.0.0-beta4

### Reproduction Steps and Sample Code

 `
const list = [
  {
    name: 'Amy Farha',
    avatar_url: 'https://s3.amazonaws.com/uifaces/faces/twitter/ladylexy/128.jpg',
    subtitle: 'test'
  },
  {
    name: 'Chris Jackson',
    avatar_url: 'https://s3.amazonaws.com/uifaces/faces/twitter/adhamdannaway/128.jpg',
    subtitle: 'Testing'
  }];

{
                    list.map((l, i) => (
                      <ListItem
                        containerStyle={{height:80, justifyContent:'center', borderBottomColor: '#E5E5E5', borderBottomWidth:0.5}}
                        roundAvatar
                        avatar={{ source: { uri: l.avatar_url } }}
                        key={i}
                        title={l.name}
                        titleStyle={{fontSize:15}}
                        subtitle={
                        <View style={styles.subtitleView}>
                          <Text style={{color: 'rgba(72, 100, 124, 1)'}}>{l.subtitle}</Text>
                          <Text style={styles.ratingText}>5 months ago</Text>
                        </View>
                        }
                        subtitleStyle={{fontSize:12}}
                        rightTitle='+150 SAT'
                        rightTitleNumberOfLines={2}
                        rightTitleStyle={{color: 'rgba(44, 62, 80, 1)'}}
                        rightIcon={{name:'zap', type: 'octicon', color: 'rgba(72, 100, 124, 1)'}}
                      />
                    ))
                  }`

We highly recommend that you try to reproduce your bug on https://snack.expo.io/ and provide a link.

If you can't reproduce the bug on Snack, provide an example of your code and briefly explain what the code is suppose to be doing.

### Solution

[...]

PR #1131, opened 2018-04-22
Update input.md

yakupad:


iRoachie:
Not sure what you're trying to accomplish in the PR. But we have different versions of the docs that you can view.

See [this comment](https://github.com/react-native-training/react-native-elements/issues/1069#issuecomment-382738633)

PR #1146, opened 2018-04-27
Bounce Form elements

iRoachie:
Since we have fully functioning v1 input, I think it's safe to remove these components


Monte9:
Agreed. Nice cleanup! 👍

PR #1152, opened 2018-04-29
Typo?

nonotest:
It seems that a touchableopacity is added - not a touchablehighlight.

shouldn't a touchablehighlight be used here?
* https://medium.com/differential/better-cross-platform-react-native-components-cb8aadeba472
* https://stackoverflow.com/questions/39123357/when-to-use-touchablenativefeedback-touchablehighlight-or-touchableopacity

thanks!


nonotest:
ok

iRoachie:
Thank you again! 💯

PR #1165, opened 2018-05-05
featute(tooltip): first proposal

AndreiCalazans:
### Proposal
Tooltip is something commonly used, I had a couple situations where I needed a tooltip and most of the time I've always encountered heavily styled ones. This is my attempt to a smart and very customisable tooltip. I decided to propose to RNE because I started using RNE recently and I'm very satisfied with it, I feel as it is exactly what we need from an UI toolkit.

Nonetheless I published this tooltip [here](https://github.com/AndreiCalazans/rn-tooltip) so I could start using it, but instead of maintaining it there I'd love to just use it with RNE.

### Usage

screen shot 2018-05-05 at 15 58 20

```javascript
import { Tooltip, Text } from 'react-native-elements';

...

<Tooltip tooltipWidth={200} tooltipText="Tooltip info goes here">
  <Text>Press me</Text>
</Tooltip>
```

### Props

```javascript
Tooltip.propTypes = {
  children: PropTypes.element,
  withPointer: PropTypes.bool,
  tooltipText: PropTypes.string,
  tooltipComponent: PropTypes.element,
  toggleOnPress: PropTypes.bool,
  tooltipHeight: PropTypes.number,
  tooltipWidth: PropTypes.number,
  tooltipContainerStyle: PropTypes.any,
  pointerColor: PropTypes.string,
  tooltipTextStyle: PropTypes.any,
  onClose: PropTypes.func,
  withOverlay: PropTypes.bool,
  backgroundColor: PropTypes.string,
  highlightColor: PropTypes.string,
  tooltipContainerDefaultStyle: PropTypes.any,
}
```

```javascript
static defaultProps = {
    withOverlay: true,
    hightlightColor: 'transparent',
    withPointer: true,
    toggleOnPress: true,
    tooltipHeight: 40,
    tooltipWidth: 150,
    tooltipContainerStyle: {},
    tooltipTextStyle: {},
    backgroundColor: Colors.darkergray,
    onClose: () => {},
  };

```

### Todos
If you guys approve this and decide that We can add the tooltip to RNE then the things below are still missing. I'll add them upon acceptance of this proposal.
- [ ] Add typescript types for Tooltip.
- [ ] Add Tooltip section to docs.

### DIFF
--- a/src/tooltip/Tooltip.js
+++ b/src/tooltip/Tooltip.js
@@ -1,0 +1,234 @@
+import * as React from 'react';
+import { Text, TouchableOpacity, Modal, View } from 'react-native';
+import NativeMethodsMixin from 'react-native/Libraries/Renderer/shims/NativeMethodsMixin';
+import PropTypes from 'prop-types';
+
+import Triangle from './Triangle';
+import { Colors, ScreenWidth, ScreenHeight, isIOS } from './helpers';
+import getTooltipCoordinate from './getTooltipCoordinate';
+
+
+class Tooltip extends React.PureComponent {
+  state = {
+    isVisible: false,
+    yOffset: 0,
+    xOffset: 0,
+    elementWidth: 0,
+    elementHeight: 0,
+  };
+
+  static defaultProps = {
+    withOverlay: true,
+    hightlightColor: 'transparent',
+    withPointer: true,
+    toggleOnPress: true,
+    tooltipHeight: 40,
+    tooltipWidth: 150,
+    tooltipContainerStyle: {},
+    tooltipTextStyle: {},
+    backgroundColor: Colors.darkergray,
+    onClose: () => {},
+  };
+
+  renderedElement;
+
+  toggleTooltip = () => {
+    const { onClose } = this.props;
+    this.setState(prevState => {
+      if (prevState.isVisible && !isIOS) {
+        onClose && onClose();
+      }
+
+      return { isVisible: !prevState.isVisible };
+    });
+  };
+
+  wrapWithPress = (toggleOnPress, children) => {
+    if (toggleOnPress) {
+      return (
+        <TouchableOpacity onPress={this.toggleTooltip} activeOpacity={1}>
+          {children}
+        </TouchableOpacity>
+      );
+    }
+
+    return children;
+  };
+
+  getTooltipStyle = () => {
+    const { yOffset, xOffset, elementHeight, elementWidth } = this.state;
+    const {
+      tooltipHeight,
+      backgroundColor,
+      tooltipWidth,
+      tooltipContainerStyle,
+      withPointer,
+      tooltipContainerDefaultStyle,
+    } = this.props;
+
+    const { x, y } = getTooltipCoordinate(
+      xOffset,
+      yOffset,
+      elementWidth,
+      elementHeight,
+      ScreenWidth,
+      ScreenHeight,
+      tooltipWidth,
+      tooltipHeight,
+      withPointer,
+    );
+    const tooltipDefaultStyle = {
+      display: 'flex',
+      alignItems: 'center',
+      justifyContent: 'center',
+      flex: 1,
+      backgroundColor,
+      borderRadius: 10,
+      padding: 10,
+    };
+
+    const defaultToUse = tooltipContainerDefaultStyle || tooltipDefaultStyle;
+    return {
+      ...defaultToUse,
+      position: 'absolute',
+      left: x,
+      top: y,
+      width: tooltipWidth,
+      height: tooltipHeight,
+      ...tooltipContainerStyle,
+    };
+  };
+
+  renderPointer = () => {
+    const { yOffset, xOffset, elementHeight, elementWidth } = this.state;
+    const { backgroundColor, pointerColor } = this.props;
+    const pastMiddleLine = yOffset > ScreenHeight / 2;
+
+    return (
+      <View
+        style={{
+          position: 'absolute',
+          top: pastMiddleLine ? yOffset - 13 : yOffset + elementHeight - 2,
+          left: xOffset + elementWidth / 2 - 7.5,
+        }}
+      >
+        <Triangle
+          style={{ borderBottomColor: pointerColor || backgroundColor }}
+          isDown={pastMiddleLine}
+        />
+      </View>
+    );
+  };
+  renderContent = (withTooltip) => {
+    const {
+      tooltipComponent,
+      withPointer,
+      tooltipText,
+      toggleOnPress,
+      tooltipTextStyle,
+      hightlightColor,
+    } = this.props;
+
+    if (!withTooltip)
+      return this.wrapWithPress(toggleOnPress, this.props.children);
+
+    const { yOffset, xOffset } = this.state;
+    return (
+      <React.Fragment>
+        <View
+          style={{
+            position: 'absolute',
+            top: yOffset,
+            left: xOffset,
+            backgroundColor: hightlightColor,
+            overflow: 'visible',
+          }}
+        >
+          {this.props.children}
+        </View>
+        {withPointer && this.renderPointer()}
+        <View style={{ ...this.getTooltipStyle() }}>
+          {tooltipComponent ? (
+            tooltipComponent
+          ) : (
+            <Text style={tooltipTextStyle}>{tooltipText}</Text>
+          )}
+        </View>
+      </React.Fragment>
+    );
+  };
+
+  componentDidMount() {
+    // wait to compute onLayout values.
+    setTimeout(this.getElementPosition, 500);
+  }
+
+  getElementPosition = () => {
+    if (this.renderedElement) {
+      NativeMethodsMixin.measureInWindow.call(
+        this.renderedElement,
+        (x, y, width, height) => {
+          this.setState({
+            xOffset: x,
+            yOffset: y,
+            elementWidth: width,
+            elementHeight: height,
+          });
+        },
+      );
+    }
+  };
+
+  render() {
+    const { isVisible } = this.state;
+    const { onClose, withOverlay } = this.props;
+
+    return (
+      <View collapsable={false} ref={e => (this.renderedElement = e)}>
+        {this.renderContent(false)}
+        <Modal
+          animationType="fade"
+          visible={isVisible}
+          transparent
+          onDismiss={onClose}
+          onRequestClose={onClose}
+        >
+          <TouchableOpacity
+            style={styles.container(withOverlay)}
+            onPress={this.toggleTooltip}
+            activeOpacity={1}
+          >
+            {this.renderContent(true)}
+          </TouchableOpacity>
+        </Modal>
+      </View>
+    );
+  }
+}
+
+Tooltip.propTypes = {
+  children: PropTypes.element,
+  withPointer: PropTypes.bool,
+  tooltipText: PropTypes.string,
+  tooltipComponent: PropTypes.element,
+  toggleOnPress: PropTypes.bool,
+  tooltipHeight: PropTypes.number,
+  tooltipWidth: PropTypes.number,
+  tooltipContainerStyle: PropTypes.any,
+  pointerColor: PropTypes.string,
+  tooltipTextStyle: PropTypes.any,
+  onClose: PropTypes.func,
+  withOverlay: PropTypes.bool,
+  backgroundColor: PropTypes.string,
+  highlightColor: PropTypes.string,
+  tooltipContainerDefaultStyle: PropTypes.any,
+}
+
+const styles = {
+  container: (withOverlay) => ({
+    backgroundColor: withOverlay ? Colors.overlay_bright : 'transparent',
+    flex: 1,
+  }),
+};
+
+export default Tooltip;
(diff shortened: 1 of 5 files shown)


Monte9:
I love this. Looks really useful and the API looks simple to use. 💯 

> Nonetheless I published this tooltip here so I could start using it, but instead of maintaining it there I'd love to just use it with RNE.

That's a good idea. As part of RNE, 
- It be available to more users (RNE has about 10,000 weekly downloads), 
- It will be easier to maintain as there were be more contributors willing to help improve it.

Wonder what @iRoachie thinks about this.

iRoachie:
This looks like a really awesome PR and I think it could be a great addition to RNE. Do you mind adding some docs or posting here in a comment (will still to add docs eventually) what each of the props do? So we could make reviewing this PR a bit easier

xavier-villelegier:
Thanks a lot for you PR ! It is really neat, code is clean, docs and types too… Awesome @AndreiCalazans. This component can be a big + for a library like RNE 💯 🔥

Just 2-3 things:
- The `tooltipComponent` props is basically what is rendered inside the tooltip, and if none is provided you just render the `tooltipText` with the `tooltipTextStyle` style. I'm thinking about React Bootstrap that has just an `overlay` props that takes a component. By doing that you can get rid of `tooltipText` and `tooltipTextStyle`, and people can just do

```jsx
<Tooltip overlay={<Text style={whaveter}>My tooltip text</Text>}>
  <Button>Click me !</Button>
</Tooltip>
```

- I think you can remove all the `tooltipXXX` prefixes (these are already `Tooltip` props) to have `containerStyle`, `onPress`, `width`, `height`…

- If you have a `onClose`, why not having a `onOpen` props ? (Using the `onShow` React Native `Modal` props)

@iRoachie Let me know what you think about that

AndreiCalazans:
Ok @xavier-villelegier, I followed your suggestions. The only thing I did different was use `popover` instead of `overlay` because there is a `withOverlay` prop that serves as a flag for the opacity overlay. I think users would get confused with that.

The props got much simpler now.
``` js
Tooltip.propTypes = {
  children: PropTypes.element,
  withPointer: PropTypes.bool,
  popover: PropTypes.element,
  toggleOnPress: PropTypes.bool,
  height: PropTypes.number,
  width: PropTypes.number,
  containerStyle: PropTypes.any,
  pointerColor: PropTypes.string,
  onClose: PropTypes.func,
  onOpen: PropTypes.func,
  withOverlay: PropTypes.bool,
  backgroundColor: PropTypes.string,
  highlightColor: PropTypes.string,
};
```

```js
import { Tooltip, Text } from 'react-native-elements';

...

<Tooltip popover={<Text>Info here</Text>}>
  <Text>Press me</Text>
</Tooltip>

```

docs/tooltip.md
```diff
@@ -25,15 +25,13 @@ import { Tooltip, Text } from 'react-native-elements';
 * [`backgroundColor`](#backgroundcolor)
```
Gregjarvez: Could you rearrange this in a aphabethical order 💪🏽

docs/tooltip.md
```diff
@@ -0,0 +1,159 @@
+---
+id: tooltip
+title: Tooltip
+---
+
+Easy to use and customisable tooltip.
+
+<img alt="tooltip example gif" width='290' src='/react-native-elements/img/tooltipExample.gif'>
+
+### Usage.
```
iRoachie: Random `.` :p

docs/tooltip.md
```diff
@@ -0,0 +1,159 @@
...
+
+---
+
+### `containerStyle`
+
+Passes style object to tooltip container
+
+|  Type   | Default  |
+| :-----: | :------: |
+| Object | {} |
```
iRoachie: For any types that are an object, let's stick to this convention. See https://github.com/react-native-training/react-native-elements/blame/master/docs/listitem.md#L307 for example

iRoachie: Also the containerStyle in the source technically has other methods on it
```js
position: 'absolute',
left: x,
top: y,
width,
height,
backgroundColor,
// default styles
display: 'flex',
alignItems: 'center',
justifyContent: 'center',
flex: 1,
borderRadius: 10,
padding: 10,
```

So you could use `inherited styling` as the value here.

docs/tooltip.md
```diff
@@ -0,0 +1,159 @@
...
+
+---
+
+### `pointerColor`
+
+Color of tooltip pointer, it defaults to the [`backgroundColor`](#backgroundcolor) if none is passed .
+
+|  Type  |                            Default                             |
+| :----: | :------------------------------------------------------------: |
+| string | [`backgroundColor`](#backgroundcolor) |
```
iRoachie: Love that you used the link to the `backgroundColor` prop here 💯

docs/tooltip.md
```diff
@@ -0,0 +1,159 @@
...
+
+
+### `width`
+
+Tooltip container width. Necessary in order to render the container in the correct place. Pass height according to the size of the content rendered inside the container.
+
+
+|      Type      | Default |
+| :------------: | :-----: |
+| Object |  {}   |
```
iRoachie: Think this is supposed to be number? And also in the source it has a default value of `150`

src/index.d.ts
```diff
@@ -1305,6 +1305,76 @@ export interface SearchBarBase extends TextInputProperties {
...
+  /**
+   * Component to be rendered as the display container.
+   */
+
+  containerStyle?: {}
+  /**
+   * Passes style object to tooltip container
+   */
+
+  height: number;
```
iRoachie: In the source you provided a default value to for `height` therefore it's not a required attribute. Use the `?` character to denote that it's optional

src/index.d.ts
```diff
@@ -1305,6 +1305,76 @@ export interface SearchBarBase extends TextInputProperties {
...
+  /**
+   * 
+   * Tooltip container height. Necessary in order to render the container in the correct place. Pass height according to the size of the content rendered inside the container.
+   */
+
+  width: number;
```
iRoachie: Same as height above ^

src/index.d.ts
```diff
@@ -1305,6 +1305,76 @@ export interface SearchBarBase extends TextInputProperties {
   onChangeText?(text: string): void;
 }
 
+export class Tooltip extends React.Component<TooltipProps, any> {}
```
iRoachie: Just cosmetic, but place the export under the Props.

src/index.d.ts
```diff
@@ -1305,6 +1305,76 @@ export interface SearchBarBase extends TextInputProperties {
...
+   * Tooltip container width. Necessary in order to render the container in the correct place. Pass height according to the size of the content rendered inside the container.
+   */
+
+
+  withOverlay?: boolean;
+  /**
+   *  Flag to determine whether or not dislay overlay shadow when tooltip is open.
+   */
+
+  withPointer?: boolean;
```
iRoachie: Cosmetic as well - the pattern for descriptions is:

```
comment
prop

comment
prop
```

So the new line is between the two props. The description of the prop goes just above it.

src/tooltip/Tooltip.js
```diff
@@ -0,0 +1,215 @@
...
+class Tooltip extends React.PureComponent {
+  state = {
+    isVisible: false,
+    yOffset: 0,
+    xOffset: 0,
+    elementWidth: 0,
+    elementHeight: 0,
+  };
+
+  static defaultProps = {
```
iRoachie: Move this to the bottom under the proptype definitions - `Tooltip.propTypes`. It's much easier for readability

src/tooltip/Tooltip.js
```diff
@@ -0,0 +1,215 @@
...
+    withOverlay: true,
+    hightlightColor: 'transparent',
+    withPointer: true,
+    toggleOnPress: true,
+    height: 40,
+    width: 150,
+    contianerStyle: {},
```
iRoachie: Think you meant `containerStyle` instead of `contianerStyle`

src/tooltip/Tooltip.js
```diff
@@ -0,0 +1,215 @@
...
+    isVisible: false,
+    yOffset: 0,
+    xOffset: 0,
+    elementWidth: 0,
+    elementHeight: 0,
+  };
+
+  static defaultProps = {
+    withOverlay: true,
+    hightlightColor: 'transparent',
```
iRoachie: Typo: `highlightColor`

src/tooltip/Tooltip.js
```diff
@@ -0,0 +1,215 @@
...
+          style={{ borderBottomColor: pointerColor || backgroundColor }}
+          isDown={pastMiddleLine}
+        />
+      </View>
+    );
+  };
+  renderContent = withTooltip => {
+    const { popover, withPointer, toggleOnPress, highlightColor } = this.props;
+
+    if (!withTooltip)
```
iRoachie: Good job on early bail out, nice pattern 💯

src/index.d.ts
```diff
@@ -1305,6 +1305,76 @@ export interface SearchBarBase extends TextInputProperties {
...
+  /**
+   * function which gets called on opening the tooltip.
+   */
+
+  pointerColor?: string;
+  /**
+   * Color of tooltip pointer, it defaults to the backgroundColor if none passed .
+   */
+
+  toggleOnPress?: Function;
```
iRoachie: For functions the type:
`functionName(param: paramType): return value`

In our case:
`toggleOnPress?(): void;`

src/index.d.ts
```diff
@@ -1305,6 +1305,76 @@ export interface SearchBarBase extends TextInputProperties {
...
+  /**
+   * Color to highlight the item the tooltip is surrounding.
+   */
+
+  highlightColor?: string;
+  /**
+   * function which gets called on closing the tooltip.
+   */
+
+  onClose?: () => any;
```
iRoachie: Let's use the other syntax for all the props that are functions:

e.g
`onClose?(): void`

src/index.d.ts
```diff
@@ -1305,6 +1305,76 @@ export interface SearchBarBase extends TextInputProperties {
...
+   /**
+    * Flag to determine to toggle or not the tooltip on press.
+    */
+
+  popover?: React.ReactElement<{}>;
+  /**
+   * Component to be rendered as the display container.
+   */
+
+  containerStyle?: {}
```
iRoachie: `containerStyle` is a style for the `<View />` component - So the type is `StyleProp<ViewStyle>`

src/index.d.ts
```diff
@@ -1305,6 +1305,76 @@ export interface SearchBarBase extends TextInputProperties {
...
+   * Tooltip container height. Necessary in order to render the container in the correct place. Pass height according to the size of the content rendered inside the container.
+   */
+
+  width: number;
+  /**
+   * Tooltip container width. Necessary in order to render the container in the correct place. Pass height according to the size of the content rendered inside the container.
+   */
+
+
+  withOverlay?: boolean;
```
iRoachie: Also for booleans, put the default value in the description.

Example:

```ts
/**
 * Determines the shape of avatar
 *
 * @default false
 */
 rounded?: boolean;
```

src/tooltip/Tooltip.js
```diff
@@ -0,0 +1,215 @@
...
+}
+
+Tooltip.propTypes = {
+  children: PropTypes.element,
+  withPointer: PropTypes.bool,
+  popover: PropTypes.element,
+  toggleOnPress: PropTypes.bool,
+  height: PropTypes.number,
+  width: PropTypes.number,
+  containerStyle: PropTypes.any,
```
iRoachie: Use the style property from `ViewPropTypes`. See https://github.com/react-native-training/react-native-elements/blob/master/src/avatar/Avatar.js#L238 for example

AndreiCalazans:
Awesome @iRoachie thanks for reviewing. I'll make the fixes as soon as I get time for it.

AndreiCalazans:
Here you go, I made the fixes suggested and also added some basic initial tests.

src/helpers/index.js
```diff
@@ -1,6 +1,20 @@
+import { Platform, Dimensions } from 'react-native';
 import renderNode from './renderNode';
 import getIconType from './getIconType';
 import normalizeText from './normalizeText';
 import nodeType from './nodeType';
 
-export { renderNode, getIconType, normalizeText, nodeType };
+const Screen = Dimensions.get('window');
+const ScreenWidth: number = Screen.width;
+const ScreenHeight: number = Screen.height;
```
iRoachie: Can you remove the flow types here?

AndreiCalazans: lol sorry!

xavier-villelegier:
@AndreiCalazans If this can help:

It works when you add something right **after**  the `Tooltip` (before it doesn't work)

```jsx
  render() {
    return (
        <Tooltip popover={<Text>Info here</Text>}>
          <Text>Press me</Text>
        </Tooltip>
        <Text>whatever</Text>
      </View>
    )
  }
}

  container: {
    flex: 1,
    backgroundColor: '#fff',
    alignItems: 'center',
    justifyContent: 'center',
  },
})
```
image

But as long as the `Tooltip` is alone, it's broken. Don't know exactly why 🤔

AndreiCalazans:
Sorry @xavier-villelegier and @iRoachie, great job testing lol 
The pointer was only considering the Y offset of the main content's view and I wasn't checking which Y direction(Up or down) the tooltip was going. Now it checks the position of the tooltip and not the content. 

I also made another correction, very simple, just pass down the content's width and height  to the content being rendered inside the modal. This guarantees that we won't have any other bugs with positioning.

PR #1167, opened 2018-05-06
Doc improvement: remove duplicate item

x3388638:
for 0.19.0 `Button`
the `loading` item in props list is duplicate


iRoachie:
Good catch @x3388638. Thanks 👍

PR #1174, opened 2018-05-07
(ListItem) Customize padding with pad prop

### LINKED ISSUE #1119: [FEATURE REQUEST] Make padding value between ListItem components customizable
If you go into the code of `ListItem`, you'll see that there is a component named `PadView` which adds spacing between the leftComponent, the title component & right component. It has a static value of 16.

This should be customizable and user should be allowed to add this pad value

Gregjarvez:
closes #1119

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -980,6 +980,7 @@ export interface ListItemProps {
   bottomDivider?: boolean;
   scaleProps?: ScaleProps;
   ViewComponent?: React.ComponentClass;
+  pad?: number;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


Gregjarvez:
done !! ✔️

src/index.d.ts
```diff
@@ -980,6 +980,7 @@ export interface ListItemProps {
   bottomDivider?: boolean;
   scaleProps?: ScaleProps;
   ViewComponent?: React.ComponentClass;
+  pad: number;
```
iRoachie: `?` <- means it's optional

Gregjarvez: 😀how did I miss that

iRoachie:
LGTM!

PR #1175, opened 2018-05-07
Doc improvement: fix url hash conflict

x3388638:
for version 0.19.0 `Lists`
both `List` and `ListItem` have `containerStyle` props, however the two items link to the same url hash (#containerstyle)


iRoachie:
Thanks for catching this!

PR #1185, opened 2018-05-09
fix: onClear for default searchbar

LouisJS:
handle `onPress` method of clearIcon on `SearchBar-default`

### DIFF
--- a/src/searchbar/SearchBar-default.js
+++ b/src/searchbar/SearchBar-default.js
@@ -15,12 +15,6 @@ const defaultSearchIcon = {
   name: 'magnify',
   color: colors.grey3,
 };
-const defaultClearIcon = {
-  type: 'material-community',
-  size: 18,
-  name: 'close',
-  color: colors.grey3,
-};
 
 class SearchBar extends Component {
   focus = () => {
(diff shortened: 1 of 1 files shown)


LouisJS:
Unfortunately i have no idea how to test it.

```
it('should render with a custom methods', () => {
    const onChangeTextMock = jest.fn();
    const onClearMock = jest.fn();
    const component = shallow(
      <SearchBar onChangeText={onChangeTextMock} onClear={onClearMock} />
    );

    expect(component.length).toBe(1);
    expect(toJson(component)).toMatchSnapshot();
  });
```
Seems like it is already tested, isn't it ?

xavier-villelegier:
@LouisJS My bad, actually this test is useless for the `clear` method as the `clearIcon` is rendered when the Input isn't empty. I'll have to do another test for that anyway.

iRoachie:
We should probably nmake a list of things we need to write tests for 😄

xavier-villelegier:
@iRoachie Sure haha, I'll open an issue for that

LouisJS:
Wouhou. First contribution into this Open Source Project 💯
Hope it will the first of a long list 💪

PR #1204, opened 2018-05-17
Allow to set chevron size in ListItem

Amurmurmur:


martinezguillaume:
Yeah, it's a good idea ! I'll do it today or tomorrow

PR #1211, opened 2018-05-19
Avatar Placeholder

martinezguillaume:
(& I removed react-native-touchable-scale + fix renderNode)


src/helpers/renderNode.js
```diff
@@ -1,21 +1,21 @@
 import React from 'react';
 
 export default (Component, content, defaultProps) => {
-  if (content == null || content === false) {
+  if (content === null || content === false) {
```
iRoachie: Couldn't we do here `if (!content)` ?

martinezguillaume: No, because content can be equal to 0

iRoachie: Gotcha 👌🏽

PR #1213, opened 2018-05-20
Add some tests 

iRoachie:
Had to upgrade enzyme.

Added tests for `Button`


iRoachie:
Testing how long test will be :\ looking like might have to scrap this

PR #1218, opened 2018-05-22
Fix iOS SearchBar

### LINKED ISSUE #861: Use % in Input and SearchBars width
We should be using percentages in `Input` and `SearchBar`s width to avoid this weird behavior

simulator screen shot - ipad pro 10 5-inch - 2018-01-28 at 17 56 35

iRoachie:
Fixes #861 

kapture 2018-05-22 at 10 49 47

### DIFF
--- a/src/searchbar/SearchBar-ios.js
+++ b/src/searchbar/SearchBar-ios.js
@@ -2,7 +2,6 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import {
   Button,
-  Dimensions,
   LayoutAnimation,
   UIManager,
   StyleSheet,
@@ -16,7 +15,6 @@ import Input from '../input/Input';
 import Icon from '../icons/Icon';
 import { renderNode, nodeType } from '../helpers';
 
-const SCREEN_WIDTH = Dimensions.get('window').width;
 const IOS_GRAY = '#7d7d7d';
 const defaultSearchIcon = {
   type: 'ionicon',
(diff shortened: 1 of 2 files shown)


src/searchbar/SearchBar-ios.js
```diff
@@ -48,28 +56,26 @@ class SearchBar extends Component {
...
+    });
   };
 
   onBlur = () => {
     this.props.onBlur();
     UIManager.configureNextLayoutAnimation && LayoutAnimation.easeInEaseOut();
-    this.setState({ hasFocus: false });
+    this.setState({
+      hasFocus: false,
+      cancelButtonTransform: this.state.cancelButtonWidth,
```
xavier-villelegier: Just curious to know why is this initialised to 0 and set to `cancelButtonWidth` on blur ? Shouldn't we set it to 0 on blur or initialise it to `cancelButtonWidth` with the `onLayout` ?

iRoachie: It's using marginLeft for the button 
https://github.com/react-native-training/react-native-elements/pull/1218/files#diff-348f551797bae8f23921aab3699399eeR145

So on blur it needs to move back -> way to be hidden. So that's why it uses the buttonWidth and not 0

xavier-villelegier: Yea but I mean, the "blur" state is somehow the default state right, as it is not focused ?

So at the beginning the `marginLeft` will be 0 -> you focus -> you blur -> it will now be `cancelButtonWidth`. Is it the expected behaviour ?

iRoachie: I think I understand what you're saying. Curious doing it that way would cause an additional animation on mount, since the `cancelButtonTransform` would then be set to a value after the onlayout. Think that may have a blip in the UI.

xavier-villelegier: Yea but you could have put 0 here to have the same "slide" effect for all the focuses (the first and the followings). Here you have a more important "slide effect" for the next focus cause the `Button` is coming from further than on the first focus ([repro snack](https://snack.expo.io/HktaWifyQ))

PR #1221, opened 2018-05-23
Fix Header (+ Refactor)

martinezguillaume:


iRoachie:
Compared to the picture we have in the docs

src/header/__tests__/Header.js
```diff
@@ -34,13 +35,15 @@ describe('Header Component', () => {
       </Header>
     );
 
-    expect(component.find('Button').length).toBe(2);
+    expect(component.length).toBe(1);
+    expect(toJson(component)).toMatchSnapshot();
```
Gregjarvez: Should we be snapshotting here ?. Can we test the actually functionality describe by the test case 😉
`should render multiple children when passed in`

`expect(component.children().length).toBe(2)`

or more explicitly 

`expect(component.find('Button').length).toBe(2)`

same with the other cases :)

martinezguillaume:
@iRoachie It's fixed
@Gregjarvez I revert the test changes

iRoachie:
@martinezguillaume Mind doing the typescript definitions or you want me to do em?

martinezguillaume:
@iroachie I’ll do it, I forgot 🙂

xavier-villelegier:
Ping @martinezguillaume 🔔

Issue #1153, opened 2018-04-29
ListItem - cannot use TouchableHighlight as a component


nonotest:
### Description

1. Tried to use TouchableHighlight as a component
2. It should work
3. The test for the divider component generates a React element and we get the 
"expected to receive a single react element child" redbox

latest beta

### Reproduction Steps and Sample Code

```
 <ListItem
        component={TouchableHighlight}
/>
```
### Solution
Shouldn't the divider be purely handled via styles here? 
Possible solutions:
* Remove the divider and handle it with pure styling
* Use another wrapper at the top..

Please consider using TouchableHighlight as a default for this as its not uncommon to need underlineColor when using 'List Items'?

Thanks!

martinezguillaume:
Yep, I think the divider should be just handled by borderBottom style you‘re right ! And for the Component to TouchableHighlight, I think you‘re right too ! Thoughts @xavier?

xavier-villelegier:
@martinezguillaume Yea I think too. The `Divider` component might be a bit overkill here

iRoachie:
Live now in [1.0.0-beta6](https://github.com/react-native-training/react-native-elements/releases/tag/v1.0.0-beta6)!

PR #1233, opened 2018-05-28
Fix ListItem dividers + customs chevron & checkmark

martinezguillaume:
divider ref: https://github.com/react-native-training/react-native-elements/issues/1153
chevron ref: https://github.com/react-native-training/react-native-elements/pull/1204

Now, listItem divider is on the container style directly
So it's easier to override

You can customize chevron & checkmark
```jsx
<ListItem
  chevron={{ color: 'yellow', ...otherProps }}
/>
```
or set it to true to have default style (like before)
```jsx
<ListItem
  chevron
/>
```


src/list/ListItem.js
```diff
@@ -16,12 +16,23 @@ import Icon from '../icons/Icon';
 import Text from '../text/Text';
 import ButtonGroup from '../buttons/ButtonGroup';
 import Input from '../input/Input';
-import Divider from '../divider/Divider';
+import { dividerColor } from '../divider/Divider';
```
iRoachie: Prefer if you added this in the [colors.js](https://github.com/react-native-training/react-native-elements/blob/master/src/config/colors.js) file since it's used between two components. Other than that everything is 💯

iRoachie:
@martinezguillaume Snapshots!

iRoachie:
Reminding myself to update typescript definitions

PR #1240, opened 2018-05-30
Add overlayStyle prop to Card

YankeeSoccerNut:
Encountered "issue" with changing the overlayContainer style:
overlayContainer: {
    flex: 1,
    alignItems: 'center',
    backgroundColor: 'rgba(0, 0, 0, 0.2)',
    alignSelf: 'stretch',
    justifyContent: 'center',
    position: 'absolute',
    top: 0,
    left: 0,
    right: 0,
    bottom: 0,
  },

It could be an oversight or I may be missing something but the opacity was messing up the background.

PR code seemed to give me more control via the overlyStyle prop.

Thanks for taking a look and giving some feedback.

### DIFF
--- a/src/card/Card.js
+++ b/src/card/Card.js
@@ -22,6 +22,7 @@ const Card = props => {
     containerStyle,
     wrapperStyle,
     imageWrapperStyle,
+    overlayStyle,
     title,
     titleStyle,
     titleNumberOfLines,
--- a/src/card/Card.js
+++ b/src/card/Card.js
@@ -84,7 +84,11 @@ const Card = props => {
               {...imageProps}
             >
               {(featuredTitle || featuredSubtitle) && (
-                <View style={styles.overlayContainer}>
+                <View style={[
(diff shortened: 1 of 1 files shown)


iRoachie:
Thanks for sending this PR @YankeeSoccerNut, seems we always had the prop there but forgot to actually apply it onto the component.

You'll just need to update the snapshots and that will be all. 
https://react-native-training.github.io/react-native-elements/docs/testing.html#snapshot-testing

justColbs:
@iRoachie What does the following line resolve to? My confusion is why both variables are on either side of the `&&`.

`overlayStyle && overlayStyle,`

iRoachie:
Honestly it was apparent that we used very early in the library to check if a prop has been actually passed in, and only if it was passed in then apply. However I don’t think we actually need to do it anymore, but I’ll do some testing with it to confirm

iRoachie:
Closing as inactive

PR #1243, opened 2018-06-06
Bump docusaurus version to 1.1.5 for `next`

endiliey:
## Motivation

Current react-native-elements `next` branch website use the forked version of Docusaurus at `https://github.com/iRoachie/Docusaurus.git#rebase` (as of today it's currently still based on Docusaurus 1.1.1 and there is no additional commit made on top on the fork).

There are some bugs related to 1.1.1 & we have released many hotfix releases for that. Few notable bug fixes are for live reload bugs & Node 10 support. Overall, 1.1.5 is a stabler version.

With many major improvement and features that Docusaurus v1.1.5 offers, I think this is the right time to update to enhance docs editing experience, Another alternative is to rebase https://github.com/iRoachie/Docusaurus,git#rebase to 1.1.5. But I think using public npm package is much cleaner

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -9,7 +9,7 @@
     "rename-version": "docusaurus-rename-version"
   },
   "devDependencies": {
-    "docusaurus": "https://github.com/iRoachie/Docusaurus.git#rebase"
+    "docusaurus": "^1.1.5"
   },
   "version": "1.0.0-beta5"
 }
--- a/website/yarn.lock
+++ b/website/yarn.lock
@@ -9,6 +9,10 @@ accepts@~1.3.5:
     mime-types "~2.1.18"
     negotiator "0.6.1"
 
+address@1.0.3, address@^1.0.1:
+  version "1.0.3"
+  resolved "https://registry.yarnpkg.com/address/-/address-1.0.3.tgz#b5f50631f8d6cec8bd20c963963afb55e06cbce9"
+
 ajv@^5.1.0:
   version "5.5.2"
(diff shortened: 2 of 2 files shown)


iRoachie:
It isn’t as easy as you think. The reason we’re using a forked version is because Docusaurus doesn’t not include support for custom syntax highlighters. See https://github.com/facebook/Docusaurus/issues/438.

The built in one does not support JSX/React, which mean all our code snippets have no syntax highlights. So we forked it to use prism js instead of highlight js for the highlighter. 

This method is also being done by many libraries that are react based, and the original idea was from react navigation’s site. 

Haven’t had any problems with the current version though. Even thought we’d surely love to be up to date, we can’t do it at this time because of lack of features from the main version.

endiliey:
@iRoachie thank you for the explanation. Does that imply you'll be more happy to be up to date if we support custom syntax highlighter ? What other features you think are lacking ? Will be happy to see other requested features filed in Docusaurus issue.

iRoachie:
Yea definitely. It's a feature that was requested pretty early on and it would be great if I didn't have to do a rebase everytime we need to upgrade. That's the only feature we’d love to have currently. 

Any other features or bugs, I usually contact @JoelMarcey. He’s the one that helped us move over to Docusaurus

JoelMarcey:
@endiliey One thing we have considered in the past is to support two highlighters - `highlight.js` and `prism` and let the user choose as a configuration option.

I recall @hramos chose `highlight.js` for a specific reason when we were first developing Docusaurus.

endiliey:
@joelmarcey
I will look into it & work on a PR

PR #1246, opened 2018-06-07
Added check for none case in info props of PricingCard.

anushbmx:
As per documentation[1] the PriceCard should be able to work without any content in the info props, this fails because there is no condition check for its value before its its mapped.

*Updated test case for minimum props : removed 'info' as it should work without info.

[1] https://react-native-training.github.io/react-native-elements/docs/0.19.0/pricing.html

### DIFF
--- a/src/pricing/PricingCard.js
+++ b/src/pricing/PricingCard.js
@@ -47,7 +47,7 @@ const PricingCard = props => {
         >
           {price}
         </Text>
-        {info.map((item, i) => {
+        {info && info.map((item, i) => {
           return (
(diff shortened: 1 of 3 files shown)


anushbmx:
@xavier-villelegier 👍

anushbmx:
@iRoachie okay, I didn't knew that.

So you want me to do the changes on branch "next" and send pr to that branch or master ?

anushbmx:
@iRoachie new pr #1250

iRoachie:
Yes thank you!

PR #1250, opened 2018-06-09
Allow PricingCard to render without info prop

anushbmx:
As per documentation[1] the PriceCard should be able to work without any content in the info props, this fails because there is no condition check for its value before its its mapped.

*Updated test case for minimum props : removed 'info' as it should work without info.

[1] https://react-native-training.github.io/react-native-elements/docs/0.19.0/pricing.html

Performing PR on branch next as requested on #1246

### DIFF
--- a/src/pricing/__tests__/PricingCard.js
+++ b/src/pricing/__tests__/PricingCard.js
@@ -6,10 +6,7 @@ import PricingCard from '../PricingCard';
 describe('PricingCard component', () => {
   it('should render without issues', () => {
     const component = shallow(
-      <PricingCard
-        info={['1 User', 'Basic Support', 'All Core Features']}
-        button={{ title: 'GET STARTED', icon: 'flight-takeoff' }}
-      />
+      <PricingCard button={{ title: 'GET STARTED', icon: 'flight-takeoff' }} />
     );
 
     expect(component.length).toBe(1);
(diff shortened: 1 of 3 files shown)


src/pricing/__tests__/PricingCard.js
```diff
@@ -6,10 +6,7 @@ import PricingCard from '../PricingCard';
 describe('PricingCard component', () => {
   it('should render without issues', () => {
     const component = shallow(
-      <PricingCard
-        info={['1 User', 'Basic Support', 'All Core Features']}
-        button={{ title: 'GET STARTED', icon: 'flight-takeoff' }}
-      />
+      <PricingCard button={{ title: 'GET STARTED', icon: 'flight-takeoff' }} />
```
iRoachie: Instead of replacing this test, let's make a new test called `'it should render without info`

anushbmx: okay got it @iRoachie

iRoachie:
You can take as much commits as you need it's fine

anushbmx:
@iRoachie Done, I have update the changes and also test files.

iRoachie:
Hey @anushbmx I didn't see you make the change to `defaultProps` to add `info: []` there. That way we wouldn't have to do `info && info.map`

iRoachie:
Thanks again! 💯

PR #1026, opened 2018-03-20
Typescript tests on CI

iRoachie:
Ref #1010 

**Tests:**
- [ ] Avatar
- [x] Badge
- [ ] Button
- [ ] ButtonGroup
- [ ] Card
- [ ] CheckBox
- [x] Divider
- [ ] Header
- [ ] Icon
- [ ] Input
- [ ] ListItem
- [x] Overlay
- [x] PricingCard
- [ ] Rating
- [ ] SearchBar
- [ ] Slider
- [ ] SocialIcon
- [x] Text
- [ ] Tile

PR #1257, opened 2018-06-12
fix: define missing InputProp for Input

ossareh:
Make typescript match https://react-native-training.github.io/react-native-elements/docs/input.html

- inputContainerStyle was missing
- containerStyle comment was incorrect

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -691,10 +691,15 @@ export class Divider extends React.Component<DividerProps, any> {}
 
 export interface InputProps extends TextInputProperties {
   /**
-   * Styling for Input Component Container (optional)
+   * Styling for view containing the label, the input and the error message (optional)
    */
   containerStyle?: StyleProp<ViewStyle>;
 
+  /**
+   * Styling for Input Component Container (optional)
+   */
+  inputContainerStyle?: StyleProp<ViewStyle>;
+  
   /**
    * Displays an icon to the left (optional)
    */


ossareh:
Recently I used typings-tester to make it possible for [middyjs/middy](https://github.com/middyjs/middy/pull/200) to test that types are configured correctly. I'd be amenable to establishing the same base work for react-native-elements if y'all think it would be useful. LMK! (I'd do it as a separate arc of work).

iRoachie:
Hey thanks @ossareh for sending this in. We already made the patch and it'll be in the next release. https://github.com/react-native-training/react-native-elements/commit/9fc469e083d22c34c66c02db47f5239444c7f89c

iRoachie:
About the typescript tests, I had started putting it together in https://github.com/react-native-training/react-native-elements/pull/1026. But it's a huge tasks since you need to add code examples for each component and prop. It's definitely something I would like to pick back up though

ossareh:
Oh neat, sorry - I did search for fixes but I missed that!

RE: #1026 - you're right, it's a huge amount of work for a project of this size. That said, I think the infra in that PR is great, would it make sense to merge just the minimum amount and then begin scaling it up over time? At least in the case of 9fc469e we'd now have a test that validates the change :)

iRoachie:
You're right @ossareh I may just open it back after the 1.0 betas.

PR #1259, opened 2018-06-14
sizing of avatar

k3ithl1m:
Realized that there is no size prop for the newest update.


iRoachie:
Yes there is. It was added in the last beta. https://github.com/react-native-training/react-native-elements/releases/tag/v1.0.0-beta5

Issue #1260, opened 2018-06-14
Avatar: can set Image Component


sabativi:
**Is your feature request related to a problem? Please Describe.**
I want to be able to use a drop in replacement of Image like [Picache](https://github.com/benawad/Picache) which allows to have all of your avatar cached magically.

**Describe the solution you'd like**
Have a imageComponent={Picache} as a new prop for the Avatar Component

**Describe alternatives you've considered**
I do not see, I have seen https://github.com/react-native-training/react-native-elements/issues/336 but is is different

I can do this ( quickly ) as long as someone shows some interests on this.

iRoachie:
Definitely seems like a good idea! Feel free to go for it. Remember to base your PR on next

iRoachie:
Implemented now in [1.0.0-beta6](https://github.com/react-native-training/react-native-elements/releases/tag/v1.0.0-beta6)!

Sbphillips19:
@iRoachie how do you implement this feature?  I have tried doing something like this, but it doesn't work/ just gives me a blank avatar:

  ```
    const fastAvatar = () => {
        return (
          <FastImage source={{ uri: this.props.user.profilePicture }} />
        )
      }
```


```
<Avatar
          size="small"
          rounded
          // source={{ uri: this.props.user.profilePicture }}
          ImageComponent={fastAvatar}
          activeOpacity={0.7}
        />
```

**Comment by iRoachie** [comment]
You can pass it directly.

```jsx
<Avatar ImageComponent={FastImage} />
```

Sbphillips19:
@iRoachie Thanks!  That worked, but I am getting an error using it inside of listItem.  Should it work inside of leftAvatar?  I was hoping to do the same- 

> ImageComponent: {FastImage}

iRoachie:
It works the same way.

```jsx
<Avatar leftAvatar={{ ImageComponent: FastImage}} />
```

codakkk:
TypeScript definitions are not working properly with this.

![image](https://user-images.githubusercontent.com/8694404/109716463-368d9200-7ba5-11eb-9c44-4096b60f3617.png)

josegoval:
> TypeScript definitions are not working properly with this.
> 
> ![image](https://user-images.githubusercontent.com/8694404/109716463-368d9200-7ba5-11eb-9c44-4096b60f3617.png)
> 
> Am I doing something wrong here?

By now I found this solution disabling ts for that line: Comment over the code
```jsx
[...]

PR #1261, opened 2018-06-14
Custom ImageComponent for Avatar

sabativi:
As discussed in #1260


iRoachie:
Thanks again for contributing

PR #1263, opened 2018-06-15
Change a few links that did not link anywhere

lwurl:
A few links changed the URL but did not change the location on the page


iRoachie:
Thanks again! 💯

PR #1269, opened 2018-06-18
Button: add the possibility to pass activeOpacity value

HristoEftimov:
The `Button` component is using the default `activeOpacity` which is `0.2`. It will be nice if the developer can pass a custom value for it. 

At the moment, only the `clear: true` property updates it to `0`, which is not enough :/


HristoEftimov:
@iRoachie, yes, you are right. It's better to move it above the `{...attributes}`. Please, check my last commit :)

xavier-villelegier:
Hey guys @HristoEftimov @iRoachie  `{...atributes}` is already overriding all props, you were looking the `master` branch. This has been fixed on `next` one month ago in https://github.com/react-native-training/react-native-elements/commit/ae0c4132f56a40b67803672392cbf482c0da70a1

HristoEftimov:
@xavier-villelegier okay, but in this case, you cannot pass a custom value for the `activeOpacity` option. Please check: https://github.com/react-native-training/react-native-elements/pull/1269/commits/aae0423e2a6aa6c84aa9cdde417f9955bad02320

iRoachie:
@HristoEftimov In the source that @xavier-villelegier posted, `activeOpacity` would be passed in as apart of the `attributes` object

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #1274, opened 2018-06-21
Allow No Borders in ButtonGroup

MrOggy85:
Update ButtonGroup.js to allow 0 values in `innerBorderStyle.width`.

Add a check for undefined instead of falsy value.

### DIFF
--- a/src/buttons/ButtonGroup.js
+++ b/src/buttons/ButtonGroup.js
@@ -75,17 +75,20 @@ const ButtonGroup = props => {
                 borderRightWidth:
                   i === 0
                     ? 0
-                    : (innerBorderStyle && innerBorderStyle.width) || 1,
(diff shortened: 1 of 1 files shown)


iRoachie:
What's the purpose of this PR? You can already disable the borders. https://snack.expo.io/SyC5XvY-m

iRoachie:
Sorry about closing this. Just realized what was happening when the check && was checking against 0.

MrOggy85:
No worries. I learned from this PR that you can set the value to `-1` to disable the borders. I just thought it was more intuitive if `0` was allowed. Please include this if you like it.

iRoachie:
And you're correct. -1 was only a cheap fix. I fixed it properly here https://github.com/react-native-training/react-native-elements/pull/1301

PR #1248, opened 2018-06-09
Add TextSegment component

suwu150:
add text-distinction component,divide the character string according to a certain character and display it in different styles after segmentation

PR #1295, opened 2018-07-06
add  textSegment in sidebars.json

suwu150:
add  textSegment in sidebars.json


iRoachie:
You don't need to make a new PR, you can add it to the one you have already open #1248

PR #1279, opened 2018-06-22
Fixing opencollective issue

drdpedroso:
nstalling react-native-elements was breaking inside a "dockerized" node env.
Replacing opencollective to opencollective-postinstall fixed the issue

Referenced [issue](https://github.com/opencollective/opencollective-cli/issues/3)

PR #1298, opened 2018-07-11
Reduce opencollective package size

nano-c:
The opencollective package added too much footprint. Replaced with a much smaller package opencollective-lite (200KB vs 24MB) which provides the same functionality for post install script.


iRoachie:
Don't think will merge this. There's no repo for that package, and if someone goes wrong, we'd have nowhere to turn to for solutions. If anything we'll look into https://github.com/react-native-training/react-native-elements/pull/1279

Issue #1136, opened 2018-04-24
"RawText "" must be wrapped in an explicit <Text> component." for errorMessage props of Input


hpelitebook745G2:
### Description

1. Assigned variable to errorMessage props for `Input` component.
2. Initially this variable is empty.
3. Upon rendering it gives `RawText "" must be wrapped in an explicit <Text> component.` error.

Current version is `"react-native-elements": "beta"`, `1.0.0-beta4:` specifically.

### Reproduction Steps and Sample Code

This is my local state: `this.state = {
	  emailError: {
		isError: false,
		message: '',
	  },
	};`

While this is the UI part:
`	<Input
	  errorStyle={{ color: 'red' }}
	  errorMessage={this.state.emailError.message}
	  shake={this.state.emailError.isError}
	/>`

On runtime, `this.state.emailError.errorMessage` will be supplied with a string. However upon initial rendering, I already got the error. How am I supposed to assign a variable to the errorMessage if I get this kind of error?

Any help would be accepted.

Thanks.

xavier-villelegier:
@hpelitebook745G2 Not sure if this comes from the error message. Can you add more context to your use case ? [Here is](https://snack.expo.io/HJMEo5n3z) a simple example of how to use it, it works fine on Snack

vsai:
@xavier-villelegier I see the same error message as above when the value passed into `errorMessage` is an empty string. Not sure if that has something to do with it.

As soon as I modify the initial state of the component, containing the errorMessage, to non-empty, the component renders fine.

xavier-villelegier:
@vsai Do you have this error only on Android ?

hpelitebook745G2:
I fixed this by using `null` value instead of having empty string. So my condition would be:
`
	errorMessage={ this.state.emailError.isError ? this.state.emailError.message : null }
`

vsai:
@xavier-villelegier I had this error on the iOS simulator.

I fixed it by using null instead of empty string too.
But just pointing out why I was seeing the same error too.

xavier-villelegier:
@vsai @hpelitebook745G2 
Okay guys, after investigating a bit, you're right, this is a bug.

Here is the source code:
https://github.com/react-native-training/react-native-elements/blob/d4da856a7f105c814cc162e507b48c49399cabef/src/input/Input.js#L109-L113

By doing this, when `errorMessage=""`, `errorMessage && …` returns `""`, and it's trying to render an empty string outside a `Text` component.

[...]

PR #1299, opened 2018-07-11
Fixed <Input> bug that caused errorMessage and label to render an empty string outside of <Text>

kirakik:
Addressed #1136 by checking for an empty string as well as null using the !! (not not) operator.

### DIFF
--- a/src/input/Input.js
+++ b/src/input/Input.js
@@ -75,7 +75,7 @@ class Input extends Component {
 
     return (
       <View style={[{ width: '90%' }, containerStyle]}>
-        {label && <Text style={[styles.label, labelStyle]}>{label}</Text>}
+        {!!label && <Text style={[styles.label, labelStyle]}>{label}</Text>}
         <Animated.View
           style={[
             styles.inputContainer,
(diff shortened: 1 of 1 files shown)


iRoachie:
Thanks for sending this in! Looks pretty good to me, just gonna run a test to make sure it doesn't render 'false' sometimes.

Gregjarvez:
`!!("   ")` is `true`. This is not a solid check for empty string. I recommend a regex.

kirakik:
@Gregjarvez `!!(" ")` is not an "empty string" though, a space is considered a character

Gregjarvez:
@kirakik they indeed are. but do we want to render the node with white spaces ?

kirakik:
@Gregjarvez it depends on the use cases we're trying to cover with this. It's common to set the default value of a text to an empty string `""`, but not so common to set it to white spaces `" "`. 

I can see a use case where someone would want to render the `errorMessage` and the `label` regardless of whether or not there is one. For example, two inline `<Input />`s, which need to stay aligned whether or not they both have error messages and/or labels. If we guard against white spaces, this would be difficult to achieve.

iRoachie:
Spaces are valid input. If the user doesn’t want to accept spaces then they can specify that in their on onchange handle

PR #1305, opened 2018-07-16
Hoist the isFocused() method from the react native input element

howey-aus:
As per other component methods (focus(), blur()) etc that are passed-through from the contained React Native component, this will simply return the result of isFocused() in that component.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -774,6 +774,13 @@ export class Input extends React.Component<InputProps, any> {
   focus(): void;
 
   /**
+   * Calls isFocused() on the Input
+   *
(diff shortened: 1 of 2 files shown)


iRoachie:
Hey @howey-aus thanks for sending this in. You'll need to make 1 additional change as well.

**Typescript definitions:**
https://github.com/react-native-training/react-native-elements/blob/master/src/index.d.ts#L749
You'll need to add `isFocused(): boolean` to this definition.

howey-aus:
Thanks @iRoachie I've added that type annotation.

iRoachie:
Thanks again!

Issue #1325, opened 2018-07-26
Name prop for FormInputs


giacomocerquone:
**Is your feature request related to a problem? Please Describe.**
Using a FormInput (or TextInput) I don't have the possibility to use a name prop. This is an unbelievable native missing feature and an opportunity for react-native-elements to integrate this thing.

**Describe the solution you'd like**
I'd like to be able to get somehow through onChange (overriding the native method maybe?) the name of the FormInput that spawned the event.

**Additional context**
I give you my usual context:

I have a form with 10 values.
The container components maintain the internal state and pass down to the presentational 10 setMethods to set the respective values in the container state (they need to be ten right because we can't use a name prop to distinguish which one fired the change text event)

xavier-villelegier:
@giacomocerquone I agree, I often do a custom wrapper to avoid having too many `onChange` functions for each input. This can be a nice enhancement. If anyone wants to tackle this issue, feel free to submit a PR 👍

giacomocerquone:
I'll happily submit it, assign it to me! I'll write it asap

xavier-villelegier:
Awesome, you rock ! 🎸

iRoachie:
Even though I get what you're saying. I'm wondering why you don't pass back up an object. This is a pattern that's used on web as well.

```js
// Container

...
updateValue = data => {
  this.setState(data)
}
...


// Children
<Input value={this.state.email} onChangeText={email => this.updateValue({ email })} />

<Input value={this.state.name} onChangeText={name => this.updateValue({ name })} />

```

Am i misunderstanding your use case?

giacomocerquone:
This is not much different than what we are doing. It's true though that you pass one function from the container to the presentational, but in this way you cannot avoid to define a new function at every render (and that is expensive in a large application I'm maintaining). Then to solve this you would define multiple function inside a presentational component and that's something you don't wanna do usually because:

1. you've got something like 50% faster components through babel transpiler when you're using a functional component
2. you're introducing some kind of logic in a pres. comp. and you don't want that

iRoachie:
Not disagreeing with you, just wanted to get your use case. Technically your argument could hold for any **controlled** component. Should we add name props to all of them?

giacomocerquone:
I'd say yes, it should hold for any controlled component since it's a react native lack.
[...]

PR #1333, opened 2018-07-28
Deleted unused var and first attempt to implement name prop in input

giacomocerquone:
According to #1325 this is a first naive attempt to resolve it (didn't even test the code). Naive because I want first to address and discuss some technical details about the "feature" I'm about to implement.

The thing I want to discuss is somehow similar to what semanticUI discussed here https://github.com/Semantic-Org/Semantic-UI-React/issues/623

So the roads I saw are these:

1. We "override" the onChange native method of Input.
  1.1 If we do this, how do we pass the name (or other data we may add in the future)? Through a second parameter? If yes, through the use of what structure? (object, directly a var, an array etc.) 
  1.2 We edit directly the synthetic event of react setting a "e.target.name" manually on the event proxy.

2. We define a new function different from onChange that returns the onChange event plus the name (or any other data). Then the options to pass the name through this new function are those defined above.

I already made a choice in this PR and I'll try to explain why, then we'll discuss pros and cons.

**Option 2**
I believe that creating a new function with a different name would confuse users and make somehow less intuitive the usage of it. So I went for the "override" of the onChange method.

**Option 1.2**
This sounds initially like a nice idea since we would adapt and standardize this thing in order to let devs use this feature just like they use it in the browser. The thing is that this would mean to step over react native's feet: we shouldn't take the place of react native developers and this means that we shouldn't edit core specs.

**Option 1.1**
This is the option I chose for exclusion.
Then I decided to go and pass an object because we may want to enhance the Input component even further in the future and so it would be silly to use a simple var parameter and not define an object or an array that can store multiple values.
Discarded the array idea since naming what are you giving back to the user code is important the structure left was the object.

Tell me yours


iRoachie:
Hey @giacomocerquone can you rebase this from `next`? They're a few conflicts and file changes

giacomocerquone:
I'm on vacation! I'll do it asap

iRoachie:
Hey @giacomocerquone, sorry for late reply. Let's get this wrapped up. 

Still outstanding are:
* Tests - Simple test that when `onChange` is called that name is in the object
* Typings, and Docs - Since we're adding 2 new props you'll need to add them

Let me know if you need any assistance.

giacomocerquone:
@iRoachie doing it just now. I noticed one thing though, you have components like SearchBarBase where I don't know why you don't extend your own input, instead the react-native one. This means that if a user would like to use this feature with that component, he can't and this sounds like a incoherence in the lib.

For now I'd say to change just the input component, ended that, we can refactor the others too

iRoachie:
@giacomocerquone Not sure exactly, but you can get an idea from the typescript definitions. https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-native/index.d.ts#L1169

iRoachie:
@giacomocerquone Still interested in working on this?

giacomocerquone:
being in vacation I should have some time to push forward this pr, sorry for the delay

PR #1350, opened 2018-08-04
Fix linear gradient button disabled state

### LINKED ISSUE #1346: Disabled styles for button don't work with LinearGradient
**Explain what you did**
Show disabled version of the Button component based on state. When not disabled, it has a linear gradient

**Expected behaviour**
Disabled styles should show when the disabled prop is true

**Describe the bug**
The LinearGradient props show regardless of the disabled prop

**To Reproduce**
https://snack.expo.io/ryqlhi1SQ

**Possible Solution**
Change the ViewComponent of the Button back to `View` when `disabled` is true, and conditionally apply `linearGradientProps`

| software | version |
| -- | -- |
| react-native-elements | 1.0.0-beta5 |
| react-native | 0.56.0 |
| npm or yarn | yarn 1.9.2 |

Gregjarvez:
fixes #1346 

Instead of using `global.expo.LinearGradient` when button is disabled. We default to using a `View` as `ViewComponent` since `disabledStyles.backgroundColor` cannot take precedence over `linearGradientProps.colors`

https://snack.expo.io/B1kSEgmSm


iRoachie:
Hey @Gregjarvez you'll need to rebase from next

ManigandanRaamanathan:
What's the code to use then?

PR #1353, opened 2018-08-07
Do not use index as a key

robinpokorny:
Using index as a key is a bad practice and should be avoided. Please see https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318 for details.

Since many devs would just copy and paste the code, I believe it is exceptionally important to promote good practices in the docs.

### DIFF
--- a/website/versioned_docs/version-0.19.0/lists.md
+++ b/website/versioned_docs/version-0.19.0/lists.md
@@ -27,11 +27,11 @@ const list = [
 
 <List containerStyle={{marginBottom: 20}}>
   {
-    list.map((l, i) => (
+    list.map((l) => (
       <ListItem
         roundAvatar
         avatar={{uri:l.avatar_url}}
-        key={i}
+        key={l.name}
         title={l.name}
       />
     ))
@@ -58,9 +58,9 @@ const list = [
 
 <List>
   {
-    list.map((item, i) => (
+    list.map((item) => (
       <ListItem
-        key={i}
+        key={item.title}
         title={item.title}
         leftIcon={{name: item.icon}}
       />


robinpokorny:
Sure, that would be the best, then adding an `id` would solve it.

However, with current data, this is the best we can do.

luissmg:
What you mean with `current data`?
From my point of view, the id should come from something like [uuid](https://www.npmjs.com/package/uuid) (already does what we need, no need to program it again). Isn't this viable here?

robinpokorny:
With current data, I mean the list created in the example. Like this:

```js
const list = [
  {
    name: 'Amy Farha',
    avatar_url: 'https://s3.amazonaws.com/uifaces/faces/twitter/ladylexy/128.jpg',
    subtitle: 'Vice President'
  },
  ... // more items
]
```

There is no id. It can be replaced by, for example:

```js
const list = [
  {
	id: '001'
    name: 'Amy Farha',
    avatar_url: 'https://s3.amazonaws.com/uifaces/faces/twitter/ladylexy/128.jpg',
    subtitle: 'Vice President'
  },
  ... // more items
]
```

Or, of course, using some id generationg library.

luissmg:
But that was what I just said (or I think I did). Why do you limit yourself to the use of the data from an item?

robinpokorny:
OK! ;) Then we are on the same page. I tried to use the current structure to make the change smaller.

iRoachie:
Thanks guys! 💯

luissmg:
@robinpokorny @iRoachie Did the changes I was talk been made? Cause if they weren't there is a problem again...

iRoachie:
They're made, but they'll show when we do a next release.

iRoachie:
In fact, i'll merge it to master as well

iRoachie:
Changes live!

luissmg:
@iRoachie No no, I mean that the changes he made in first place solves on problem and creates another. If I pass the same title/name for 2 different items in the list  there will be 2 items with the same key. Or did I miss something?

luissmg:
I mean... Forget it, it is just an example at the end 😆 The developer must check the errors when they occur.

iRoachie:
Lol you're right! But for that example based on the data, `name` should be enough

luissmg:
Yes, suits the example and that's what matters in this case! 👍

PR #1361, opened 2018-08-09
fix: add missing "size" prop to CheckBox types

mklopets:
🖖 Heya. The `size` prop was missing in CheckBox's type defs.


iRoachie:
Thanks! Good catch 💯

PR #1366, opened 2018-08-13
feat(Icon): Add disabled and disabledStyle

iRoachie:
Adds `disabled` and `disabledStyle` props to the Icon component. 

Ideally, we should do this for all our interactable components. It's a common use case for like forms, where when submitting the fields are disabled.

Usage:

```jsx
<Icon disabled />
<Icon disabled disabledStyle={{ backgroundColor: 'pink' }} />
```

### DIFF
--- a/src/icons/Icon.js
+++ b/src/icons/Icon.js
@@ -1,5 +1,5 @@
-import PropTypes from 'prop-types';
 import React from 'react';
+import PropTypes from 'prop-types';
 import {
   Platform,
   TouchableHighlight,
@@ -22,6 +22,8 @@ const Icon = props => {
     raised,
     containerStyle,
     reverseColor,
+    disabled,
+    disabledStyle,
     onPress,
     component: Component = onPress ? TouchableHighlight : View,
     ...attributes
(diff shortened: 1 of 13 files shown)


src/icons/Icon.js
```diff
@@ -46,11 +48,17 @@ const Icon = props => {
             alignItems: 'center',
             justifyContent: 'center',
           },
-        ]}
+          disabled && styles.disabled,
+          disabled && disabledStyle,
+        ])}
+        {...onPress && { disabled }}
```
Gregjarvez: just to improve readability should should we put the conditional statement in parens ? ` {...(onPress && { disabled })}`

iRoachie: Its an object so we have to spread it. What this is doing is only adding disabled to the props if onPress is passed in.

Gregjarvez: yeah we do spread. I am just saying we wrap that expression in parens to enhance readability 😉

iRoachie: Oh lol, prettier removes unnecessary brackets.

PR #1378, opened 2018-08-20
Optionally render a Badge over a Button

howey-aus:
Adds a badge prop to a button (similar to how a badge is applied to a ListItem).

I've often found myself needing to display a button that has a simple indicator (that renders over part of the icon). Rather than manually render a Badge, this allows one to just supply a badge prop to a button.

eg:

`
<Button 
      icon={...}
      badge={{ value: 42, containerStyle: {...}, wrapperStyle: {...} }}
/>
`


iRoachie:
@xavier-villelegier @Gregjarvez @martinezguillaume Thoughts?

Gregjarvez:
I don't think this is a good idea. We should leave the badge out of the button component. Consumers should put that in there if they want

iRoachie:
Closing as we're not going with this

PR #1385, opened 2018-08-23
Avatar component to render multiple image sources

dev-dano:
## Summary
* Source prop can be either an image source to render a single image, or an array of image sources to render up to 4 images.

## Use case
* Group Chat

## Example
```js
// Avatar Example
<Avatar
  rounded
  size='large'
  source={{ uri: '...' }} OR source={ [{ uri: '...' }, { uri: '...' }, { uri: '...' }] }
  ...
/>

// ListItem Example
<ListItem
  hideChevron
  roundAvatar
  avatar={{ uri: '...' }} OR avatar={ [{ uri: '...' }, { uri: '...' }, { uri: '...' }] }
  ...
/>
```

### DIFF
--- a/src/avatar/Avatar.js
+++ b/src/avatar/Avatar.js
@@ -137,10 +137,10 @@ const Avatar = ({
       {...attributes}
     >
       <FadeInImage
-        enableFadeIn
+        enableFadeIn={enableFadeIn}
         PlaceholderContent={PlaceholderContent}
         containerStyle={overlayContainerStyle}
-        ImageComponent={renderImageComponent}
+        renderImageComponent={renderImageComponent}
       />
       {Utils}
     </Component>
@@ -256,7 +256,6 @@ Avatar.defaultProps = {
     underlayColor: DEFAULT_COLORS[0],
     style: null,
   },
-  ImageComponent: Image,
 };
 
 
@@ -292,7 +291,7 @@ const MultiAvatar = ({
             key={index}
             style={{ marginBottom: 1, width: imageWidth, height: imageHeight }}
             source={source}
-            onLoadEnd={onLoadEndCounter}
+            onLoadEnd={onLoadEnd && onLoadEndCounter}
           />
         );
       })}
@@ -322,26 +321,29 @@ class FadeInImage extends React.PureComponent {
       enableFadeIn,
       PlaceholderContent,
       containerStyle,
-      ImageComponent,
-      ...attributes
+      renderImageComponent
     } = this.props;
     
-    const PlaceHolder = enableFadeIn ? Animated.View : View;
+    let PlaceHolder = View;
+    let ImageComponent = renderImageComponent();
+    let opacity = 0;
+    
+    if (enableFadeIn) {
+      PlaceHolder = Animated.View;
+      ImageComponent = React.cloneElement(ImageComponent, { onLoadEnd: this.onLoadEnd });
+      opacity = this.placeholderContainerOpacity;
+    }
     
     return (
       <View style={[styles.overlayContainer, containerStyle]}>
-        {
-          enableFadeIn ?
-            React.cloneElement(ImageComponent(), { onLoadEnd: this.onLoadEnd }) :
-            ImageComponent()
-        }
+        {ImageComponent}
         <PlaceHolder
           style={[
             styles.placeholderContainer,
-            { opacity: enableFadeIn ? this.placeholderContainerOpacity : 0 }
+            { opacity }
           ]}>
           {PlaceholderContent}
-        </Animated.View>
+        </PlaceHolder>
       </View>
     );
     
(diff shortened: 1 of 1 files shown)


iRoachie:
This is really cool idea! What do you think guys? @xavier-villelegier @Gregjarvez @martinezguillaume

martinezguillaume:
That’s a really cool feature !! I agree with that !

iRoachie:
Hey @dev-dano, would really like to get starting to integrate this. Can you first rebase your PR from the `next` branch? That's where we do all our development and right now there's a merge conflict. Let me know if you need any help

Gregjarvez:
I love this !!!!!!!!

dev-dano:
Awesome! I am glad everyone loves this feature! Just a quick question @iRoachie, since I worked directly on master branch, should I rebase master onto next? I'm new to Git and still getting used to concepts haha

xavier-villelegier:
Wow I love it ! Must have feature. Thanks for this PR @dev-dano ✌️

iRoachie:
Yes, exactly @dev-dano.

dev-dano:
It's done! I'm also updating the doc. @iRoachie

dev-dano:
Hey @iRoachie, it is updated now. 

_Main Changes:_
1. Added MultiAvatar feature 
2. Made fade in effect available on Android platform (Tested on android virtual device and worked beautifully. It can still be disabled by sending prop, enableFadeIn={false})

iRoachie:
So it's not available on iOS you're saying?

xavier-villelegier:
I remember @martinezguillaume did something about the fade in avatar

dev-dano:
Originally, fade in was enabled on iOS and disabled on android. It is now enabled on both platforms by default, and consumer has an option to disable it using enableFadeIn prop.

Note: disabling fade in effect helps stabilizing performance when you have a large list of Avatars or ListItems.

martinezguillaume:
Fade in image is enabled by default in Android, you can control it with fadeDuration props. That’s why you don’t see it in the code 🙂

dev-dano:
Oh, my mistake! 😅 Yet it was a change I had to make since fadeDuration prop won't do the work for multiAvatar. We want to fade in a view containing multiple images rather than each image separately.

We could stick with the current changes or handle **two cases** separately. Use fadeDuration prop in android **when rendering a single image source**, and use onReachEnd function **when rendering multiple images**. Any suggestion? @martinezguillaume

martinezguillaume:
I have an interesting suggestion I think, maybe we can use Avatar inside MultiAvatar. You see what I mean ? And user can change props of each avatar if he wants.

We can use a getProps pattern. It’ll be a prop function called getAvatarProps with the index of the avatar in argument and return a props object for this specific avatar.
I’ll do an API example later today !

I don’t know if it’s the right solution but maybe MultiAvatar is a different component than Avatar 😯

iRoachie:
That's another suggestion. How hard would it be to replicate this behaviour with a separate component

martinezguillaume:
Maybe we can do even simpler, like this:
```js
<MultiAvatar separatorColor="yellow" separatorWidth={2}>
  <Avatar ... />
  <Avatar ... />
  // ... (up to 4)
</MultiAvatar>
```
I think in theory it's not hard at all. And I like that the logic of render multiple avatar is not in the Avatar component.

dev-dano:
That is a great suggestion! In that case, could we give MultiAvatar an option to render a single placeholder for multiple Avatars? Originally, what I wanted was a single placeholder content for multiple images. Do you know what I mean?

iRoachie:
Yea i think I get you, like one placeholder that waits until all images are loaded, and then it switches out. Would be cool to have the option

dev-dano:
Hey guys, just want to clarify this before start working on it. So, the feature I suggested is very different from @martinezguillaume's suggestion.

- My suggestion is more like Avatar with multiple image & title sources.
ex)
// showEditButton=true
 <- Outer view gets 1 edit button and when pressed, onEditButton will fire.

- martinezguillaume's suggestion is more like a View containing many Avatars.
ex)
 <- Each Avatar inside gets an edit button and when pressed, onEditButton corresponding to the pressed avatar will fire.

They both have good use cases. Just wanted to double check. Please let me know!

iRoachie:
Can we maybe do a combination of both?

It can still be possible to have MultiAvatar as it's own component, but allow it to have, `onPress`, `showEditButton` etc.

dev-dano:
Yes! In that case, I suggest creating separate components for Utils and PlaceholderContent so we could reuse them in MultiAvatar.

**Usage Example**
```js
<MultiAvatar
  rounded
  size='large'
  separatorColor="yellow"
  separatorWidth={2}
  onPress={...}
  showEditButton={...}>

  <Avatar ... />
  <Avatar ... />
  // ... (up to 4)

  OR

  { data.map((item, i) => <Avatar ... />) }
</MultiAvatar>
```

iRoachie:
Definitely for that approach!

PR #1386, opened 2018-08-23
On start rating event

OtavioBraga:
When using a rating component inside a ScrollView on iOS, the scrollview don't lock the scroll causing the rating component to get very laggy.

demo gif

The onStartRating callback is useful on iOS because I can lock the scroll via the scrollEnabled prop on the start rating callback.

### DIFF
--- a/docs/rating.md
+++ b/docs/rating.md
@@ -21,6 +21,7 @@ ratingCompleted(rating) {
 <Rating
   showRating
   onFinishRating={this.ratingCompleted}
+  onStartRating={this.ratingStarted}
   style={{ paddingVertical: 10 }}
 />
 
@@ -32,6 +33,7 @@ ratingCompleted(rating) {
   readonly
   imageSize={40}
   onFinishRating={this.ratingCompleted}
+  onStartRating={this.ratingStarted}
   style={{ paddingVertical: 10 }}
 />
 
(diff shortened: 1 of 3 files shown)


iRoachie:
How would you use this?

OtavioBraga:
Here is a short example of how i'm using it.

```es6

class Form extends React.Component {
  constructor (props) {
    super(props)
    this.state = { scrollEnabled: true }
  }

  onStartRating = () => {
    this.setState({ scrollEnabled: false })
  }

  onFinishRating = (itemValue) => {
    const { setFieldValue } = this.props
    setFieldValue('attendanceRating', itemValue)
    this.setState({ scrollEnabled: true })
  }

  render () {
    return (
      <ScrollView scrollEnabled={this.state.scrollEnabled} >
        <View>
        <Text>Como você avalia o seu atendimento na Unimed Caruaru?</Text>
        <Rating
          type='heart'
          ratingCount={5}
          startingValue={0}
          imageSize={40}
          onFinishRating={this.onFinishRating}
          onStartRating={this.onStartRating}
        />
      </View>
      </ScrollView>
    )
  }
}


```

I think that are other ways to use onStartRating. Maybe to fire some animation. For me, blocking the scroll of the view was a good reason since I was using 5 rating components os the same screen.

OtavioBraga:
Sure, I’ll do it as soon as possible.

OtavioBraga:
@iRoachie I've made the requested changes. 😄

iRoachie:
Thanks again!

PR #1404, opened 2018-09-03
input width changed from 90% to 100%

kevintamarus:
input width changed to 100%, input is offsetting to the left and not centered when at 90%

### DIFF
--- a/src/input/Input.js
+++ b/src/input/Input.js
@@ -74,7 +74,7 @@ class Input extends Component {
     });
 
     return (
-      <View style={[{ width: '90%' }, containerStyle]}>
+      <View style={[{ width: '100%' }, containerStyle]}>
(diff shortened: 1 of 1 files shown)


piu130:
Doesn't look well but much better than before :) Would be nice to have a 90% width centered input but can't find a way to do this.

iRoachie:
I think the main problem is that the placeholder and text is inset from the container, and that looks weird.

So maybe we can do 100% width on the container style, and remove the marginLeft 10 from the inputStyle

iRoachie:
Hey @kevintamarus  we just released beta6 so there are some conflicts. Can you resolve them or rebase from next?

iRoachie:
Closing as inactive. Feel free to message or create a new issue if this is still a concern.

PR #1409, opened 2018-09-07
[ListItem] Add android as default for list styling

piu130:
Currently there is no styling for `ListItem` in `react-native-web` (and other platforms) because they are only applied to `android` and `ios`.

I changed `android` to `default` so android style is applied.

`react-native-web` example:

Before:
before
After:
after

### DIFF
--- a/src/list/ListItem.js
+++ b/src/list/ListItem.js
@@ -194,7 +194,7 @@ const styles = StyleSheet.create({
       ios: {
         padding: 14,
       },
-      android: {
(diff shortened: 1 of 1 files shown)


piu130:
Yep. I tested it in an android emulator. You can find the code for `Platform.select` [here](https://github.com/facebook/react-native/blob/master/Libraries/Utilities/Platform.android.js#L29). It takes `android` if exist, otherwise `default`.

PR #1410, opened 2018-09-08
package: Use opencollective-postinstall for smaller install size

LinusU:
The `opencollective` dependency is huuuuuuuge ([**11.6 MB**](https://packagephobia.now.sh/result?p=opencollective)). This patch replaces it with the officially supported smaller package which is just the postinstall hook:

https://github.com/opencollective/opencollective-postinstall

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "test:ci": "jest --runInBand",
     "test:watch": "jest --watch",
     "test:coverage": "jest --coverage",
-    "postinstall": "opencollective postinstall",
+    "postinstall": "opencollective-postinstall",
     "lint": "eslint src/",
     "precommit": "lint-staged && npm test",
     "prettify": "prettier --write 'src/**/*.js'"
@@ -35,7 +35,7 @@
   "dependencies": {
     "lodash.isempty": "^4.4.0",
     "lodash.times": "^4.3.2",
-    "opencollective": "^1.0.3",
+    "opencollective-postinstall": "^2.0.0",
     "prop-types": "^15.5.8"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


LinusU:
Also, the reason for noticing this at all is because it caused a conflict with the `regenerator-runtime` shipped with react-native, which caused a build issue for me. Basically what happened was that `regenerator-runtime` wouldn't get hoisted and my app would not be able to require it:

```text
my-project@0.0.0 /Users/linus/coding/my-project
├─┬ react-native@0.56.0
│ ├─┬ fbjs-scripts@0.8.3
│ │ └─┬ babel-core@6.26.3
│ │   └─┬ babel-runtime@6.26.0
│ │     └── regenerator-runtime@0.11.1  deduped
│ └── regenerator-runtime@0.11.1 
└─┬ react-native-elements@1.0.0-beta5
  └─┬ opencollective@1.0.3
    └─┬ babel-polyfill@6.23.0
      └── regenerator-runtime@0.10.5 
```

iRoachie:
Ahh they've been a few PRs for this. I don't mind taking this change. Btw, Do you know of any of them that work with yarn? Even the current one doesn't output the message.

For the other maintainers @xavier-villelegier @Gregjarvez @martinezguillaume this is the change in terminal output.

Before:
screen shot 2018-09-10 at 10 09 57 am

After:
screen shot 2018-09-10 at 10 13 47 am

LinusU:
> Btw, Do you know of any of them that work with yarn? Even the current one doesn't output the message.

If that is the case I think that `yarn` doesn't execute `postinstall` at all, and there is probably nothing we can do about that.

LinusU:
ping @iRoachie, any updates on this? ☺️

iRoachie:
Was just waiting on someone else to look at this, if nothing by weekend, i'll merge

LinusU:
Thanks!

PR #1413, opened 2018-09-10
feat(Input): add inputComponent prop (closes #1408)

### LINKED ISSUE #1408: Allow global/local overriding of TextInput element used?
Would be nice to always be able to override the native elements used by the components:

For instance if you want to use a custom `TextInput`, in my case I wanted to try https://github.com/zackify/react-native-autofocus, which exports an HOC to the native `TextInput`. Alas, there is no way to override the `TextInput` component used by your `Input` component…

Possible API for local override:
```js
import { TextInput } from 'react-native-autofocus'
<Input inputComponent={TextInput} />
```

It is something that is already possible for the `Icon` component. Would be great to have this as well for `Input`.

Could quickly work on a PR if there is consensus.

---

Edit: looks like the Icon component no longer have an `inputComponent` prop. Only a `component` prop which will be the WrapperView used.

Looking at Button there is a `ViewComponent` prop. Looks like the API needs to be unified on this. (lowercase/camelcase, Component vs. ViewComponent`.

IMO, not fond of the `ViewComponent` upperfirst case pattern: a component isn't necessarily a class (can be a simple function).

Also it's strange that the `component` prop for the Icon does end up being the wrapper component used (and not the Icon one as expected by the API).

mgcrea:
- Adds initial unit test for the `Input` component,

- Closes #1408 (support for `inputComponent` prop to pass a custom `TextInput` compatible component)

Thanks!

### DIFF
--- a/docs/input.md
+++ b/docs/input.md
@@ -47,20 +47,21 @@ import { Input } from 'react-native-elements';
 
 > This component inherits [all native TextInput props that come with a standard React Native TextInput element](https://facebook.github.io/react-native/docs/textinput.html), along with the following:
 
-* [`containerStyle`](#containerstyle)
-* [`inputContainerStyle`](#inputcontainerstyle)
-* [`errorMessage`](#errormessage)
-* [`errorStyle`](#errorstyle)
-* [`errorProps`](#errorprops)
-* [`inputStyle`](#inputstyle)
-* [`label`](#label)
-* [`labelStyle`](#labelStyle)
-* [`labelProps`](#labelprops)
-* [`leftIcon`](#lefticon)
-* [`leftIconContainerStyle`](#lefticoncontainerstyle)
-* [`rightIcon`](#righticon)
-* [`rightIconContainerStyle`](#righticoncontainerstyle)
-* [`shake`](#shake)
+- [`containerStyle`](#containerstyle)
+- [`inputContainerStyle`](#inputcontainerstyle)
+- [`errorMessage`](#errormessage)
+- [`errorStyle`](#errorstyle)
+- [`errorProps`](#errorprops)
+- [`inputComponent`](#inputComponent)
+- [`inputStyle`](#inputstyle)
+- [`label`](#label)
+- [`labelStyle`](#labelStyle)
+- [`labelProps`](#labelprops)
+- [`leftIcon`](#lefticon)
+- [`leftIconContainerStyle`](#lefticoncontainerstyle)
+- [`rightIcon`](#righticon)
+- [`rightIconContainerStyle`](#righticoncontainerstyle)
+- [`shake`](#shake)
 
 ---
 
@@ -113,6 +114,16 @@ props to be passed to the React Native `Text` component used to display the erro
 |                                      Type                                       | Default |
 | :-----------------------------------------------------------------------------: | :-----: |
 | {[...Text props](https://facebook.github.io/react-native/docs/text.html#props)} |  none   |
+
+---
+
+### `inputComponent`
+
+component that will be rendered in place of the React Native `TextInput` (optional)
+
+|          Type          |  Default  |
+| :--------------------: | :-------: |
+| React Native Component | TextInput |
 
 ---
 
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -740,6 +740,11 @@ export interface InputProps extends TextInputProperties {
   rightIconContainerStyle?: StyleProp<ViewStyle>;
 
   /**
+   * Renders component in place of the React Native `TextInput` (optional)
+   */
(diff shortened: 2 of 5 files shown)


iRoachie:
Could you rebase your PR from `next`? There have been a few changes between `master` and `next` .

mgcrea:
@iRoachie rebase done, thanks for the quick follow up.

src/input/Input.js
```diff
@@ -133,6 +133,7 @@ Input.propTypes = {
   rightIconContainerStyle: ViewPropTypes.style,
 
   inputStyle: Text.propTypes.style,
+  inputComponent: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
```
iRoachie: Why is PropTypes.string used here?

mgcrea: > `func` covers types from "classic" react components, ES6 classes, and stateless function components. `string` covers the native element case (eg "div").

From https://stackoverflow.com/a/33087573

Not sure if it does apply to react-native though, (maybe react-native-web?).

mgcrea:
@iRoachie Is there anything I can do to have this merged? Thanks!

src/input/Input.js
iRoachie: Ah okay, wasn't sure.

iRoachie:
Sure thing, still left to do is:
- docs
- typescript definitions

For the docs you have to add the property. See https://react-native-training.github.io/react-native-elements/docs/0.19.0/contributing.html#changes-for-an-existing-doc-on-a-version-not-yet-released

For typescript definitions, this prop needs to be added to [this interface](https://github.com/react-native-training/react-native-elements/blob/next/src/index.d.ts#L711)

mgcrea:
@iRoachie done, thanks for the kind guiding.

src/index.d.ts
```diff
@@ -739,6 +739,11 @@ export interface InputProps extends TextInputProperties {
    */
   rightIconContainerStyle?: StyleProp<ViewStyle>;
 
+  /**
+   * Renders component in place of the React Native `TextInput` (optional)
+   */
+  inputComponent?: React.ComponentClass<any> | React.StatelessComponent<any>;
```
iRoachie: Just use the component class

mgcrea: Not proficient with typescript sor not sure I understand correctly, do you mean to use `React.Component`?

If my understanding is correct, `React.Component` is for a rendered component (eg. `<TextInput />`) while `React.ComponentClass` is for a component constructor (eg. `TextInput`) which is what we need here.

Thanks.

iRoachie: Sorry, meant leave only `React.ComponentClass<any>`

mgcrea: Is the a reason why we should forbid pure stateless components?

iRoachie: Stateless components can't have refs, so methods we have on the `Input` component such as focus, blur etc will fail

mgcrea: Did not know that, it's fixed, thanks.

docs/input.md
```diff
@@ -116,6 +117,16 @@ props to be passed to the React Native `Text` component used to display the erro
 
 ---
 
+### `inputComponent`
+
+component that will be rendered in place of the React Native `TextInput` (optional)
+
+|              Type              | Default |
+| :----------------------------: | :-----: |
+| Component (string \| function) |  none   |
```
iRoachie: Change this definition here. See https://react-native-training.github.io/react-native-elements/docs/0.19.0/button.html#component for example. The default would also be TextInput

src/index.d.ts
mgcrea: In fact I have a use-case for stateless components (I'm using them right now in my app), where I replace the TextInput with a `Touchable` component (that triggers a picker) to build a `Select` component:

```js
const SelectInputComponent = props => {
  const {onFocus, value, placeholder} = props;
  return (
    <TouchableWithoutFeedback onPress={onFocus}>
      <View style={{flex: 1, height: '100%'}}>
        <Text style={{paddingTop: 20, paddingLeft: 10, color: value ? 'black' : '#86939e'}}>
          {value || placeholder}
        </Text>
      </View>
    </TouchableWithoutFeedback>
  );
};
```

I think we should also allow these custom stateless components even if they don't work like a TextInput at all.

iRoachie: What i'm saying is that if you use a stateless component you will get errors by we are assigning a ref to it.

Also is there any reason you changed the formatting on the whole file?

mgcrea:
@iRoachie I've made all the required changes, I think we are good to go 🚀.

Tell me if there is anything else I could/should do.

Thanks.

PR #1420, opened 2018-09-11
Update checkbox.md

Cazaimi:
There is an option to set `borderColor` of the checkbox, but was not exposed in the documentation. Added the listing.

### DIFF
--- a/docs/checkbox.md
+++ b/docs/checkbox.md
@@ -232,3 +232,11 @@ Specify different font family
 |  Type  |                      Default                      |
 | :----: | :-----------------------------------------------: |
 | string | System font bold (iOS), Sans Serif Bold (android) |
+
+### `borderColor`
+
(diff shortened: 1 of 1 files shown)


iRoachie:
There's no borderColor prop on the CheckBox component.

Cazaimi:
@iRoachie , I made the mistake of adding it as a prop. Actually, when you pass `borderColor` as a key in `containerStyle`, it seems to make the border's color change. There was no way of knowing that for me, apart from digging around the code.

iRoachie:
All the props for the CheckBox are here https://react-native-training.github.io/react-native-elements/docs/0.19.0/checkbox.html#props. 

The [Customization docs](https://react-native-training.github.io/react-native-elements/docs/0.19.0/customization.html) explain how to style components.

Are you suggesting they're not clear?

Issue #1419, opened 2018-09-11
Make Android style default


piu130:
**Is your feature request related to a problem? Please Describe.**
Currently there is no consistent style applied to components for `react-native-web` (Platform `web`), `electron` (Platform `desktop`), etc.

**Describe the solution you'd like**
We should set `android` style as default for all platforms.

**Describe alternatives you've considered**
We could set `ios` as default or create a third `default` style.

**Additional context**
Some examples:

- [Avatar](https://github.com/react-native-training/react-native-elements/blob/master/src/avatar/Avatar.js#L173) has no shadow
- Button is a bit random: [here](https://github.com/react-native-training/react-native-elements/blob/master/src/buttons/Button.js#L186) and [here](https://github.com/react-native-training/react-native-elements/blob/master/src/buttons/Button.js#L196) not android and [here](https://github.com/react-native-training/react-native-elements/blob/master/src/buttons/Button.js#L208) and [here](https://github.com/react-native-training/react-native-elements/blob/master/src/buttons/Button.js#L224) no style
- [Button](https://github.com/react-native-training/react-native-elements/blob/master/src/checkbox/CheckBox.js#L123) text is not bold
- ...

iRoachie:
Sounds like a great idea, Can you make a PR to do this?

iRoachie:
Also could you outline the steps you did to get react-native-web working? Would love to put it in the docs.

piu130:
Created a PR [here](https://github.com/react-native-training/react-native-elements/pull/1421).

I use [neutrino](https://neutrinojs.org/) for the setup. Just create a new project as described [here](https://neutrinojs.org/installation/create-new-project.html) and select react.

Then add `react-native-web`, `babel-preset-react-native`, (`react-art` - is optional but for my setup it fails without it), `react-native-elements` and `react-native-vector-icons` with your package manager.

Currently you must add `.web.js` and `.web.jsx` to webpacks resolve extensions (see `.neutrinorc.js` below). But you can remove it with the next neutrino release (fixed [here](https://github.com/neutrinojs/neutrino/pull/1064)).

Then you must add the react-native modules to webpacks include (see `.neutrinorc.js` below), because webpack does not transpile `node_modules` (webpack assumes they are already transpiled).

My `.neutrinorc.js` looks like this:

```js
const { join } = require('path');

const nodeModulesDir = join(__dirname, 'node_modules')

module.exports = {
  use: [
    [
      '@neutrinojs/react',
[...]

PR #1421, opened 2018-09-11
[WIP] added android as default style

piu130:
As discussed [here](https://github.com/react-native-training/react-native-elements/issues/1419) the PR that makes android style as default for all Platforms that do not override the style, expect `ios`.


piu130:
There are several places like [this](https://github.com/react-native-training/react-native-elements/blob/master/src/header/Header.js#L140), where we have a one liner checking for ios. Should I change it to `Platform.select` or leave it?
This is when the same style (`height` here) is applied to all platforms.

iRoachie:
I think maybe we can leave it. The Platform.select with an empty iOS object looks weird

piu130:
This exmple would change from
```js
height: Platform.OS === 'ios' ? 70 : 70 - getStatusBarHeight()
```
to
```js
...Platform.select({
  ios: {
    height: 70
  },
  default: {
    height: 70 - getStatusBarHeight()
  },
}),
```

How would you prevent an empty iOS object? :)

iRoachie:
Not sure it matters, I don't have a preference once the behaviour is the same. Does getStatusBarHeight() work for RNW?

piu130:
@iRoachie I added default `getStatusBarHeight` value to 0. Now it works.

piu130:
How do I fix codecov? I added tests for `getStatueBarHeight` but it does not change anything.

iRoachie:
Hey @piu130 sorry for the conflicts, had to merge some older PRs that fixed bugs. Can you resolve them?

piu130:
@iRoachie merged. But codecov still not :)

iRoachie:
Hey @piu130 just updating you. Gonna push this back to v1.0.0 because i know more conflicts will come up. Will roll out beta6 this week and then we can continue from there

iRoachie:
Hey @piu130 we just released beta6 so there are a ton of conflicts. Can you resolve them?

piu130:
@iRoachie merged.

piu130:
Need to wait for [this](https://github.com/ovr/react-native-status-bar-height/pull/8) PR. Then bump `react-native-status-bar-height` and remove `statusBar.js`.

piu130:
@iRoachie is it ok if I create a PR for each component so we could merge the small ones faster?

iRoachie:
That works!

iRoachie:
Cause now on next we use [react-native-status-bar-height](https://github.com/react-native-training/react-native-elements/blob/next/package.json#L41) So we don't need have statusBar.js anymore

iRoachie:
Closing this since our convo.

PR #1435, opened 2018-09-23
feat(icons): Add support for FontAwesome5

iRoachie:
Adds basic support for FontAwesome5.

We need to find a way to support the font weights though. FontAwesome supports different weights https://github.com/oblador/react-native-vector-icons/blob/master/FONTAWESOME5.md#usage

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,12 +60,12 @@
     "react": "16.3.1",
     "react-dom": "16.2.0",
     "react-native": "0.55.4",
-    "react-native-vector-icons": "^4.4.2",
+    "react-native-vector-icons": "^5.0.0",
     "react-test-renderer": "16.3.1",
     "webpack": "^2.2.1"
   },
   "peerDependencies": {
-    "react-native-vector-icons": "^4.2.0 || ^5.0.0"
+    "react-native-vector-icons": "^5.0.0"
   },
   "jest": {
     "preset": "react-native",
--- a/.ci/setupTests.js
+++ b/.ci/setupTests.js
@@ -1,4 +1,8 @@
 import { configure } from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 
+jest.mock('react-native-vector-icons/FontAwesome5', () => {
+  return console.log;
+});
+
 configure({ adapter: new Adapter() });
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -60,12 +60,12 @@
...
     "react-dom": "16.2.0",
     "react-native": "0.55.4",
-    "react-native-vector-icons": "^4.4.2",
+    "react-native-vector-icons": "^5.0.0",
     "react-test-renderer": "16.3.1",
     "webpack": "^2.2.1"
   },
   "peerDependencies": {
-    "react-native-vector-icons": "^4.2.0 || ^5.0.0"
+    "react-native-vector-icons": "^5.0.0"
```
ahilles107: Why you disallow v4 usage?

iRoachie: Because it'll break if version below 5 is used. FontAwesome5 is only in version 5. So the import will break the app

ahilles107: Then imo You should replace current font awesome with version 5 in code instead adding it as new font provider. With this change You can't install locally fontawesome4 and use it.

iRoachie: That's not true, react-native-vector-icons v5 supports both font-awesome@4^ and 5. https://github.com/oblador/react-native-vector-icons#bundled-icon-sets. 

You'll only need to update the react-native-vector-icons package when you update react-native-elements.

likern:
I think that is not a right direction to move. What about using **Font Awesome Pro** or other fonts? The problem is that we need to be able to pass custom properties which are specific for exact **Icon** component. **light / solid / brand** for **Font Awesome 5 / Pro ** case.
I proposed other solution - #1480

iRoachie:
@likern what is not the right direction to move? Passing custom properties to the Icon is an **enhancement** over this. As i mentioned in the first comment, we're still looking for the correct way to do this. 

This PR is very much a WIP

likern:
> @likern what is not the right direction to move? Passing custom properties to the Icon is an **enhancement** over this.
It might be my bad, I overlooked this. Yes, supporting **Awesome Font 5** with **Awesome Font 5 Pro** in mind is desired.

Slapbox:
Did anyone ever figure out a way to support Font Awesome 5 Pro?

PR #1449, opened 2018-10-06
Add custom styles to iOS SearchBar cancel button

tux-tn:
This PR adds two attributes to ```cancelbuttonprops``` in SearchBar IOS: 

- ```buttonStyle``` :  Cancel button custom styles
- ```buttonTextStyle``` : Cancel button's text custom styles

 The main goal is to allow customization of cancel button since react-native's ```Button``` doesn't allow that


iRoachie:
Hey thanks for sending the PR. i think we were going to tackle this by using the Button component from RNE. @xavier-villelegier can you confirm?

tux-tn:
@iRoachie do you want me to update my PR and use RNE Button or should i close it? BTW, i am using the exact same styling as RN ```Button``` component

iRoachie:
@xavier-villelegier

xavier-villelegier:
I think it's okay to use a simple TouchableOpacity for now, we don't need all the Button's props.
But maybe we should use the `primary` color of the theme ?

tux-tn:
@xavier-villelegier Don't you think we should keep default iOS color and it's material design equivalent since it's an iOS searchbar?

xavier-villelegier:
@tux-tn Actually, as you said it's an iOS SearchBar, so I was just wondering if the Button should look different on iOS and Android ? I mean, why should the SearchBar be all the same accross platforms except the button ? 🤔 Maybe we should just keep the iOS style on both platforms (cc @iRoachie)

iRoachie:
Sounds good to me. A change needs to be made anyway cause right now on android the cancel button is the big blue material raised button

xavier-villelegier:
Yea that's why @tux-tn work is necessary to make it consistent between platforms, actually you just need to remove the `Platform.select` from the styles and keep the iOS one

tux-tn:
I agree with that, i will make the necessary changes

src/searchbar/SearchBar-ios.js
```diff
@@ -155,11 +164,33 @@ class SearchBar extends Component {
...
             onPress={this.cancel}
-            {...cancelButtonProps}
-          />
+            disabled={buttonDisabled}
+            {...otherCancelButtonProps}
+          >
+            <View
+              style={[
+                buttonColor &&
+                  Platform.OS === 'android' && { backgroundColor: buttonColor },
```
xavier-villelegier: Actually as we said you shouldn't have any differences between iOS and Android here (no background on iOS, neither on Android)

src/searchbar/SearchBar-ios.js
```diff
@@ -155,11 +164,33 @@ class SearchBar extends Component {
...
+                buttonStyle,
+              ]}
+            >
+              <Text
+                style={[
+                  styles.buttonTextStyle,
+                  buttonColor &&
+                    Platform.OS === 'ios' && { color: buttonColor },
```
xavier-villelegier: Same goes for this one, both iOS and Android should have this color

tux-tn:
The reason i kept this conditional styling is to follow RN Button documentation. We need to either keep that or change searchbar-ios documentation to stop referencing RN Button props as props for cancel button. What do you think ? 



Le 18 nov. 2018 à 1:56 PM, à 1:56 PM, "Xavier Villelégier" <notifications@github.com> a écrit:

xavier-villelegier:
Yea I see what you mean, let's update the doc then. The iOS SearchBar with an android `Button` looks weird in my opinion, it's better to keep all the iOS styling for the "ios" SearchBar

tux-tn:
I removed Platform specific styling. I also added props for disabled cancel Button styling (The main purpose of my PR was to add the ability to customize the cancel Button).

tux-tn:
Thank you @xavier-villelegier, should i merge next updates and resolve conflicts?

xavier-villelegier:
Let's merge that and wait for some feedbacks to polish it.

Thanks @tux-tn, congrats for your first-time contribution to RNE 👏 🥇

iRoachie:
Just realize we merged this with adding typescript definitions...

iRoachie:
Tracked over here https://github.com/react-native-training/react-native-elements/issues/1606

PR #1454, opened 2018-10-10
Add Theming

iRoachie:
Implements #216 

Finally, after countless ages, theming is here 🎉!

Not just colours but:
### EVERY, SINGLE, PROP

Exports three new utils:
- ThemeProvider
- ThemeConsumer
- withTheme

### Preview
kapture 2018-10-09 at 20 57 23

### Breaking Changes
This is a massive breaking change, all apps will need up wrap their Top level component with the `<ThemeProvider />`


Gregjarvez:
this is so awesome !

iRoachie:
Alright! ThemeProvider now optional if you don't want theming. That's means no more breaking change!

Monte9:
@iRoachie yes!! that's huge! Thanks for putting in the extra work to make that possible, I know a lot of users would appreciate that.

PR #1457, opened 2018-10-11
Add TextStyle to possible checkmark types.

janhesters:
This type is missing. The example of the changelog for the `<ListItem />`:

```ts
// Before
<ListItem checkmark checkmarkColor="red" />

// After
<ListItem checkmark={{ color: 'red' }} />
```

Currently throws an error. After this fix, it doesn't anymore.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1022,7 +1022,7 @@ export interface ListItemProps {
   contentContainerStyle?: StyleProp<ViewStyle>;
   rightContentContainerStyle?: StyleProp<ViewStyle>;
   chevron?: boolean | IconProps;
-  checkmark?: boolean | IconProps;
+  checkmark?: boolean | IconProps | TextStyle;
   onPress?(): void;
   onLongPress?(): void;
   title?: string | React.ReactElement<{}>;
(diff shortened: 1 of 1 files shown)


iRoachie:
Strange, IconProps has color in it https://github.com/react-native-training/react-native-elements/blob/ea502132e724c8af182ce03c764c5e77778276b5/src/index.d.ts#L912

janhesters:
Did you see my screenshot in the slack? The linter complains 🤷🏻‍♂️

I also tried what happens when you use a regular object like `{ color: "red" }`:

screen shot 2018-10-11 at 19 57 32

This also throws an error. Probably because even though `IconProps` contains color, the linter doesn't consider an object with just color of type `IconProps`, because the required `name` key is missing.

iRoachie:
Ahh i see, that isn't the correct fix though. The problem is that name is required on IconProps. 

Really the definition should be:
```ts
checkmark?: boolean | Partial<IconProps>;
```

janhesters:
Changed it.

iRoachie:
Could you also test some of the other props that use IconProps? I’m assuming it’s more than just chevron

janhesters:
Yes I could and I will. I'm just busy right now, but I will try it out!

PR #1462, opened 2018-10-12
1.0.0-beta7 - (Beta6 regression)

iRoachie:
Incorrectly published 1.0.0-beta6 as a **stable** version on npm. Since we can't use version numbers, we had to unpublish beta6 and publish a new beta


Monte9:
Thanks for jumping on this quickly @iRoachie! 💯

PR #1474, opened 2018-10-17
Added AntDesign to icons in src/helps/getIconType.js

d-p-s:
I noticed that Ant Design was missing from the switch statement in this code, so I imported Ant Design and added it to the switch statement to fix it. This is my first time making a pull request like this so apologies if I've messed anything up.


iRoachie:
It’s not as easy as adding this line, I think we may also have to bump our dependency of react native vector icons

iRoachie:
Closing in favour of #1529

Issue #1415, opened 2018-09-10
Allow 'array' or 'object' prop type for Tile component


beltrancaliz:
**Explain what you did**
I have an array of icons as the caption prop for the Tile component (v0.19.0), but I get warnings as the documentation states it should expect a string prop.

**Expected behaviour**
Tile component accepts JSX elements (array or object) as valid props for the caption.

**Describe the bug**
The tile component does not accept objects or arrays as props for the caption.

**To Reproduce**
(Note: warning will not appear on 'Tap to Play')  https://snack.expo.io/HkO77l4uQ
Have an icon or an array of icons as the caption prop for the Tile component on version 0.19.0.
Example: 
```
<Ionicons
        name='ios-star'
        size={18}
        style={{color: "#FFD700"}}
/>
```

**Possible Solution**
Allows icons and any other JSX elements as valid prop types.

**Screenshots**
![image](https://user-images.githubusercontent.com/21249916/45299730-7abee080-b4da-11e8-9c65-307b74ec80ba.png)
![image](https://user-images.githubusercontent.com/21249916/45299733-7eeafe00-b4da-11e8-9d29-91286f90b103.png)

**Your Environment (please complete the following information):**

| software | version |
| -- | -- |
| react-native-elements | 0.19.0 |
| react-native | 0.52.0 |
| npm or yarn | npm |

iRoachie:
Are you suggesting a feature? The `caption` prop for the Tile component only accepts strings.

beltrancaliz:
@iRoachie I suppose I am requesting a feature then, accepting JSX elements should be a common case for the caption props.

PR #1510, opened 2018-10-26
feat(Tile): Allow caption to be a React element

MHase:
PR created to solve [issue #1415](https://github.com/react-native-training/react-native-elements/issues/1415) created by [beltrancaliz](https://github.com/beltrancaliz)
I've added possibility to pass object as propTypes to Tile/FeaturedTile instead of passing only simple string.
Also `captionStyle` won't work if you'll pass an object, because you can style it without this prop :)

### DIFF
--- a/docs/tile.md
+++ b/docs/tile.md
@@ -102,13 +102,13 @@ Text inside the tilt when tile is featured
 
 |  Type  | Default |
 | :----: | :-----: |
-| string |  none   |
+| string **OR** React element or component |  none   |
 
 ---
 
 ### `captionStyle`
 
-Styling for the caption (optional)
+Styling for the caption (optional); use this along with `caption` type `string`
 
 |      Type      | Default |
 | :------------: | :-----: |
--- a/src/tile/FeaturedTile.js
+++ b/src/tile/FeaturedTile.js
@@ -10,9 +10,16 @@ import {
 } from 'react-native';
 
 import { ViewPropTypes, BackgroundImage, withTheme } from '../config';
+import { renderNode } from '../helpers';
 
 import Text from '../text/Text';
 import Icon from '../icons/Icon';
+
+const renderText = (content, defaultProps, style) =>
+  renderNode(Text, content, {
+    ...defaultProps,
+    style: StyleSheet.flatten([style, defaultProps && defaultProps.style]),
+  });
 
 const FeaturedTile = props => {
   const {
@@ -116,18 +123,13 @@ const FeaturedTile = props => {
           >
             {title}
           </Text>
-          {typeof caption === 'string' ? (
-            <Text
-              style={StyleSheet.flatten([
-                styles.text,
-                captionStyle && captionStyle,
-              ])}
-            >
-              {caption}
-            </Text>
-          ) : (
-            caption
-          )}
+          {typeof caption === 'string'
+            ? renderText(
+                caption,
+                { style: captionStyle },
+                StyleSheet.flatten(styles.text)
+              )
+            : caption}
         </View>
       </BackgroundImage>
     </TouchableOpacity>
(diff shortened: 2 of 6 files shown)


src/tile/Tile.js
```diff
@@ -122,7 +122,7 @@ const Tile = props => {
 Tile.propTypes = {
   title: PropTypes.string,
   icon: PropTypes.object,
-  caption: PropTypes.string,
+  caption: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
```
iRoachie: This will still throw an error if you pass a component to it. Look at PropTypes.element or PropTypes.node. https://reactjs.org/docs/typechecking-with-proptypes.html

iRoachie:
Also would you mind adding a test when `caption` is just a string?

src/tile/Tile.js
MHase: I tried it in react-native-element-app and it worked quite well without any errors. Also when I passed simple `<Text>Test</Text>` the `typeof` it was still an `object`

iRoachie: Did you use a Text component from react-native or the native one?

docs/tile.md
```diff
@@ -102,13 +102,13 @@ Text inside the tilt when tile is featured
...
 | :----: | :-----: |
-| string |  none   |
+| string **OR** React element or component |  none   |
 
 ---
 
 ### `captionStyle`
 
-Styling for the caption (optional)
+Styling for the caption (optional); use this along with `caption` type `string`
```
iRoachie: Let's change the wording here to be bit more clear.

```suggestion
Styling for the caption (optional);  You only use this if `caption` is a string
```

src/tile/Tile.js
MHase: I tried Text from react-native and react-native-elements
Also Tried with 
```
<View>
  <Avatar
    size="large"
    source={{uri: "https://s3.amazonaws.com/uifaces/faces/twitter/brynn/128.jpg"}}
    onPress={() => console.log("Works!")}
    activeOpacity={0.7}
  />
  <Text>Test</Text>
</View>
```
and still `typeof caption` is equal `string`

MHase: Maybe we can change it to `PropTypes.node`
```
Anything that can be rendered: numbers, strings, elements or an array
(or fragment) containing these types.
```
and it will accept everything (strings and components) needed without any problems :)

iRoachie: Alright then

src/tile/FeaturedTile.js
```diff
@@ -116,14 +116,18 @@ const FeaturedTile = props => {
...
           </Text>
-          <Text
-            style={StyleSheet.flatten([
-              styles.text,
-              captionStyle && captionStyle,
-            ])}
-          >
-            {caption}
-          </Text>
+          {typeof caption === 'string' ? (
```
iRoachie: Is it possible that you can use the approach here https://github.com/react-native-training/react-native-elements/blob/master/src/list/ListItem.js#L37

We have a helper for doing this kind of thing.

src/tile/FeaturedTile.js
```diff
@@ -116,14 +123,13 @@ const FeaturedTile = props => {
...
           </Text>
-          <Text
-            style={StyleSheet.flatten([
-              styles.text,
-              captionStyle && captionStyle,
-            ])}
-          >
-            {caption}
-          </Text>
+          {typeof caption === 'string'
```
iRoachie: You don't need to do the check here, `renderNode` already does it

src/tile/FeaturedTile.js
```diff
@@ -116,14 +123,11 @@ const FeaturedTile = props => {
...
-              styles.text,
-              captionStyle && captionStyle,
-            ])}
-          >
-            {caption}
-          </Text>
+          {renderText(
+            caption,
+            { style: captionStyle },
+            StyleSheet.flatten(styles.text)
```
iRoachie: You can remove the flatten here, since it’s only one style

iRoachie:
Thank you 💯

Issue #1371, opened 2018-08-15
Click star to rate


mbackermann:
**Is your feature request related to a problem? Please Describe.**
No

**Describe the solution you'd like**
It would be nice to instead of gesture to rate, the user could only click the star to rate.

MHase:
PR created for this issue #1521 :)

PR #1521, opened 2018-10-29
Rating on single star press

MHase:
Add functionality to rate on press on single star (bell, rocket or sth else).
Minimal value is 1 and max is max number of 'stars'

Issue: #1371


iRoachie:
Closing this since we're using the Rating component is now being used from react-native-ratings. https://github.com/react-native-training/react-native-elements/pull/1671

PR #1528, opened 2018-10-30
Remove expo ci from PRs

xavier-villelegier:
I just removed the line from the `.travis.yml` and keep the `.ci` dir, in case we want to add this again later


iRoachie:
I guess we could remove the file as well, it's on git anyway.

PR #1530, opened 2018-10-30
Fix small listitem docs error.

janhesters:
The docs in the listitem incorrectly say that the default implementation of the item is a `TouchableOpacity` but it is a `TouchableHighlight`.


iRoachie:
Thank you! Gratz on first contribution

Issue #1424, opened 2018-09-12
Allow Input Label to accept React component


arthurfranca:
I'm using Input component v1.0.0-beta5. When i pass a fragment or text component as a label prop, i get a warning saying ```Invalid prop 'label' of type 'object' supplied to 'Input', expected 'string'```.

Currently i can have a label like ```I am a label``` (string)
It would be nice not getting a warning when i have one of these too as label prop:
```<>I am a label <Text style={...}>Optional</Text></> ```
```<React.Fragment>I am a label <Text style={...}>Optional</Text></React.Fragment> ```
```<Text>I am a label <Text style={...}>Optional</Text></Text>```

The solution would be changing the proptype that currently only accepts strings.

My use case is adding an "Optional" subtitle next to the label when i pass a boolean "optional" property

PR #1531, opened 2018-10-30
feat(Input): Allow label to be a React element

MHase:
PR created to solve issue #1424 
I've added possibility to pass object as propTypes to Input instead of passing only simple string.
Also `labelStyle` will work only if `label` is a simple string.
`labelProps` are also passed to component provided in `label` prop


docs/input.md
```diff
@@ -151,13 +151,13 @@ add a label on top of the input (optional)
...
 | :----: | :-----: |
-| string |  none   |
+| string **OR** React element or component |  none   |
 
 ---
 
 ### `labelStyle`
 
-styling for the label (optional)
+styling for the label (optional);  You only use this if `label` is a string
```
iRoachie:
```suggestion
styling for the label (optional);  You can only use this if `label` is a string
```

MHase: Completely forgot about test! Thank you 🥇 and sorry for that

iRoachie:
Thank you again! 💯

Issue #872, opened 2018-01-31
Type prop for SocialIcon


Oize:
Is it possible to set icon type for SocialIcon?
You are currently using FontAwesome, but there are other fonts in react-native-vector-icons.
I would like to use Zocial, so now i have to kinda recreate your SocialIcon from scratch.

xavier-villelegier:
Hey @Oize, this is currently not possible, but this is an interesting feature to have though, I'll let this open as an enhancement.
We have other priorities for the `v1.0.0` betas for now (React Native Web support, theming…) but feel free to submit a PR 👍

Oize:
Ok, i was just wondering if it is acceptable because of different network icon names in different fonts and stuff like this.

xavier-villelegier:
Yea this can be a bit tricky to implement properly. Let's see if someone takes the lead on this, otherwise if it becomes inactive I'll close it.

Oize:
After diving a little into this i ended up with an Icon component with my own network-color map.
I also tried parsing type with regexps like that:
`if (/stack-?overflow/.test(type)) return '#f27f33';`
Works well, but imo we're losing the simplicity.

LouisJS:
I don't understand the point of using SocialIcon when you can implement the same thing with the Icon Component such as : 

`<Icon
  name='sc-telegram'
  type='evilicon'
  color='#517fa4'
/>`

Am i missing something ?

LouisJS:
You can still do this by using the `Icon` Component
Such as `<Icon raised reverse name='instagram' type='font-awesome' color='#517fa4' />`

which supports `type (defaults to material, options are material-community, zocial, font-awesome, octicon, ionicon, foundation, evilicon, simple-line-icon, feather or entypo`

So @Oize, you can use this for now
@xavier-villelegier You could implement SocialIcon type the same way you did for `Icon`, isn't it ?

Oize:
@LouisJS yes, i can, just wanted to use more specific component with such an expressive name if it exists.
But currently `SocialIcon` looks way too specific to be usable.

PR #1534, opened 2018-10-31
feat(Socialicon): Add iconsSet prop

MHase:
Issue: #872 
Added new propTypes `iconsSet` to let user choose which they would like to use.
Default is font-awesome, but for normal social 'names' like facebook or github user can also choose between `zocial` and `entypo`.

```
<SocialIcon
  title='Sign In With Facebook'
  button
  type='facebook'
/>
<SocialIcon
  title='Sign In With Facebook'
  button
  iconsSet="zocial"
  type='facebook'
/>
<SocialIcon
  title='Sign In With Facebook'
  button
  iconsSet="entypo"
  type='facebook'
/>
```

img_4150

img_4149


iRoachie:
Man you are awesome, on fire with these PRs. I think we can take this chance to change some on the props to make it similar to the Icon.

I think we should make a breaking change here and change:
`type` to `name`
 `iconSet` to `type`

@xavier-villelegier @Monte9 @MHase what do you guys think? (just discussion no need to change anything yet)

xavier-villelegier:
@iRoachie I'm totally up for it, this needs to be standardized across all icon components !

Shout out to @MHase, you really help !

iRoachie:
*krrsh* Greeen light @MHase, i repeat, we've got the green light *krrsh*

MHase:
_krrsh_ roger that! 🙆‍♂️  _krrsh_

docs/social_icons.md
```diff
@@ -70,6 +70,7 @@ import { SocialIcon } from 'react-native-elements'
 * [`iconColor`](#iconcolor)
 * [`iconSize`](#iconsize)
 * [`iconStyle`](#iconstyle)
+* [`iconsSet`](#iconsset)
```
iRoachie: Think you missed the doc :D

MHase: Sorry for that! 😅

src/social/SocialIcon.js
```diff
@@ -69,11 +69,14 @@ const SocialIcon = props => {
...
     style,
     title,
-    type,
+    name,
     underlayColor,
+    type,
     ...attributes
   } = props;
 
+  let Icon = getIconType(type || 'font-awesome');
```
iRoachie: Use of using `type || 'font-awesome'`  add a default prop at the bottom setting `type` to 'font-awesome'

src/social/SocialIcon.js
```diff
@@ -141,7 +144,7 @@ const SocialIcon = props => {
 
 SocialIcon.propTypes = {
   Component: PropTypes.func,
-  type: PropTypes.string,
+  name: PropTypes.string,
```
iRoachie: may also be a good idea here to have like,

PropTypes.oneOf([ 'facebook']) etc. Since we actually don't accept "any string" the user passes

MHase: should I copy and paste it from the docs?
```
social media type (angellist, codepen, envelope, etsy, facebook,
foursquare, github-alt, github, gitlab, instagram, linkedin, medium,
pinterest, quora, reddit-alien, soundcloud, stack-overflow,
steam, stumbleupon, tumblr, twitch, twitter,
google-plus-official, vimeo, wordpress, youtube)
```

src/social/SocialIcon.js
MHase: I copied this line of code from `Icon` component, so if you want me to change it in `SocialIcon` maybe we should do it there too?

iRoachie: Yea maybe, but not in this PR

src/social/SocialIcon.js
iRoachie: Yea, will need to do it for typescript as well

src/index.d.ts
```diff
@@ -1729,6 +1729,11 @@ export interface SocialIconProps {
    */
   button?: boolean;
 
+  /**
+   * Social media name
+   */
+  name: SocialMediaType;
```
iRoachie: This isn't correct, this is supposed to be like `facebook` `google` etc

MHase: I just exported strings into variable called [`SocialMediaType`](https://github.com/react-native-training/react-native-elements/blob/8fedf7906c7d235844917cb2d101836d04e46794/src/index.d.ts#L1684) just like [`IconType`](https://github.com/react-native-training/react-native-elements/blob/8fedf7906c7d235844917cb2d101836d04e46794/src/index.d.ts#L27) is defined.

should I do it inline?

iRoachie: Sorry about that! I just realized you have two separate ones SocialMediaType and SocialIconType. It looked like the same when i took glance. 😄

src/social/SocialIcon.js
```diff
@@ -156,6 +186,7 @@ SocialIcon.propTypes = {
   activityIndicatorStyle: ViewPropTypes.style,
   small: PropTypes.string,
   iconSize: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  type: PropTypes.string,
```
iRoachie: Lets do the same thing you did with `name` to `type`

flyingcircle:
Unfortunately, this PR will have to be reworked. Some of props edited in this PR are used with different names. It'll be best to take the work here as an example and redo the changes in a new commit and PR.

PR #1535, opened 2018-10-31
fix: Overlay needs keys for children

### LINKED ISSUE #1517: Themed Overlay needs keys for children
**Explain what you did**
I'm using `ThemeProvider` with Overlays

**Expected behaviour**
Adding multiple children without keys should be possible

**Describe the bug**
If you add children to an themed overlay, they get added as an array and therefore need a key

**To Reproduce**
I don't know how to display warnings in snacks, but the warnings should get thrown here:
https://snack.expo.io/@lucidity/dGhlbW

| software | version |
| -- | -- |
| react-native-elements | 1.0.0-beta7 |
| react-native | 0.57.3 |
| npm or yarn | 5.6.0 |

haruelrovix:
_Problem:_
Each child in an array or iterator should have a unique "key" prop, [references](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318).

_Solution:_
Generate unique key using [shortid](https://www.npmjs.com/package/shortid) for Overlay children if it's an array.

|Before|After|
|-|-|
|||

Fixes #1517 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,8 @@
     "lodash.merge": "^4.6.1",
     "lodash.times": "^4.3.2",
     "opencollective-postinstall": "^2.0.0",
-    "prop-types": "^15.5.8"
+    "prop-types": "^15.5.8",
+    "shortid": "^2.2.13"
   },
   "devDependencies": {
     "@types/react": "^16.4.16",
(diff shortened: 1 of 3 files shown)


Monte9:
@iRoachie I noticed this too. When I used themes... all the children started complaining about missing keys.

I think seems like a bigger theming issue that needs to be fixed at it's core.

iRoachie:
Better fixed by https://github.com/react-native-training/react-native-elements/pull/1648/commits/d389d652e8b92cea56fecedb8cabb485f2281dcd

Issue #1551, opened 2018-11-04
How to add background image in header?


jazzystring1:
I wonder if it's possible to add background image in header? By reading the documentation (https://react-native-training.github.io/react-native-elements/docs/header.html), I can only change the background color but what I want to happen is to remove the background color and replace it with background image. Thanks

PR #1556, opened 2018-11-05
feat(Header): Add background image

MHase:
Issue: #1551 
Added possibility to pass `backgroundImage` prop which is simply `source` prop like in `Image` component :) e.g. :
```
source={{ uri: 'http://google.com' }}
```
img_4187

### DIFF
--- a/src/header/Header.js
+++ b/src/header/Header.js
@@ -45,6 +45,7 @@ const Header = ({
   rightContainerStyle,
   backgroundColor,
   backgroundImage,
+  backgroundImageStyle,
   containerStyle,
   placement,
   barStyle,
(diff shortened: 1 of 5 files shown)


iRoachie:
Brooooo this PR is 🔥

src/header/Header.js
```diff
@@ -110,6 +112,7 @@ Header.propTypes = {
   rightContainerStyle: ViewPropTypes.style,
   backgroundColor: PropTypes.string,
   backgroundImage: PropTypes.object,
+  backgroundImageStyle: PropTypes.style,
```
iRoachie: There's no PropTypes.style.

MHase: my bad, I was too quick. do you think simple `PropTypes.object` should be enough?

iRoachie: I think ViewPropTypes.style could work or even `Image.propTypes.style`. 

The latter would be more accurate but I am not sure if it works.

PR #824, opened 2018-01-16
ref issue #814 - Add click action on the image of the card component

giveli:
ref #814

Issue #814, opened 2018-01-11
Clickable image on the cards


giveli:
Hi, I was trying to add click action on the image of a card component. 

Here is my code :

```
<View style={styles.container}>
  <FlatList
      data={ this.state.interests }
      extraData={ this.state.showDetails }
      renderItem={
          ({item}) => {              
              return (
                  <Card
                      image={{ uri: item.picture }}
                      imageStyle= {styles.image}
                      backgroundColor='white'
                  >
                      {this.renderInterests(item)}
                  </Card> 
              );
          }
      }
      keyExtractor={(item, index) => index}
  />
 </View>
```

It seems from the documentation there isn't specific property on the card component. 
I definitely want to use the card's property image because I want to keep the header image style.

How I can do that ?  I need your help please !

iRoachie:
You want to be able to click on the card?

iRoachie:
Ah so @giveli we actually don't have this supported now although it sounds like a really great feature to have. Currently right now if you want to do this, you'll have to use an empty Card and handle the contents(image) as children yourself.

However if anyone wants to make this feature happen, they'll need to place a touchable around the image here  
https://github.com/react-native-training/react-native-elements/blob/c851d9ace8109db4e7b9b96b347ed4e53ed184a1/src/card/Card.js#L78-L105

giveli:
Thank you so much for your help ! 
I locally add a touchable around the image on Card.js ! It works ! Thank  you ! 
I will add a pull request as soon as possible.

iRoachie:
Great! I'll leave the issue open until it's implemented

k3ithl1m:
I would like to help out with this

ishigamii:
This feature has been added? will be added ? :D

iRoachie:
Not as yet @ishigamii, Feel free to try implementing it and sending a PR!

ishigamii:
@iRoachie I thought #814 was the PR for it my bad ;) I will maybe take a look at it though I don't know if I have the skills to create it but I can try :3

iRoachie:
It was https://github.com/react-native-training/react-native-elements/pull/824, but the user never finished.

lbelavoir:
@iRoachie is this still an issue ? I could take time to work on it.

iRoachie:
Yea it’s still open.

sudeepto-dutta:
@iRoachie I would like to work on this issue.  But I need some clarifications before I start with the implementation.

[...]

PR #1560, opened 2018-11-06
feat(Card): Add onPress props for image

MHase:
My approach for issue #814 
I saw that there was an attempt (#824), but unfortunately wasn't finished :(

I also created conditionalWrap helper to wrap component only if condition is met.


iRoachie:
I'm planning to refractor this component for v1 to make it more useful compared to Tile. So i'm not gonna merge this as it wouldn't make sense.

Issue #1247, opened 2018-06-08
SearchBar clear button doesn't appear when changing props.value


niklasgrahl:
**Explain what you did**

I'm using the search bar as a controlled component (supplying props.value) but changing value from outside the component doesn't show/hide the clear button.
I installed the next version because I wanted the fix for the onClear prop, but I'm still experiencing issues.

**Expected behaviour**
I expect the clear button to show up when setting props.value to anything but an empty string.

**Describe the bug**
The clear button isn't shown/hidden when props.value changes.

**To Reproduce**
I couldn't recreate it on snack because I couldn't install next version on snack, but here's a simple example:

```javascript
export default class App extends Component {
  constructor(props){
    super(props);
    this.state = {
      value: ""
    }
  }
  render() {
    return (
      <View>
        <SearchBar value={this.state.value} />
        <Button title="abc" onPress={() => this.setState({value: 'abc'})} />
      </View>
    );
  }
}
```

```bash
# Snack
Unable to install next version on snack

# Steps
1. Type in the search bar (clear button appears)
2. Click clear button
3. Click 'abc' button.
4. Search bar text changes but clear button doesn't appear.
```

**Possible Solution**
Recalculate this.state.isEmpty in src/searchbar/SearchBar-{ios,android,default}.js when props.value changes

**Your Environment (please complete the following information):**

| software | version |
| -- | -- |
| react-native-elements | https://github.com/react-native-training/react-native-elements.git#next (commit a521fc53ead7c33d1d2ac08fcaae36ec51df4df4) |
| react-native | 0.54.2 |
| npm or yarn | yarn |

xavier-villelegier:
@niklasgrahl You need to control the value using the `onChangeText` props, as you would on a React Native `TextInput`.

```jsx
<SearchBar
  value={this.state.value}
  onChangeText={value => this.setState({ value })}
/>
```

niklasgrahl:
Yes, exactly. Sorry, in an effort to simplify the example I accidentally removed `onChangeText`, but of course that should be there. Thanks for the quick reply!

franck-nadeau:
I needed this for my app, so I changed the code and created a PR for it. Waiting on a code review. 

https://github.com/react-native-training/react-native-elements/pull/1575

PR #1575, opened 2018-11-14
Check to make sure that the SearchBar is really empty

franck-nadeau:
Changed the code to take into account when a value has been set on the props of the SearchBar.

Fix for this issue https://github.com/react-native-training/react-native-elements/issues/1247


iRoachie:
Thank you!

PR #1591, opened 2018-11-20
Fix avatarprops to align with the docs.

janhesters:
The avatar had no props for its placeholder within its TypeScript declaration file. Now it has 😊


iRoachie:
Good catch but someone else already added it in https://github.com/react-native-training/react-native-elements/pull/1494

PR #1602, opened 2018-11-26
[Avatar] added android as default style

piu130:


iRoachie:
Shouldn't this be the opposite? Isn't it that react-native-web _doesn't_ support the elevation style?

iRoachie:
I think the latter might be best. Especially for cases such as TouchableOpacity over TouchableNativeFeedback. Let's use iOS as the default and Android will have it's own styles

iRoachie:
Thanks again 🎉

PR #1604, opened 2018-11-26
HOC to give any component a badge

janhesters:
This HOC can give any component a badge. It works with minimal syntax:
```
withBadge(<SomeComponent />)(1)
```
while still allowing the user to customize anything he / she wants. The argument of the first functoin is . the component that should get the badge. The `1` at the end is the value within the badge. If you look at the code you'll see the rest of the API (like x - y offset of badge, programatically hide/show badge etc.).

By default it hides the badge automatically for empty string / number = 0.


xavier-villelegier:
Hey @janhesters, as a lover of HOC, I really like the idea ! 💯 
Could you just update from the `next` branch ?

Here are my thoughts:

### 1) Your decorator should be used in this order:
```jsx
const DecoratedComponent = withBadge(2)(BaseComponent)
```
It's what most decorators do.

### 2) It *must* propagate the props
If you do this:
```jsx
<DecoratedComponent myProps={anything} />
```

You expect your `BaseComponent` to receive `myProps`.

### 3) It should be able to take a callback
In a lot of cases, you're just going to take the badge value from the props. Let's imagine an example where you have your notifications in your store and you use `connect`, you must be able to use it like that:

```jsx
@connect(state => ({
  notifications: state.notifications,
}))
@withBadge(props => props.notifications.length)
export default class MyDecoratedIcon extends React.Component {
  render() {
    return (
      <Icon type="ionicon" name="md-cart" />
    );
  }
}
```

--------
### Proposal
Taking this into account, here is my proposal:

```js
export const withBadge = (value) => (WrappedComponent) => {
  return class extends React.Component {
    render() {
      const badgeValue = typeof value === 'function' ? value(this.props) : value
      return (
        <React.Fragment>
          <WrappedComponent {...this.props} />
          {// Add your badge here with all the styling with `value={badgeValue}`}
        </React.Fragment>
      )
    }
  } 
}
```

janhesters:
> Could you just update from the next branch ?

What exactly do you mean with this (I'm not that familiar with GitHub, took me forever to find out how to do PRs 😄)

xavier-villelegier:
Haha it's okay !
From your local fork, just run `git pull upstream next`, and then push the changes !

janhesters:
@xavier-villelegier I get the message: "fatal: 'upstream' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists."

xavier-villelegier:
@janhesters I updated your repo, I think you didn't have the `upstream` remote configured to the official RNE repo. Here is the doc about how to sync a fork with the original repo, might be useful in the future 😉 https://help.github.com/articles/syncing-a-fork/

janhesters:
@xavier-villelegier BTW I love your changes and implemented them and added them to the PR (also I converted it from TS to JS).

The only small problem I have with these changes is that in my previous version I could do this:
 static navigationOptions = {
    headerTitle: "List",
    headerRight: withBadge(<Icon />)(1)
  };
Now, since this
```
static navigationOptions = {
  headerTitle: "List",
  headerRight: <withBadge(1)(Icon) />
};
```
is not possible, we would have to do something like this:
```
static navigationOptions = () => {
  const BadgedIcon = withBadge(1)(Icon);
  return {
    headerTitle: "List",
    headerRight: <BadgedIcon />
  };
};
```

janhesters:
@xavier-villelegier Thank you for the help with git!

xavier-villelegier:
What about

```
static navigationOptions = {
  headerTitle: "List",
  headerRight: withBadge(1)(<Icon />)
};
```

?

janhesters:
After your improvements
```
static navigationOptions = {
  headerTitle: "List",
  headerRight: withBadge(1)(<Icon />)
};
```
throws:
```
Functions are not valid as a React child. This may happen if you return a Component instead of <Component /> from render. Or maybe you meant to call this function rather than return it.
```
because in my original code I wrote `{Component}` instead of `<Component />` in the HOC's code.

xavier-villelegier:
@janhesters Oops I meant:

```jsx
static navigationOptions = {
  headerTitle: "List",
  headerRight: withBadge(1)(Icon)
};
```

iRoachie:
@xavier-villelegier with that way, how would you customize props for the Icon

janhesters:
@xavier-villelegier Yup, what @iRoachie said.
```
static navigationOptions = () => {
  const BadgedIcon = withBadge(1)(Icon);
  return {
    headerTitle: "List",
    headerRight: <BadgedIcon />
  };
};
```

Is the only way.

But to be honest, I'm already using the HOC improved by your Feedback in production, an assigning badges via:
```
@connect(state => ({
  notifications: state.notifications,
}))
@withBadge(props => props.notifications.length)
export default class MyDecoratedIcon extends React.Component {
  render() {
    return (
      <Icon type="ionicon" name="md-cart" />
    );
  }
}
```
Is super dope (as long as you only have on source of truth for you badges; otherwise you'll need several of these components for each enriched Icon). And this way you don't have to write much code in your `headerLeft` or `headerRight`.

janhesters:
@iRoachie @xavier-villelegier I worked on this a bit more, because as I said we are using this HOC right now in production, and I have a very elaborate TypeScript implementation of it. Let me know if you want to see it, or if you want to add this HOC as a `.tsx` instead of `.js` in the first place 😊

iRoachie:
@janhesters I think we'd like to move to TS sometime in the future but that may be a bit difficult right now so let's stick with js.

src/badge/WithBadge.js
```diff
@@ -0,0 +1,58 @@
...
+      const badgeValue = typeof value === "function" ? value(this.props) : value;
+      return (
+        <View>
+          <WrappedComponent {...this.props} />
+          {!hidden && (
+            <Badge
+              containerStyle={styles.badge}
+              textStyle={styles.badgeText}
+              value={badgeValue}
+              wrapperStyle={[styles.badgeContainer, { bottom: offSetY, left: offSetX }]}
```
iRoachie: Also you should probably look over the Badge. We've made some changes to badge on next as apart of the 1.0.0 release. So you might want to check those changes first. 

Here's the PR https://github.com/react-native-training/react-native-elements/pull/1545

_PS_ there's no more `wrapperStyle`

janhesters: Thank you, I edited the commit. If you are okay with the code and the API, I would like to also add the docs etc. to this PR 😊

src/index.d.ts
```diff
@@ -632,6 +632,15 @@ export interface ButtonGroupProps {
...
 export class ButtonGroup extends React.Component<ButtonGroupProps> {}
 
+export function withBadge(
+  value: ReactNode | (() => ReactNode),
+  status: 'primary' | 'success' | 'warning' | 'error',
+  offSetX?: number,
+  offSetY?: number,
+  badgeProps?: BadgeProps,
+  hidden?: boolean;
+): <P extends object>(WrappedComponent: ComponentClass<P>) => ComponentClass<P>;
```
iRoachie: Does this mean that this can't be used with a function component?

Should probably use `React.ComponentType` instead of `ComponentClass`

janhesters: Updated.

janhesters:
Question to both of you concerning the API.

Right now it is like thisL you can either manually define `value`, color (via `status`), `offSetX` and `offSetY`, which are probably the most commonly used, or overwrite everything with `badgeProps`. In the latter case you still have to provide values for the first three though, since otherwise you can't reach `badgeProps`.

Is this okay? If not, should we either leave out full customisability (= no `badgeProps` for this HOC), or overload the function to either take in these most common values, or `badgeProps`?

xavier-villelegier:
I think we should have only two args, the value, and the options. With options containing everything as an object:

```
withBadge(value, { offsetX, offsetY, badgeProps, … })(Component)
```

Doing that you can use any params you want without having to specify others just to reach the one you want

janhesters:
@xavier-villelegier And then ignore `value` if it's also in `badgeProps`?

iRoachie:
You can destructure the incoming props and get the needed for the positioning.
```jsx
const withBadge = (
   value,
   {
      offsetX,
      offsetY,
      ...rest,
   }
 )
```

janhesters:
Either I'm too dumb to implement it, or the suggested change with using an object doesn't work 😣

Here is what I tried:

```js
const withBadge = (
  value,
  options = {
    status: "error",
    offSetX: 10,
    offSetY: 22,
    hidden:
      typeof value === "number"
        ? value === 0
        : typeof value === "string"
        ? value !== ""
        : value !== undefined && value !== null,
    badgeProps: {}
  }
) => WrappedComponent =>
  class extends React.Component {
    render() {
      const { status, offSetX, offSetY, hidden, badgeProps } = options;
      const badgeValue = typeof value === "function" ? value(this.props) : value;
      return (
        <React.Fragment>
          <WrappedComponent {...this.props} />
          {!hidden && (
            <Badge
              badgeStyle={styles.badge}
              textStyle={styles.badgeText}
              // textStyle={styles.badgeText}
              value={badgeValue}
              status={status}
              containerStyle={[styles.badgeContainer, { bottom: offSetY, left: offSetX }]}
              {...badgeProps}
            />
          )}
        </React.Fragment>
      );
    }
  };
```
And this way, as soon as there is one parameter for options (e.g. just hidden), everything collapses and all other parameters are overwritten.

xavier-villelegier:
@janhesters Yes because this syntax:

```jsx
const withBadge = (
  value,
  options = {
    status: "error",
    offSetX: 10,
    offSetY: 22,
    hidden:
      typeof value === "number"
        ? value === 0
        : typeof value === "string"
        ? value !== ""
        : value !== undefined && value !== null,
    badgeProps: {}
  }
)
```

Set a default value to `options`. If you provide anything to `options` its value is overwritten, it doesn't merge what you provided and the default one.

Instead of providing a default `options` object, you can provide a default value to each property:

```jsx
const withBadge = (
  value,
  options = {}
) => WrappedComponent =>
  class extends React.Component {
    render() {
      const {
        status = 'error',
        offSetX = 10,
        offSetY = 22,
        hidden,
        badgeProps
      } = options;
      const badgeValue = typeof value === "function" ? value(this.props) : value;
      …
```

By the way, instead of exposing offSetX and offSetY, I think we should expose top, left, bottom, and right. Because with just bottom and left, if you want to put your badge on the right you have to know the dimensions of your Icon instead of just using `right: 0` 😅

janhesters:
@xavier-villelegier Thank you for your help, I'm learning so much through this PR. Updated the component, declaration file and docs :)

janhesters:
@iRoachie @xavier-villelegier Anything else missing from this? Or is this now good to go? 😊

docs/badge.md
```diff
@@ -142,3 +159,63 @@ Custom component to replace the badge outer component
...
+### `value`
+Text value to be displayed by badge, defaults to empty
+
+|                    Type                    | Default |
+| :----------------------------------------: | :-----: |
+| String OR Number OR React Native Component |  none   |
+
+---
+
+### `options`
```
xavier-villelegier: Specify that it is an object with the following values

docs/badge.md
```diff
@@ -57,6 +64,9 @@ import { Avatar, Badge } from 'react-native-elements'
     containerStyle={{ position: 'absolute', top: -4, right: -4 }}
   />
 </View>
+
+const BadgedIcon = withBadge(1)(Icon)
+<BadgedIcon type="ionicon" name="ios-chatbubbles" />
 ```
```
xavier-villelegier: Could you just add the example with the `connect` thing we talked above ? I think it's a pretty common use case

src/badge/withBadge.js
```diff
@@ -0,0 +1,63 @@
...
+        status = "error",
+        left = 10,
+        bottom = 15,
+        top = 0,
+        right = 0,
+        hidden = typeof value === "number"
+          ? value === 0
+          : typeof value === "string"
+          ? value !== ""
+          : value !== undefined && value !== null,
```
xavier-villelegier: I think you can just set `hidden = !value`, as empty string and 0 will be falsy values.

src/badge/withBadge.js
```diff
@@ -0,0 +1,63 @@
...
+
+const withBadge = (value, options = {}) => WrappedComponent =>
+  class extends React.Component {
+    render() {
+      const {
+        status = "error",
+        left = 10,
+        bottom = 15,
+        top = 0,
+        right = 0,
```
xavier-villelegier: If you expect left 10 and bottom 15 to put the badge on the top right hand corner, you should use top and right to not worry about the wrapped component dimension

janhesters:
@xavier-villelegier I implemented the requested changes 😊

xavier-villelegier:
Could you just update your code to fit the lint rules ?
https://travis-ci.org/react-native-training/react-native-elements/builds/462786647#L492

It's just about single quotes, unused import, and display name

janhesters:
@xavier-villelegier Done!

janhesters:
@xavier-villelegier Thank you for the kind words and thank you for guiding me! This was insightful and fun as f*ck 😄

docs/badge.md
```diff
@@ -57,6 +64,22 @@ import { Avatar, Badge } from 'react-native-elements'
     containerStyle={{ position: 'absolute', top: -4, right: -4 }}
   />
 </View>
+
+const BadgedIcon = withBadge(1)(Icon)
+<BadgedIcon type="ionicon" name="ios-chatbubbles" />
+
+// Using the decorator proposal
+@connect(state => ({
```
iRoachie: Not sure we should add this here, I assure you we'll get issues asking why my code doesn't work.

Stick to standard aAPIs

src/badge/withBadge.js
```diff
@@ -0,0 +1,66 @@
...
+    fontSize: 10,
+    paddingHorizontal: 0
+  }
+});
+
+const withBadge = (value, options = {}) => WrappedComponent => {
+  class WithBadge extends React.Component {
+    render() {
+      const {
+        status = 'error',
```
iRoachie: Don't duplicate props from `Badge`, badge already has a `status`, `onPress` and `value` prop. If anything, pull it from badgeProps

iRoachie: Maybe an easier design for the props would be to let it accept all the badgeProps:
```js
left,
bottom,
top,
right,
hidden,
...badgeProps
```

docs/badge.md
xavier-villelegier: Actually I suggested @janhesters to add this as it is a pretty common use case for a decorator

iRoachie: Decorators aren't apart of the spec, this is advanced usage. Someone that's already knowledgeable of decorators would already know that they can use the hoc in that way, so this would only confuse beginners(newer users).

src/badge/withBadge.js
janhesters: You mean like this:

```jsx
const {
        left = 0,
        bottom = 0,
        top = -5,
        right = 8,
        hidden = !value,
        badgeProps = {
          status = "error",
          onPress
        },
      } = options;
      const badgeValue = typeof value === "function" ? value(this.props) : value;
      return (
        <View>
          <WrappedComponent {...this.props} />
          {!hidden && (
            <Badge
              badgeStyle={styles.badge}
              textStyle={styles.badgeText}
              value={badgeValue}
              status={status}
              containerStyle={[styles.badgeContainer, { bottom, left, right, top }]}
              onPress={onPress}
              {...badgeProps}
            />
          )}
        </View>
      );
```

iRoachie: As in they'll have no use for `badgeProps` option, any properties outside left, bottom, top, right, hidden, would be considered badgeProps

docs/badge.md
xavier-villelegier: So if you don't show how to use it in the `Usage` section, where do you explain to the user how to use this decorator ?

docs/badge.md
```diff
@@ -142,3 +172,72 @@ Custom component to replace the badge outer component
...
+---
+
+## Reference for withBadge
+
+### `value`
+Text value to be displayed by badge, defaults to empty
+
+|                    Type                    | Default |
+| :----------------------------------------: | :-----: |
+| String OR Number OR React Native Component |  none   |
```
xavier-villelegier: Oops I just forgot this one: `value` can also be a callback that takes the props as parameter

janhesters: Added it.

src/badge/withBadge.js
janhesters: The problem is how would you define default values then if you use the rest operator?

```js
const { left = 0, bottom = 0, top = -5, right = 8, hidden = !value, ...badgeProps = {status: "error", onPress} } = options;
```
Isn't possible.

janhesters: Okay I got it working implementing right now...

janhesters: Done.

janhesters:
Is anything missing here, or is this ok now?

docs/badge.md
janhesters: Any news on this?

docs/badge.md
```diff
@@ -19,18 +19,25 @@ indicate the status of an item to the user.
   <img src="/react-native-elements/img/badge/badge--indicator.jpg" alt="Badge as Indicator" />
     <figcaption>Badge as Indicator</figcaption>
   </figure>
+  <figure>
```
iRoachie: Let's drop this image from here

docs/badge.md
```diff
@@ -57,6 +64,22 @@ import { Avatar, Badge } from 'react-native-elements'
     containerStyle={{ position: 'absolute', top: -4, right: -4 }}
   />
 </View>
+
```
iRoachie: Add a section comment here:

```
// withBadge hoc
```

docs/badge.md
```diff
@@ -73,6 +96,13 @@ import { Avatar, Badge } from 'react-native-elements'
 
 ---
 
+## Props for withBadge
```
iRoachie: Let's move this down with the reference for withBadge.

So it should read:
Props
Reference
Props for withBadge
Reference for withBadge

docs/badge.md
```diff
@@ -142,3 +172,52 @@ Custom component to replace the badge outer component
...
+| :----: | :-----: |
+| number |   15    |
+
+#### `top`
+
+|  Type  | Default |
+| :----: | :-----: |
+| number |   15    |
+
+#### `badgeProps` (Spreaded out)
```
iRoachie: Remove this instead use similar pattern we use in other components

e.g. https://github.com/react-native-training/react-native-elements/blame/next/docs/image.md#L38 just link it to badge props instead

src/badge/withBadge.js
```diff
@@ -0,0 +1,49 @@
...
+    fontSize: 10,
+    paddingHorizontal: 0
+  }
+});
+
+const withBadge = (value, options = {}) => WrappedComponent => {
+  class WithBadge extends React.Component {
+    render() {
+      const { left = 0, bottom = 0, top = -5, right = 0, hidden = !value, ...badgeProps } = options;
+      const badgeValue = typeof value === 'function' ? value(this.props) : value;
```
iRoachie: When would this be a function?

src/badge/withBadge.js
```diff
@@ -0,0 +1,49 @@
...
+              status="error"
+              containerStyle={[styles.badgeContainer, { bottom, left, right, top }]}
+              {...badgeProps}
+            />
+          )}
+        </View>
+      );
+    }
+  }
+  WithBadge.displayName = `WithSubscription(${WrappedComponent.displayName ||
```
iRoachie: WithSubscription is a little vague, why not WithBadge?

src/index.d.ts
```diff
@@ -388,6 +388,28 @@ export interface BadgeProps {
...
+   * Options to configure the badge
+   */
+  options: {,
+    bottom?: number,
+    left?: number,
+    right?: number,
+    top?: number,
+    hidden?: boolean,
+  } & BadgeProps,
+): <P extends object>(WrappedComponent: React.ComponentType<P>) => React.ComponentType<P>;
```
iRoachie: Use the `<P = {}> syntax here`

src/badge/withBadge.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import { StyleSheet, View } from 'react-native';
+import Badge from './Badge';
+
+const styles = StyleSheet.create({
```
iRoachie: Just convention, but move the styles after the component

janhesters:
@iRoachie Done! Thank you for your guidance and your response..

docs/badge.md
```diff
@@ -142,3 +163,58 @@ Custom component to replace the badge outer component
...
+ 
+> Also receives all
```
iRoachie: Move this up to be right under ###options

src/index.d.ts
```diff
@@ -388,6 +388,28 @@ export interface BadgeProps {
...
+   * Options to configure the badge
+   */
+  options: {,
+    bottom?: number,
+    left?: number,
+    right?: number,
+    top?: number,
+    hidden?: boolean,
+  } & BadgeProps,
+): <P extends {}>(WrappedComponent: React.ComponentType<P>) => React.ComponentType<P>;
```
iRoachie: Remove the `extends` just use `<P = {}>`

See https://github.com/react-native-training/react-native-elements/blob/next/src/index.d.ts#L2083 for example

src/badge/withBadge.js
```diff
@@ -0,0 +1,49 @@
...
+      );
+    }
+  }
+  WithBadge.displayName = `WithBadge(${WrappedComponent.displayName ||
+    WrappedComponent.name ||
+    'Component'})`;
+  return WithBadge;
+};
+
+export default withBadge;
```
iRoachie: Move this to the very bottom

janhesters: I'm sorry I got that wrong.

docs/badge.md
```diff
@@ -26,11 +26,14 @@ indicate the status of an item to the user.
...
 
+### withBadge Higher-Order Component
+The withBadge HOC allows you to easily add badges to icons and other components.
+
 ## Usage
 
 ```js
 import { Text, View } from 'react-native'
-import { Avatar, Badge } from 'react-native-elements'
+import { Avatar, Badge, Icon } from 'react-native-elements'
```
iRoachie: The import here for `withBadge`

janhesters: I'm sorry for all the mistakes that slipped by. Thank you for catching them!

iRoachie: It's no problem man! Don't feel bad about it. It's normal 👍

iRoachie:
Hey @janhesters I realize you've been making these changes through github.com instead of clone the repo locally.

Is there any reason you're doing this? The last thing that needs to be done is to write tests for this, but you'll have to clone your forked repo locally

iRoachie:
I also added a commit to run prettier on your code. Would you like some help on the tests part?

janhesters:
The reason is that I always got confused when trying to fork things. Branches wouldn't show up, or I couldn't `git push`. So I found that editing these files in the browser more reliable. Probably a noob move 😄 But it worked for me.

janhesters:
@iRoachie  Yes I would! If you guide me, I would be happy to learn!

iRoachie:
Okay getting started you want to get it set up locally.

```bash
git clone https://github.com/janhesters/react-native-elements.git
cd react-native-elements
git checkout patch-1
npm install
```

Let me know if you get through this far

janhesters:
@iRoachie Done! (Gonna go on a trian ride now though, so my next response might take a well, because there is no wifi on the train).

src/badge/withBadge.js
xavier-villelegier: When you want to get the badge value from the props.

```jsx
withBadge(props => props.notifications.length)…
```

Still the same use case with the `connect` thing in the docs

iRoachie: Gotcha! I don't use decorators, so i had no idea! Thanks though

iRoachie:
Alright man safe travel. Your next step is to run 

```bash
npm test
```

It should give you a table layout in the terminal, look for the line with `withBadge` and you'll see the lines that are uncovered by tests on the right

janhesters:
@iRoachie Shall I just generate SnapShot tests or should I unit tests with Jest and Enzyme?

iRoachie:
Go with snapshots first and then depending on which lines are left uncovered then we can add enzyme tests

janhesters:
@iRoachie Did the snapshot tests. See if you like them :)

iRoachie:
Test are green, great job!

src/index.d.ts
```diff
@@ -388,6 +388,28 @@ export interface BadgeProps {
...
+ */
+export function withBadge(
+  /**
+   * Text value to be displayed by badge
+   */
+  value: React.ReactNode | (() => React.ReactNode),
+  /**
+   * Options to configure the badge
+   */
+  options: {,
```
iRoachie: Lastly this should be optional, so `options?: `

janhesters: Updated!

src/badge/withBadge.js
```diff
@@ -0,0 +1,64 @@
...
+
+const withBadge = (value, options = {}) => WrappedComponent => {
+  class WithBadge extends React.Component {
+    render() {
+      const {
+        left = 0,
+        bottom = 0,
+        top = -5,
+        right = 0,
+        hidden = !value,
```
iRoachie: Also we need to change this logic here.

The badge also has a mini badge mode that shows when no value is given to the Badge. Currently, right now this will always be hidden if the user is trying to use the mini badge.

```jsx
// Badge will never show since `hidden = !value` will be true
const StatusAvatar = withBadge(undefined)(Avatar);
```

Think we may need to just set hidden to false by default instead of checking for the value. @xavier-villelegier

janhesters: Sounds logical.

iRoachie: Yea let's do that

xavier-villelegier: Sounds good to me, go for it !

iRoachie:
```suggestion
        hidden = false,
```

xavier-villelegier:
image

This is how you spot a huge PR lol, well done @janhesters and thanks again for your motivation to make this PR clean ✨!

janhesters:
Thank you guys for guiding me! Had tons of fun and learned a lot 🔥. I might take a look at the milestones next if I have time and see if I can help with 1.0 release.

src/badge/withBadge.js
```diff
@@ -0,0 +1,64 @@
...
+        right = 0,
+        hidden = false,
+        ...badgeProps
+      } = options;
+
+      const badgeValue =
+        typeof value === 'function' ? value(this.props) : value;
+
+      return (
+        <View>
```
iRoachie: I think we may need to add some styling here. There's a weird display bug.

Generated by: 
```jsx
const BadgedAvatar = withBadge(1)(Avatar);

  <BadgedAvatar
    size="large"
    source={{ uri: 'https://randomuser.me/api/portraits/men/44.jpg' }}
  />
</View>

  container: {
    flex: 1,
    padding: 30,
    paddingTop: Constants.statusBarHeight,
    // alignItems: 'center',
    justifyContent: 'center',
    backgroundColor: 'hsl(116, 10%, 98%)',
  },
});
```
simulator screen shot - iphone x - 2018-12-07 at 11 26 12

Styling on the outer view maybe?
```suggestion
        <View style={{ position: 'relative', alignItems:'center', justifyContent: 'center' }}>
```

simulator screen shot - iphone x - 2018-12-07 at 11 29 38

janhesters:
I commited the changes, but I'm gonna test them now. Give me a couple of minutes.

src/badge/withBadge.js
```diff
@@ -0,0 +1,69 @@
...
+      const badgeValue =
+        typeof value === 'function' ? value(this.props) : value;
+
+      return (
+        <View style={styles.container}>
+          <WrappedComponent {...this.props} />
+
+          {!hidden && (
+            <Badge
+              badgeStyle={styles.badge}
```
iRoachie: Lastly lets remove these default styles (badgeStyle, textStyle). That way we can use the default values from the Badge itself.

Also it would allow for the mini badge to be rendered.
screen shot 2018-12-09 at 4 09 23 pm

```jsx
const BadgedAvatar = withBadge(null, {
  right: -30,
  status: 'success',
})(Avatar);

<BadgedAvatar size="medium" title="MD" rounded />
```

janhesters: Sounds good!

src/badge/withBadge.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import { StyleSheet, View } from 'react-native';
+
+import Badge from './Badge';
+
+const withBadge = (value, options = {}) => WrappedComponent => {
+  class WithBadge extends React.Component {
+    render() {
+      const {
+        left = -7,
```
iRoachie: What are these values based on?

image

```jsx
const BadgedAvatar = withBadge(1, {
    status: 'success',
  })(Avatar);

  const BadgedIcon = withBadge(1, {
    status: 'success',
  })(Icon);

<BadgedAvatar size="medium" title="MD" rounded />
<BadgedIcon name="chat" size={30} />
```

Any ideas for defaults?

iRoachie: It's to the left by default I see

janhesters: They are based on icons for headers and tabs in react navigation.

```js
export class SomeScreen extends Component {
  static navigationOptions = () => {
    const BadgedIcon = withBadge(1)(Icon);
    return {
      headerTitle: "Title",
      headerRight: (
        <BadgedIcon
          name={`${Platform.OS === "ios" ? "ios" : "md"}-chatbubbles`}
          type="ionicon"
          color="white"
          containerStyle={styles.padRight}
        />
      )
    };
  };
```

But, since this HOC lives here, it might be better to base them on components for RNE 😄 So should we optimize the defaults for `<Avatar />`s in `<ListItem />`s?

```js
renderItem = ({ item }) => {
  const BadgedAvatar = withBadge(1)(Avatar);
  return (
    <ListItem
      title={item.name}
      leftAvatar={<BadgedAvatar source={{ uri: item.avatar }} rounded={true} />}
      bottomDivider={true}
    />
  );
};
```

iRoachie: We can decide now. The badge defaults are pretty big, so i think i'll change the sizes of the original badge. I took some ideas from the twitter app.

screen shot 2018-12-09 at 5 46 25 pm

Will make those changes then let you know

janhesters: Okay. I took the sizes (that I corrected earlier today), from iOS (WhatsApp and Telegram).

iRoachie: Alright updated! It looks pretty good.

Should we assume the badge will probably go to the top right?

iRoachie: screenshot_1544398184

Whatcha think? This is with:
```js
const {
  left = 45,
  bottom = null,
  top = 4,
  right = null,
  hidden = false,
  ...badgeProps
} = options;
```

cc @janhesters @xavier-villelegier

janhesters: I think it should be a little more to the right, and a little more down. Here is an example from WhatsApp to compare:

whatsappexample

Also in the right most tab for status we have a reference for the mini badge. Furthermore if the value of the badge is in the double digits, as in the second listitem in this screen shot, it should have flat top and bottom borders (I also tested this behavior in WhatsApp with the badge from the tabs by having a friend spamming me 😄). Btw this look is consistent in Telegram, too.

janhesters: Okay, I tweaked it to resemble the WhatsApp look:

withbadgeexample

Here is the change that I did for that:

```js
const { bottom, hidden = false, left, ...badgeProps } = options;a
let { right = -16, top = -1 } = options;
if (!value) {
  (right = -3), (top = 3);
}
```

@iRoachie Either way, when default the badge to the top right, we should use top and right as reference values instead of left (or bottom).

**On another note:** Notice how in your Twitter the big badge example and in my WhatsApp example the mini badge, has a background around itself. Maybe we should add a boolean to the `<Badge />` that does this? 🤔 Otherwise people would have to do this styling manually.

iRoachie: The mini badge doesn't have any additional styling other than size https://github.com/react-native-training/react-native-elements/blob/next/src/badge/Badge.js#L72. Seems to be a bug with border radius and background color, but it only happens on iOS.

```jsx
if (!value) {
  (right = -3), (top = 3);
}
```

^ I think this might a bit too much logic.

> @iRoachie Either way, when default the badge to the top right, we should use top and right as reference values instead of left (or bottom).


The right I did the value from the left is so that when the badge value it increases that it doesn't move and cover the icon. e.g `99+` would start to cover the icon. If the positioning is set from the left then it never covers.

> Furthermore if the value of the badge is in the double digits, as in the second listitem in this screen shot, it should have flat top and bottom borders 

The badge does this already!

janhesters: >The badge does this already!

Just noticed this, too (awesome).

>^ I think this might a bit too much logic.

Sometimes extra logic is good. I mean without this line, the MiniBadge would always be off, resulting in each user having to add their own position. I think the extra logic is worth taking of this labour step :)

BTW, I see another problem, that both our suggestions have:

A very common thing for (icon-/ text) button's in the header is to add padding or margin so that they don't stick to the side of the screen.

E.g. using React Navigation and RNE:

```js
  public static navigationOptions = () => {
    const BadgedIcon = withBadge()(Icon);
    return {
      headerLeft: <Button title="Edit" type="clear" containerStyle={styles.padLeft} />,
      headerRight: (
        <BadgedIcon name="edit" type="feather" containerStyle={styles.padRight} color="#007aff" />
      ),
      title: 'Chats'
    };
  };
```

Now the problem with this is that with the current way this HOC is written it will expand the container `<View />` of the icon and place the `<Badge />` related to this container. This results in this:

headerbug

Any idea for a solution to tackle all these little details?

janhesters: Okay actually the last thing might not be that big of a problem, cause one could always do:

```js
  public static navigationOptions = () => {
    const BadgedIcon = withBadge()(Icon);
    return {
      headerLeft: <Button title="Edit" type="clear" containerStyle={styles.padLeft} />,
      headerRight: (
        <View style={styles.padRight}>
          <BadgedIcon name="edit" type="feather" color="#007aff" />
        </View>
      ),
      title: 'Chats'
    };
  };
```

Or we add a parameter to the HOC that let's the user define padding for the HOC's container?

iRoachie: > A very common thing for (icon-/ text) buttons within the header is to add padding or margin so that they don't stick to the side of the screen.

I understand the needed for it but i'm not sure why a dev would use padding here and not margin. Wouldn't margin be more fitting here?

> Or we add a parameter to the HOC that let's the user define styles for the HOC's container?
I agree with this also!

janhesters: Done. The only thing that we have to think about now are the default values.

iRoachie:
What did you do just now? I think you just added some other changes to files

janhesters:
Yeah, god damnit, somehow some files formatting changed (even though I didn't touch them 😦). I see if I can revert that somehow ...

Other than that I added the suggested containerStyle prop.

iRoachie:
Nevermind i see those are from the precommit hook.

iRoachie:
Yea what happen is that we added the precommit linting and formatting after that code was added :D so by you running it just now it updated.

janhesters:
So they are okay?

iRoachie:
It's kinda strange though, it looks like it's doing different kind of formatting than it's supposed to :| lol. Don't worry about it

janhesters:
So @iRoachie what defaults should we decide to use? And should we implement an if statement to consider the MiniBadge?

docs/badge.md
```diff
@@ -142,3 +163,63 @@ Custom component to replace the badge outer component
...
+
+### `options`
+
+> Also receives all
+> [Badge](https://facebook.github.io/react-native/docs/badge#props) props.
+
+Object with the following (optional) keys:
+
+#### `bottom`
+
```
iRoachie: Hey @janhesters we're just gonna go forward and merge it.

Can you just update the default values to match the code here?

janhesters: Done 👍

janhesters:
Thank you for teaching me so much!

I'm getting kinda hooked and passionate about RNE 😄 . It's the best UI library for RN imo! 🔥

I would love to help here and there (if I find the time). I've looked into the issues for the milestones, and tried to fix some of those without PR, but couldn't fix any. But I will keep an eye out and try to help improve this library going forward.

janhesters:
@xavier-villelegier Thank you guys, but I would like to politely decline. To me the rewards are the learnings 📈.

I would rather go back to what @iRoachie said and help out here and there with maintaining and enhancing RNE if I find the time and a proper challenge 🚀

xavier-villelegier:
Your call man ! Feel free to reach us on Slack if you change your mind. Remember that the Open Collective is **only** for contributors !

Glad you enjoyed so much your work on RNE though, we're always looking for motivated people for open sourcing 🙋‍♂️

PR #1620, opened 2018-11-27
fix(SearchBar): Switch from material-community to material

piu130:
This PR changes the `SearchBar` icons from `material-community` to `material`.

The default icon is `material`. We should use it here too, so RNW does not have to load two fonts just because the `SearchBar` uses another one than the default.

This increases the performance of the app and needs less network traffic, because only one font is needed by default.

Question: Do we need it and should we change it in SearchBar.android too?

### DIFF
--- a/src/searchbar/SearchBar-android.js
+++ b/src/searchbar/SearchBar-android.js
@@ -18,24 +18,24 @@ const SCREEN_WIDTH = Dimensions.get('window').width;
 const ANDROID_GRAY = 'rgba(0, 0, 0, 0.54)';
 
 const defaultSearchIcon = {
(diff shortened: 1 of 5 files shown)


iRoachie:
Yep let's do that! Just so you don't have to check I tested the values for the android searchbar.

```jsx
searchIcon={{ type: 'material', name: 'search' }}
clearIcon={{ type: 'material', name: 'clear' }}
cancelIcon={{ type: 'material', name: 'arrow-back' }}
```

piu130:
@iRoachie Done. Thanks for the names :)

iRoachie:
Thanks again @piu130 💯

PR #1624, opened 2018-11-28
Add official platform colors

janhesters:
This PR redefines the default colors to be the platform specific colors according to their respective design guidelines ([iOS](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/color/), [Android](https://material.io/design/color/the-color-system.html#tools-for-picking-colors)).


iRoachie:
Wooo not too sure about that one. I think people would prefer having the same colours across platforms when they build an app.

janhesters:
Well if they overwrite it in their themeprovider they will have the same colors on any platform. I just thought it would be nice to provide a feeling as native as possible for React Native Elements out of the box 😊

xavier-villelegier:
Mmh I agree that it might be better to have the same color across platforms. Sometimes you want a different design to be consistent with the platforms, which is legit, but imo colors are more related to the graphic charter of your application, which is the same on iOS and Android (I guess ?)

janhesters:
> (We could still export them for users if they want to use platform colors though)
Sounds good, too. Maybe expose them through platform variables? So additionally to
```js
primary: '#2089dc
```
provide
```js
platformPrimary: Platform.select({
  default: '#2196f3', // rgb(33, 150, 243) Android
  ios: '#007aff' // rgb(0, 122, 255)
}),
```
?

iRoachie:
Also, how'd you determine error and success colors?

janhesters:
@iRoachie on iOS it's simple, because there is only one green and one red in the color palette. For I found that the most commonly used values from the color palette are just 500 so I took their respective hex values.

Here are the pallettes again: [iOS](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/color/), [Android](https://material.io/design/color/the-color-system.html#tools-for-picking-colors).

iRoachie:
AH okay thanks for posting (slipped me the first time). Maybe a `platform` property?

```js
const theme: {
  platform: {
    colors: {
      ios: //
      android: //
    }
  }
}
```

thoughts?

janhesters:
Like
```js
const theme: {
  platform: {
    colors: {
      ios: {
        primary: '#007aff',
        error: `#...`,
        // ...
      },
      android: // same
    }
  }
}
platformPrimary: Platform.select({
  default: '#2196f3', // rgb(33, 150, 243) Android
  ios: '#007aff' // rgb(0, 122, 255)
}),
```
? I like it 😊

iRoachie:
Alright then! 🚢

janhesters:
@iRoachie Done

iRoachie:
Think you missed the `platform` key

janhesters:
Yup 👍

iRoachie:
Snapshots now?

janhesters:
If yes, I'm gonna do the docs.

janhesters:
Just did `npm install`. Wow there are a lot of outdated packages 😂 Also vulnerabilities:
```
(1 low, 14 moderate, 1 high)
```
Is this intentional?

Anyway, I made the snapshot 😊

iRoachie:
Ha well, I think it's that we depend on packages, but the ones that have vulnerabilities are dependencies of other packages.

docs/customization.md
```diff
@@ -167,7 +167,9 @@ style as well as the red color set in the theme.
 ### The Theme Object
 
 By default, the theme object looks like this. You can add whatever values you
-want to the theme, and they will be merged with the default.
+want to the theme, and they will be merged with the default. By default the 
+platform colors aren't used anywhere. These native colors are added for
```
iRoachie: This is nice! Good copy here 👍

docs/customization.md
```diff
@@ -258,3 +274,24 @@ const MyComponent = () => (
...
+import { colors } from 'react-native-elements';
+
+const theme = {
+  colors: {
+    Platform.select({
+      default: ...colors.platform.android,
+      ios: ...colors.platform.ios
+    })
+  }
+}
```
iRoachie: Maybe right under this, add a simple example with Text element color

janhesters: Sure about that? I mean at this point the user has probably understood that all `primary` variables are overwritten.

Or do you mean an example showing how to apply the platform colors without spreading them like in this example?

janhesters: I added a simple example.

iRoachie: Not like that, but using the color directly. 
```jsx
<Text style={{ color: theme.colors.platform.ios.sucess}}>Hey</Text>
```

janhesters: Done.

iRoachie: @janhesters Sorry about that, now that i've read the entire document it looks a bit weird. Let's put back the example you had first

janhesters: Lol, no worries, okay 👍

janhesters:
@iRoachie Gladly! Thank you for guiding me 😊

PR #1628, opened 2018-11-30
Accessibility Support :

akshaypunchh:
added 'accessibilityLabelSelectedSuffix', 'accessibilityLabelUnSelectedSuffix'  as a props. It will allow to added customised accessibilityLabel for ADA support.  

e.g. I have a Button Groups of two tabs 'Tab1', 'Tab2', and Tab1 is selected 
it will speak for Tab1 ==> Tab1  selected (Will add Suffix for selected tab)
It will speak for Tab2 ==> Tab2 double tap to activate (Will add Suffix for unselected tab)

```
                                        <ButtonGroup
						accessibilityLabelSelectedSuffix={'button selected'}
						accessibilityLabelUnSelectedSuffix={'button double tap to activate'}
						buttons={buttons}
					/>
```

### DIFF
--- a/src/buttons/ButtonGroup.js
+++ b/src/buttons/ButtonGroup.js
@@ -37,6 +37,8 @@ const ButtonGroup = props => {
     setOpacityTo,
     containerBorderRadius,
     disableSelected,
+    accessibilityLabelSelectedSuffix,
+    accessibilityLabelUnSelectedSuffix,
     ...attributes
   } = rest;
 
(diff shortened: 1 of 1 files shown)


iRoachie:
Hey still trying to grasp this change. Can you show and example usage and what would the screen reader speak?

akshaypunchh:
@iRoachie 
image
Just check the tab of whatsapp 
there are three tabs CHATS, STATUS, CALLS
and Selected tab is CHATS.
if Accessibility(talk back) is enable from phone's settings then for default condition it will talk while focus on tabs... It will talk back only tab's text (CHATS, STATUS or CALLS). We can't determine that which tab is selected or which is not selected, So I am sending different suffix for selected of unselected case ( 'accessibilityLabelSelectedSuffix', 'accessibilityLabelUnSelectedSuffix').

let accessibilityLabelSelectedSuffix= " Tab is selected"
accessibilityLabelUnSelectedSuffix= " Tab is unselected"
Now it will talkback CHATS Tab is selected , STATUS Tab is unselected , so on

Ref: https://facebook.github.io/react-native/docs/accessibility

flyingcircle:
This is not the correct way to go about solving this issue. Button state should be controlled by the `AccessibilityState` prop, not a unique 'suffix' prop. I'll write up a formal issue ticket for a full screen reader compliance review. It is indeed one of the weak points of RNE

Issue #1244, opened 2018-06-06
Sliders issue - wrong rendering values


rdewolff:
**Explain what you did**
Trying to use 2 sliders on one screen. With the second minimal value being bound to the first one's value.

**Expected behaviour**
It should work as the `react-native` slider without any problem.

**Describe the bug**
The second slider's value gets wrongly placed and cannot be used anymore.

**To Reproduce**

```bash
# Snack
https://snack.expo.io/rkJEqYBe7

# Steps
1. Go to the snack provided here : https://snack.expo.io/rkJEqYBe7
2. Tap to play :)
3. Move the "Slider 1" value to more than 50.
4. You'll see the 2nd slider's wrongly displayed.

**Possible Solution**
If you change the import `import { Slider } from 'react-native-elements';` to `import { Slider } from 'react-native';`, the behaviour will be correct.

There seems to be a mistake in the render method, not checking the values provided are within the `minimumValue`.

**Screenshots**
<img width="322" alt="screen shot 2018-06-06 at 18 30 15" src="https://user-images.githubusercontent.com/130704/41051903-b45ffe14-69b7-11e8-88c9-c5d601be5029.png">

**Your Environment:**

| software | version |
| -- | -- |
| react-native-elements |1.0.0-beta5 |
| react-native |0.55.1|
| npm or yarn |yarn 1.6.0|

xavier-villelegier:
Hey @rdewolff, thanks for reporting that !

If anyone wants to tackle this, you need to make sure the `value` is between min and max here
https://github.com/react-native-training/react-native-elements/blob/e0406c72f341b354f2861b85eb9a432863a50768/src/slider/Slider.js#L70-L72

PR #1651, opened 2018-12-07
Bound slider value between min and max

xavier-villelegier:
Refs #1244

### DIFF
--- a/src/slider/Slider.js
+++ b/src/slider/Slider.js
@@ -19,6 +19,13 @@ var DEFAULT_ANIMATION_CONFIGS = {
   },
 };
 
+const getBoundedValue = ({ value, maximumValue, minimumValue }) =>
+  value > maximumValue
+    ? maximumValue
+    : value < minimumValue
+    ? minimumValue
+    : value;
+
 class Rect {
   constructor(x, y, width, height) {
     this.x = x;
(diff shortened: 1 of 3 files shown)


iRoachie:
Finished here?

xavier-villelegier:
Yep

iRoachie:
simulator screen shot - iphone x - 2018-12-07 at 10 18 56

Getting this on iOS, seems the value is null at some point?

xavier-villelegier:
@iRoachie My bad I copy pasted too fast what I did on Snack to fix it. Fixed 👍

iRoachie:
Is this how it's to work? Maybe if the new incoming value is outside the max, min then we should do nothing and not update?

kapture 2018-12-07 at 10 28 20

xavier-villelegier:
The use case here is different, you updates the bounds (`minimumValue` and `maximumValue`), not the value.
This would mean we don't update new max and min from the props if value is not between these two, which is a bit weird to my mind. It means that the user has to take care itself that the value is between the bounds, otherwise he won't be able do update the bounds once the value isn't between them anymore

iRoachie:
Ahh you're saying the example from the issue, isn't really a real-world case and just contrived to show that the slider can go out of bounds?

xavier-villelegier:
Everything can be a real world use case haha but yea with this PR the slider cannot go out of bounds by changing dynamically the bounds anymore

iRoachie:
🚢

PR #1668, opened 2018-12-14
[Icon] Add reverseRatio prop

martinezguillaume:
Hi ! 😊
This is just a little useful feature to control the reverse ratio of an Icon.

### Example

| reverseRatio | Preview |
|--------------|---------|
| 1            |capture d ecran 2018-12-14 a 11 52 48|
| 1.5 (default)          |capture d ecran 2018-12-14 a 11 52 59|
| 2            |capture d ecran 2018-12-14 a 11 53 10|

You get the idea ! 😘

### DIFF
--- a/src/icons/Icon.js
+++ b/src/icons/Icon.js
@@ -48,9 +48,9 @@ const Icon = props => {
         style={StyleSheet.flatten([
           (reverse || raised) && styles.button,
           (reverse || raised) && {
-            borderRadius: (size * reverseRatio) / 2,
-            height: size * reverseRatio,
-            width: size * reverseRatio,
+            borderRadius: (size * reverseRatio + 4) / 2,
+            height: size * reverseRatio + 4,
+            width: size * reverseRatio + 4,
           },
           raised && styles.raised,
           {
@@ -103,7 +103,7 @@ Icon.defaultProps = {
   raised: false,
   size: 24,
   color: 'black',
-  reverseRatio: 1.5,
+  reverseRatio: 2,
   reverseColor: 'white',
   disabled: false,
   type: 'material',
--- a/src/icons/Icon.js
+++ b/src/icons/Icon.js
@@ -22,6 +22,7 @@ const Icon = props => {
     reverse,
     raised,
     containerStyle,
+    reverseRatio,
     reverseColor,
     disabled,
     disabledStyle,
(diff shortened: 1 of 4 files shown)


iRoachie:
What would be the use cause of something like this? 

Can we give it a better name that reverseRatio ? It's not immediately clear what it does

martinezguillaume:
It works with the `reverse` prop

src/icons/Icon.js
```diff
@@ -47,9 +48,9 @@ const Icon = props => {
         style={StyleSheet.flatten([
           (reverse || raised) && styles.button,
           (reverse || raised) && {
-            borderRadius: size + 4,
-            height: size * 2 + 4,
-            width: size * 2 + 4,
+            borderRadius: (size * reverseRatio) / 2,
+            height: size * reverseRatio,
+            width: size * reverseRatio,
```
iRoachie: This would be a breaking change no?

Say example size = 30

Before: 
```
height: 30 * 2 + 4 = 64
```

After:
```
height: 30 * 1.5 = 45
```

This would result in different sized icons if you were using icon previously

martinezguillaume: We can define `reverseRatio` to 2 by default, to not change anything.
I removed the `+ 4`, I don't see why it is there

iRoachie: Not sure why it's there either, but it means we'd have to update all the components that use Icon (which is a lot). As well as in user's apps. Your call

martinezguillaume: We can see that later
For now, let's leave `+ 4` and reverseRatio to `2` by default to not change anything

iRoachie:
Snapshots!

martinezguillaume:
To discuss about the prop name, it's directly related to the `reverse` prop. To control the colored container size compared to the icon size. Maybe we can have a more generic name like `sizeRatio` or `containerSizeRatio` but we lose the link with `reverse` prop. For me, the most acceptable name is `reverseRatio` and I find it pretty explicit

iRoachie:
Just add to the typescript definitions and we’re ready to go

martinezguillaume:
Hi agaiiiiin, if i'm fixing conflicts, are we ready ?

flyingcircle:
This is possible to do by altering the `containerStyle` sizes. Let me know if this is still a desired prop

PR #1671, opened 2018-12-16
Add React Native Ratings dependency

Monte9:
This PR imports `Rating` from `react-native-ratings` package and re-exports it through the index file.  

## Context: 
I created the [react-native-ratings](https://github.com/Monte9/react-native-ratings) and wanted it to be available through RNE. Hence at the time, I copied the source over into RNE. But this has led to a disconnect between the two packages and hence we see a lot of issues related to something that might have already been fixed in the other package.

Hence as a work around, I am removing the source code for ratings from RNE and will just be maintaining it over at `react-native-ratings`. This way, all issues can be directed to that repo and closed here since they are non-actionable through RNE.


src/index.js
```diff
@@ -4,6 +4,7 @@
 // https://material.angularjs.org/latest/demo/button
 
 // Core
+import { AirbnbRating, Rating } from 'react-native-ratings';
```
Monte9: this is not core, but prettier moves this to the top automatically.

iRoachie: Add this on the first line.

`/* eslint-disable import/order */`

iRoachie:
Are there any docs that need to be updated?

Monte9:
Ah good catch. Will update docs too.

iRoachie:
Alright, only thing left would be typescript definitions. Instead of adding them here I'll PR over on your repo.

iRoachie:
PR is here https://github.com/Monte9/react-native-ratings/pull/36. So the process is:
1. Merge the typescript PR
2. Release a new version of react-native-ratings
3. Add new version here in package.json
4. Delete local types and use the ones from your library directly (i can do this)

Monte9:
@iRoachie I have published a new version of `react-native-ratings` v6.3 and have updated this PR to use that version. fyi.

iRoachie:
Alright @Monte9, off to you now to add the docs part for AirbnbRating and that should be it

iRoachie:
Just tried it with the ThemeProvider and everything works.

```jsx
const theme = {
  Rating: {
    type: 'bell',
  },
  AirbnbRating: {
    count: 3,
  },
};

export default () => (
  <ThemeProvider theme={theme}>
      <AirbnbRating />
      <Rating />
    </View>
  </ThemeProvider>
);
```

screen shot 2019-01-05 at 7 32 22 pm

Monte9:
Awesome! Great work! 💯 

I'll follow up with the docs tomorrow and we can merge it in.

PR #1676, opened 2018-12-18
Fix type of Input's inputComponent.

etanzapinsky:
This was React.ComponentClass, but should be React.ComponentType given
how it's used.


iRoachie:
This one actually can't be ComponentType, ComponentType includes React.SFC. In our case it can only be ComponentClass because the Input component uses refs and refs can only be used in class components.

etanzapinsky:
Ah, got it. Thanks for the quick response.

Issue #1232, opened 2018-05-26
[Input] Direct Manipulation


mcwissink:
**Is your feature request related to a problem? Please Describe.**
There seems to be no way to call setNativeProps on the TextInput wrapped by the Input element.
It would be nice to be able to manipulate the actual text displayed by Input

**Describe the solution you'd like**
A setNativeProps prop on Input that forwards function onto the TextInput child.

**Describe alternatives you've considered**
Using the react-native TextInput and styling it similar to the other Input elements on the page, a poor solution.

**Additional context**
This would be useful for [clearing](https://facebook.github.io/react-native/docs/direct-manipulation.html) Input, or in my specific case, using a formatter for phone number input

iRoachie:
Hey @mcwissink while you can call actually call the `clear` method on the input it'll clear the text value. Here's an example https://snack.expo.io/rkFRK2Pk7.

I do however understand that some persons may want to access the TextInput directly, so we could add this as a method on the Input

iRoachie:
Don't think we are going forward this with this at this time.

PR #1682, opened 2018-12-21
[Input] Direct Manipulation

marcelolx:
Based on #1232  added setNativeProps(nativeProps) method to Input.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -887,6 +887,13 @@ export class Input extends React.Component<InputProps, any> {
    * eg `this.inputRef.clear()`
    */
   clear(): void;
+
+  /**
+   * Calls setNativeProps on the Input
+   * 
+   * eg `this.inputRef.setNativeProps({ text: 'any text' })`
+   */
+  setNativeProps(nativeProps: Object): void;
 }
 
 export interface HeaderIcon extends IconObject {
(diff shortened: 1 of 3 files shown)


src/index.d.ts
```diff
@@ -887,6 +887,13 @@ export class Input extends React.Component<InputProps, any> {
    * eg `this.inputRef.clear()`
    */
   clear(): void;
+
+  /**
+   * Calls setNativeProps on the Input
+   * 
+   * eg `this.inputRef.setNativeProps({ text: 'any text' })`
+   */
+  setNativeProps(nativeProps: Object): void;
```
iRoachie: We could probably type this better.

```suggestion
  setNativeProps(nativeProps: Partial<TextInputProperties>): void;
```

marcelolx: `setNativeProps (nativeProps: Partial <TextInputProps>): void;` not a better option? Because [TextInputProperties](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/bcedcaf254ac774013661e14b64d811f503076dd/types/react-native/legacy-properties.d.ts#L98) are deprecated.

iRoachie: Sure go ahead, they’re technically the same thing. One aliases the other

iRoachie:
Thanks! 💯

PR #1686, opened 2018-12-22
feat(Tile): Allow Custom ImageComponent

BradenM:
Added ImageComponent prop to both Tile and FeaturedTile to customize the default `BackgroundImage` component used.

Followed naming-style of #1261 to keep it consistent.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1937,6 +1937,12 @@ export interface TileProps {
    * @default Device Width * 0.8
    */
   height?: number;
+
+  /**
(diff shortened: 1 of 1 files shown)


iRoachie:
Awesome work! Thanks for sending this in. Do you mind adding to the typescript definition as well? 

Tile - https://github.com/react-native-training/react-native-elements/blob/next/src/index.d.ts#L1859

BradenM:
Sure! Thanks.

PR #1695, opened 2018-12-28
fix(types): Fix ListItem checkBox props typo

josephktcheung:
`ListItem`'s `checkBox` props in `index.d.ts` is misspelled. It should be [`checkBox`](https://github.com/react-native-training/react-native-elements/blob/master/src/list/ListItem.js#L332) not [`checkbox`](https://github.com/react-native-training/react-native-elements/blob/master/src/index.d.ts#L1049).


iRoachie:
Thanks! Good catch 💯

Issue #1281, opened 2018-06-25
Bad sizing for vertical sliders


matejcik:
**Explain what you did**

I am trying to build an application that has a slider on the left-hand side of the screen, and some descriptions in the rest of the space. I expect to do this through nesting two Views like this:

```jsx
<View style={{ flex: 1, flexDirection: "row", alignItems: "stretch" }}>
	<Slider orientation="vertical" />
	<View>
		<Text>hello world</Text>
		<Button title='Click me' onPress={function() {}} />
	</View>
</View>
```

**Expected behaviour**

If on the first view I leave the default flexDirection ("column"), and the Slider is horizontal, it takes up full width of the screen:
```
------O----
Hello world
[Click me]
```
I expect that when the slider is vertical, it will take up the full *height* of the screen:
```
|   Hello world
|
|   [Click me]
|
O
|
```

**Describe the bug**
The slider isn't visible at all.
When setting distinct background colors to the elements, the Slider element is completely missing.

**To Reproduce**
https://snack.expo.io/By8f7vRbQ

**Possible Solution**
It seems that sizing for the slider is messed up in several kinds of ways. For example, using a horizontal slider but not specifying its width (e.g., with flex layout and `alignItems: "center"`) will make the slider disappear completely.
(the native Slider will take up a minimum width)

AFAICT sizing works the same way for horizontal and vertical, so a vertical slider will reserve the (in)appropriate height, but zero width.
I tried to dive into the code and figure out how to change the sizing, but apparently the rotation is triggered by a CSS transform and I have no idea whatsoever how that works in regard to sizing.

xavier-villelegier:
@matejcik Thanks for reporting that, I agree that the slider styling is a bit messy and needs to be improved. Feel free to give it a shot if you have time.

YeonV:
The problem is, while being "vertical"-orientated, the slider still takes the same space in width.
So you can only make it work in a ugly-hacky way:
styles:
` 
 container: {
    flexDirection: "row",
    alignItems: "center",
    justifyContent: "space-between"
  },
wrapper: {
    alignItems: "center",
    flexDirection: "row",
    marginLeft: 10,
    width: 50,
    height: 180
  },
  slider: {
    flex: 1,
    flexBasis: 150,
    marginLeft: -52
  },`
Code:
`<View style={styles.container}>
       <View style={styles.wrapper}>
            <Slider
                orientation="vertical"
                minimumValue={0}
                maximumValue={127}
[...]

PR #1698, opened 2018-12-30
Slider flexbox layout

armenbadalyan:
Slider component is relying on` transform: [{ rotate: '90deg' }]` style to implement vertical orientation. This seems to cause issues similar to the ones described in #1281. This PR refactors the Slider component to use flexbox layout thus making Slider size and position more predictable in both vertical and horizontal orientations.


iRoachie:
Can you resolve the conflicts?

armenbadalyan:
@Monte9 , thank you for taking the time to review my PR.  In horizontal orientation, since the width in your example was not specified explicitly, according to react-native layout rules it takes 100% of the parent container width. In vertical orientation one should specify the width either explicitly or using flexbox layout. In other words it behaves just like a regular View.
Here's an example of vertically oriented slider:

```
<Slider
        orientation="vertical"
        minimumValue={0}
        maximumValue={100}
        minimumTrackTintColor="blue"
        maximumTrackTintColor="black"
        trackStyle={{ height: 7, borderRadius: 5 }}
        thumbStyle={{
          backgroundColor: "green",
          width: 30,
          height: 30,
          borderRadius: 15
        }}
        value={this.state.value}
        onValueChange={value => this.setState({ value })}
        style={{ height: 200 }}
        step={10}
      />

```

Monte9:
Ah that makes sense. Thanks for the detailed explanation.
screen shot 2019-01-13 at 10 55 19 am

Looks good on my end. Lets just make sure we add a note on the docs mentioning that when in "vertical" orientation, height is also required.

Monte9:
Also appreciate you submitting the PR! We use OpenCollective to give back to contributors like yourself so I'd be happy to send $20 your way for your efforts. Let me know if this sounds good to you and I can tell you how to submit your expense.

iRoachie:
Are we waiting on the change to the docs ?

src/slider/Slider.js
```diff
@@ -547,21 +595,37 @@ Slider.defaultProps = {
...
+    flex: 1,
+    width: TRACK_SIZE,
+  },
   thumb: {
     position: 'absolute',
     width: THUMB_SIZE,
     height: THUMB_SIZE,
     borderRadius: THUMB_SIZE / 2,
+  },
+  thumbHoizontal: {
```
markj-: Small typo here. Great work with the PR. I've just stumbled upon this issue myself.

iRoachie:
Thanks again @armenbadalyan 💯

armenbadalyan:
Thank you guys, it was my pleasure :)

PR #1706, opened 2019-01-07
fix(SearchBar iOS): Allow margins to be added to containerStyle

iRoachie:
Same fixes as #1701 but for iOS SearchBar

### DIFF
--- a/src/searchbar/SearchBar-ios.js
+++ b/src/searchbar/SearchBar-ios.js
@@ -33,11 +33,11 @@ class SearchBar extends Component {
   constructor(props) {
     super(props);
     const { value } = props;
+
     this.state = {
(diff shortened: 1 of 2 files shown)


src/searchbar/SearchBar-ios.js
```diff
@@ -163,7 +163,13 @@ class SearchBar extends Component {
         />
 
         <View
-          style={{ marginLeft: this.state.cancelButtonTransform }}
+          style={StyleSheet.flatten([
+            styles.cancelButtonContainer,
+            {
+              opacity: this.state.cancelButtonWidth === null ? 0 : 1,
```
iRoachie: To prevent flicker on android before onLayout is called

Issue #1710, opened 2019-01-08
Tooltip coordinate system


cidevant:
**Explain what you did**
I am trying to show tooltips inside ScrollView(horizontal)

**Expected behaviour**
I expect to see tooltip in visible area

**Describe the bug**
If element has width 1000 (for example) - tooltip is rendered over center of the element (~500px) and nobody can see it. Tooltip coordinate system should reflect visible area and render tooltip over visible area of element.

**To Reproduce**
Horizontal ScrollView with elements wider than visible area.

**Possible Solution**
Rewrite coordinate system and render methods:
https://github.com/react-native-training/react-native-elements/blob/master/src/tooltip/getTooltipCoordinate.js
https://github.com/react-native-training/react-native-elements/blob/master/src/tooltip/Tooltip.js

**Screenshots**
![capture](https://danielprsala.cz/xxx.png)

**Your Environment (please complete the following information):**

| software | version |
| -- | -- |
| react-native-elements | ^1.0.0-beta7 |
| react-native | 0.57.3 |
| npm or yarn | yarn 1.12.3 |

Monte9:
Off topic: that Tab Bar looks 😍 Can you share what package you used or how you did it?

cidevant:
@Monte9 It is `react-navigation` + `react-navigation-tabs`. Using `createBottomTabNavigator` with this config `tabBarComponent: MyCustomBottomTabBar`. Component `MyCustomBottomTabBar` wraps native component `import { BottomTabBar } from 'react-navigation-tabs';`. Write me a mail, if you want source code.

cidevant:
I solved it by properly calculating visible area of wrapped element.
```javascript
export const getElementVisibleWidth = (elementWidth, xOffset, ScreenWidth) => {
  // Element is fully visible OR scrolled right
  if (xOffset >= 0) {
    return (xOffset + elementWidth) <= ScreenWidth // is element fully visible?
      ? elementWidth // element is fully visible;
      : (ScreenWidth - xOffset) // calculate visible width of scrolled element
  } 
  // Element is scrolled LEFT
  else {
    return elementWidth - xOffset; // calculate visible width of scrolled element
  }
};
```
And then it should be used here:
```javascript
//react-native-elements/src/tooltip/getTooltipCoordinate.js

const center = [x + getElementVisibleWidth(width, x, ScreenWidth) / 2, y + height / 2];
```
and here (`elementVisibleWidth`)
```javascript
[...]

PR #1711, opened 2019-01-09
fix(tooltip): proper calculation of tooltip position based on visible element width

cidevant:
This merge request fixes tooltip position calculations. Now we use visible width of wrapped element for tooltip placement. Fixes [#1710](https://github.com/react-native-training/react-native-elements/issues/1710)

### DIFF
--- a/src/tooltip/Tooltip.js
+++ b/src/tooltip/Tooltip.js
@@ -6,7 +6,7 @@ import { ViewPropTypes, withTheme } from '../config/';
 import { ScreenWidth, ScreenHeight, isIOS } from '../helpers';
 
 import Triangle from './Triangle';
-import getTooltipCoordinate from './getTooltipCoordinate';
+import getTooltipCoordinate, { getElementVisibleWidth } from './getTooltipCoordinate';
 
 class Tooltip extends React.PureComponent {
   state = {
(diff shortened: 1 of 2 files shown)


Monte9:
@cidevant thanks for submitting the PR! 💯 

I don't fully understand the math that is going on but if it fixes it for you, I am ok with approving it. 

Also can you please share a screenshot of what it looks like before & after the fix. This will help us visualize the changes.

cidevant:
Here you can see calculations
capture

Before
capture

After
capture

iRoachie:
Does this affect any behavior when the element is not in a scrollview? Also can you change the target branch to `next`

Monte9:
@cidevant this is great. Thanks for the detailed explanation.

Appreciate you submitting the PR! We use OpenCollective to give back to contributors like yourself so I'd be happy to send $20 your way for your efforts. Let me know if this sounds good to you and I can tell you how to submit your expense.

cidevant:
@Monte9 I've registered on [OpenCollective](https://opencollective.com/cidevant) ;-)

xavier-villelegier:
Thanks @cidevant, I just approved your expense on OpenCollective. Your app looks dope !! 🔥 Is it supposed to be released on the stores one day ? Or is it a personal project ?

cidevant:
@xavier-villelegier Thank you for approval ;-) App is still under heavy development, but you can see it now on Google play store: Bassoe Offshore Rig database. We are facing some performance issues, so we working on them now.

PR #1731, opened 2019-01-28
Update 2019-01-27-1.0-release.md

YikSanChan:
Add `value` prop in `SearchBar`


iRoachie:
We didn't add a `value` prop. It's always been there. You may want to look at controlled inputs vs uncontrolled https://reactjs.org/docs/forms.html#controlled-components

Issue #1738, opened 2019-01-30
After update from 1.0.0.beta7 to 1.0.0, minimum slider track does not have the right height 


msevestre:
**Explain what you did**
I have updated an application to use v1.0.0. I notice that among the issue I already mentioned in #1737, there as also another issue with the slider component

When changing the height of the slider component, I am expecting the height of the minimumTrack to also change. This was happening in v1.0.0.beta7 but is not the case anymore. Instead, the minimumTrack has the constant height of 4 it seems 

**Expected behaviour**
I am expecting the minimumTrack to have the same height as the track

**Describe the bug**
BEFORE UPDATE
![image](https://user-images.githubusercontent.com/1041237/52006778-f8d84f80-249a-11e9-9e54-0e2e1318b615.png)

AFTER UPDATE
![image](https://user-images.githubusercontent.com/1041237/52006811-0beb1f80-249b-11e9-969c-358bf4842a3b.png)

If I make the track even bigger, the effect is quite noticeable
![image](https://user-images.githubusercontent.com/1041237/52006865-26bd9400-249b-11e9-82fe-aaba3f83221d.png)

**To Reproduce**
Put the slider into a component and change set the track like so
```
<Slider
              value={this.state.sliderValue}
              onValueChange={this.handleSliderValueChanged}
              trackStyle={styles.track}
              minimumTrackTintColor={color}
/>


with 
styles track: {
    height: 16,
    borderRadius: 10,
  },
```

I'll create a snack and add it here when I have time. I just wanted to report the issue before I forget

rawad663:
Hey @msevestre, I've been able to reproduce your problem. I noticed that the thumb's positioning is set to a fixed value. I created a solution that takes the height/width provided into account.

I have a PR ready if anyone is willing to go over it GH-1739

PR #1739, opened 2019-01-31
Fixed false positioning of Slider Thumb

rawad663:
For Issue GH-1738
- changed style of mainStyles.thumbHorizontal, mainStylesthumbVertical to take height/width into consideration (respectively)
- passed props.trackStyle width/height if provided by user

### DIFF
--- a/src/slider/Slider.js
+++ b/src/slider/Slider.js
@@ -425,6 +425,7 @@ class Slider extends Component {
             orientation === 'vertical'
               ? mainStyles.trackVertical
               : mainStyles.trackHorizontal,
+              trackStyle,
             minimumTrackStyle,
           ])}
         />
--- a/src/slider/Slider.js
+++ b/src/slider/Slider.js
@@ -435,8 +435,8 @@ class Slider extends Component {
             { backgroundColor: thumbTintColor },
             mainStyles.thumb,
             orientation === 'vertical'
-              ? mainStyles.thumbVertical
-              : mainStyles.thumbHorizontal,
+            ? mainStyles.thumbVertical(trackStyle && trackStyle.width)
+            : mainStyles.thumbHorizontal(trackStyle && trackStyle.height),
             thumbStyle,
             {
               transform: [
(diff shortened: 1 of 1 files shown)


Monte9:
Thanks for the fix @rawad663 💯 

That math is always hard to understand just by looking at it. Can you please attach screenshots of what it looks like with your PR. Also, can you try the `orientation="vertical"` and make sure that looks fine too.

rawad663:
Hey @Monte9, I've tested the solution with `orientation="vertical"` and it's worked well with the different width values. Here are the screenshots of both orientations after the fix

screenshot_20190131-124549
screenshot_20190131-124617
screenshot_20190131-124627
screenshot_20190131-124649
screenshot_20190131-124708

msevestre:
@rawad663 What's the dark gray background? Shouldn't it be hidden?
I am going to try it in my project as well to let you know how it looks

rawad663:
@msevestre I don't see how the gray background can be hidden. In your code in GH-1738, you've set the height of the track to 10px and the gray background is shown

msevestre:
@msevestre Yeah that's the bug in my opinion. i want the red part to be wider in your example.
Is there a way to make the red section wider then? As you can see in my example, the gray section was not visible BEFORE the update and it is now. I just changed the track height to something big to show the bug in a clearer way.

rawad663:
I see! you want the colored portion of the track to cover the gray portion too. I definitely add that

rawad663:
The fix is pushed, I added trackStyle to the colored portion
```
<Animated.View
          style={StyleSheet.flatten([
            mainStyles.track,
            orientation === 'vertical'
              ? mainStyles.trackVertical
              : mainStyles.trackHorizontal,
              trackStyle,
            minimumTrackStyle,
          ])}
```

slider-fix

msevestre:
@rawad663 looks great. Going to update my screenshot with the code

msevestre:
YES!! You fixed two bugs: the Thumb not centered and the background! Well done @rawad663 
image

iRoachie:
Thanks for the combined effort here. Amazing 💯

PR #1741, opened 2019-02-01
ESLint error in Text.js Component

rawad663:
- added propType fontFamily in src/text/Text.js
- added defaultProp fontFamily in src/text/Text.js

### DIFF
--- a/src/text/Text.js
+++ b/src/text/Text.js
@@ -6,7 +6,7 @@ import { fonts, withTheme } from '../config';
 import normalize from '../helpers/normalizeText';
 
 const TextElement = props => {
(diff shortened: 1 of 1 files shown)


rawad663:
After fix:
image

iRoachie:
Hey @rawad663 thanks for spotting this. However font family should be here as a prop. The styles for the text component are passsed i’m through the `style` prop. Instead this PR should remove font family from the props spread on line 
https://github.com/react-native-training/react-native-elements/blob/master/src/text/Text.js#L9

rawad663:
Hello @iRoachie, I haven't realized that fontFamily isn't even used in the file!
I will update the PR removing the references of it.

iRoachie:
Thanks! 💯

Issue #1744, opened 2019-02-01
Tooltip positioning issue on android


kufii:
I'm using version 1.0.0-beta7

When using the Tooltip on iOS it works fine, but on Android, it's pushed down, and the icon the tooltip is on is duplicated. It's worth noting that the tooltip is within a component within a react-navigation stack

https://snack.expo.io/@adpyke/courageous-kiwi

![image](https://user-images.githubusercontent.com/4913552/52144631-e42bc100-262c-11e9-87dc-ffc6ccd51c8c.png)

michaelclark2:
I'm having the same issue, even on a blank app with no react-navigation.

After inspecting with react-devtools, it appears that reducing the yOffset by 24 puts the position in the correct place.

Issue #1743, opened 2019-02-01
A question about "Statusbar" and "toolTip"


dbenzzz:
On android devices, I use "Header" and "Tooltip" at the same time, set the "Statusbar" component "translucent" value is true. When I touch the inside of the "Tooltip" Icon, tip appears on the interface, but also the emergence of a same Icon, I want to make this extra Icon disappear, does anyone have a good plan?I am a beginner, and my English is very bad, I do not know if you can understand, but thank you very much!
![ccc](https://user-images.githubusercontent.com/47247128/52125366-3689d380-2667-11e9-9b52-0f922a0a57aa.png)
![ddd](https://user-images.githubusercontent.com/47247128/52125376-45708600-2667-11e9-8190-674cce6b301a.png)

michaelclark2:
this appears to happen with tooltip on android even on a blank app

#1744

PR #1748, opened 2019-02-03
Tooltip: reduces yOffset by 24 if android to position correctly

michaelclark2:
This fixes the improper positioning of the Tooltip when the Tooltip is visible.  
I discovered the yOffset value was off by 24, so I compensated for that difference where it counts.


michaelclark2:
#1743 #1744

PR #1749, opened 2019-02-03
Fix mix of import and module.exports

yatsyk:
Similar bug was fixed previously: https://github.com/react-native-training/react-native-elements/pull/448

eslint rule could be useful to prevent this in future.


iRoachie:
Thanks! 💯

PR #1751, opened 2019-02-04
reduces yOffset by StatusBar.currentHeight on android devices

michaelclark2:
fixes issues #1744 #1743 

compensates for the StatusBar.currentHeight to position the active tooltip properly on android devices.

### DIFF
--- a/src/tooltip/Tooltip.js
+++ b/src/tooltip/Tooltip.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { TouchableOpacity, Modal, View } from 'react-native';
+import { TouchableOpacity, Modal, View, StatusBar } from 'react-native';
 
 import { ViewPropTypes, withTheme } from '../config';
 import { ScreenWidth, ScreenHeight, isIOS } from '../helpers';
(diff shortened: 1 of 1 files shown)


kufii:
format like this so that it passes the build

```javascript
            yOffset: isIOS
              ? pageOffsetY
              : pageOffsetY - StatusBar.currentHeight,
```

iRoachie:
Thank you! Good catch 💯. Works fine in my testing

kufii:
@iRoachie Just thought I should let you know, I just ejected from expo and I'm no longer experiencing the tooltip positioning issue, so this pull request might negatively effect non-expo users

michaelclark2:
@kufii I unfortunately did not test for this edge case, and your assumption is correct.  would it be viable to check if the project is expo or not, in the same way you check for the platform?

kufii:
@michaelclark2 @iRoachie I found another edge case, when running with expo, run StatusBar.setHidden(true) and the issue stops occurring. Honestly, with so many edge cases, I think the best solution is to not bother trying to have it perfectly positioned for every scenario by default, and just add a `yOffset` prop that the developer can add themselves if the positioning is messed up. Thoughts?

Issue #1747, opened 2019-02-03
[web] updateTheme property on react-dom


hedikasmanto:
**Explain what you did**
I use "react-native-web" + "react-native-elements" with ThemeProvider like this below render

```
<ThemeProvider>
        <View>
          <Header
            leftComponent={{ icon: 'menu', color: '#fff' }}
            centerComponent={{ text: 'MY TITLE', style: { color: '#fff' } }}
            rightComponent={{ icon: 'home', color: '#fff' }}
          />
     </View>
</ThemeProvider>

```
**Expected behaviour**
No Error should be thrown

**Describe the bug**
Warning: React does not recognize the `updateTheme` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `updatetheme` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in div (created by Text)
    in Text (at Text.js:12)
    in TextElement (at withTheme.js:42)
    in Themed.Text (at renderNode.js:18)
    in div (created by View)
    in View (at Header.js:26)
    in Children (at Header.js:94)
    in div (created by View)
    in View (created by ImageBackground)
    in ImageBackground (at Header.js:70)
    in Header (at withTheme.js:42)
    in Themed.Header (at App.js:21)
    in div (created by View)
    in View (at App.js:20)
    in ThemeProvider (at App.js:19)
    in App (at js/index.js:6)

**Your Environment (please complete the following information):**

| software              | version |
| --------------------- | ------- |
| react-native-elements |  1.0.0       |
| react-native          |   web 0.10      |
| npm or yarn           |    yarn 1.13     |

xavier-villelegier:
@hedikasmanto Does it work even with this warning ?
Actually the `ThemeProvider` injects two props: `theme` and `updateTheme`, that provides you the ability to update the theme dynamically.

I guess react-native-web transforms the `View` into a `div` or something similar, so you end up with a `div` that has an `updateTheme` props, obviously not a standard DOM props.

Maybe you could just remove the `View` that is wrapping your `Header` component ? If you really want to get rid of this warning, and you doesn't need to update your theme dynamically, you can still write a custom component and exclude the `updateTheme` props manually.

hedikasmanto:
Yes. It still working. And I found, withTheme HoC cause this issue..

xavier-villelegier:
[...]

PR #1758, opened 2019-02-06
Fix withTheme

hedikasmanto:
This is patch withTheme script for exclude updateTheme prop on react-dom (RNW) #1747

### DIFF
--- a/src/config/withTheme.js
+++ b/src/config/withTheme.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { Platform } from 'react-native'; 
 import merge from 'lodash.merge';
 import hoistNonReactStatics from 'hoist-non-react-statics';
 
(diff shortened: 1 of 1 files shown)


xavier-villelegier:
I don't think it is the way to go. What if you want to update the theme dynamically on the web ?

cc @iRoachie

hedikasmanto:
Yeah. Before I submitted this patch, I have known dynamic theme will not working on web. But currently, I will accept this limitation, and I will search another workaround for dynamic theme purpose. on web

xavier-villelegier:
@hedikasmanto I understand that you would accept this limitation but maybe this is not the case for everyone haha 😅 I don't use React Native Web, but if I do I'd personally rather have a warning that not having this feature.

There is surely a way to avoid this warning without removing the feature. Did you try wrapping a custom component and not passing the `updateTheme` props to the children ? Try to not use the `{...props}` syntax in the wrapped component, because it will spread the `updateTheme` too

hedikasmanto:
First of all, CMIIW! I think if I use wrapped component, I will have to wrap more than one component to use theme from this library, and create many custom components that exclude updateTheme props? or something that I missed ? Please elaborate me.
And finally, If this PR can't be accepted (includes the limitation 😅haha), it's ok. I think I will go with keep warning exists.

iRoachie:
Proper fix released for this in [`v1.2.4`](https://github.com/react-native-training/react-native-elements/releases/tag/v1.2.4)

PR #1760, opened 2019-02-06
Added divider color in TypeScript definition

danielholmes:
From `src/config/colors.js`: https://github.com/react-native-training/react-native-elements/blob/next/src/config/colors.js#L19


iRoachie:
Must've slipped my mind 🙈. Thanks

PR #1761, opened 2019-02-06
Add ReactElement<{}> to ListItem.chevron and checkmark

danielholmes:
`ListItem` props `chevron` and `checkmark` [are rendered](https://github.com/react-native-training/react-native-elements/blob/master/src/list/ListItem.js#L213):

```js
{renderNode(Icon, checkmark, checkmarkDefaultProps(theme))}
{renderNode(Icon, chevron, chevronDefaultProps)}
```

Using the [`renderNode` helper](https://github.com/react-native-training/react-native-elements/blob/master/src/helpers/renderNode.js), which includes the option of a `ReactElement`:

```js
const renderNode = (Component, content, defaultProps) => {
  if (content == null || content === false) {
    return null;
  }
  if (React.isValidElement(content)) {
    return content;
  }
```

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1099,8 +1099,8 @@ export interface ListItemProps {
   containerStyle?: StyleProp<ViewStyle>;
   contentContainerStyle?: StyleProp<ViewStyle>;
   rightContentContainerStyle?: StyleProp<ViewStyle>;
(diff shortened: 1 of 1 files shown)


iRoachie:
Good catch, thanks

PR #1769, opened 2019-02-09
fix: add underlayColor to SocialIconProps interface

Rawphs:
Added `underlayColor` to the `SocialIconProps` interface.

I wasn't sure as for what to write as a default, since it is only white if the button is light and the default color varies according to the button type.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1775,7 +1775,7 @@ export interface SocialIconProps {
   /**
    * Specify underlayColor for TouchableHighlight
    *
-   * @default colors[type] || 'white'
+   * @default 'white' if `light` prop is true, otherwise defaults to icon color.
    */
   underlayColor?: string;
 }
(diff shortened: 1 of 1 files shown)


src/index.d.ts
```diff
@@ -1771,6 +1771,13 @@ export interface SocialIconProps {
    * @default false
    */
   loading?: boolean;
+  
+  /**
+   * Specify underlayColor for TouchableHighlight
+   *
+   * @default colors[type] || 'white'
```
iRoachie:
```suggestion
   * @default 'white' if `light` prop is true, otherwise defaults to icon color.
```

iRoachie:
Hey there @Rawphs! Thanks for sending this in and congrats on your first PR to React Native Elements.

There's just one change to the description I'd like to make, but other than that this is fine!

src/index.d.ts
Rawphs: @iRoachie Done!

Rawphs:
cmfkfgp - imgur

iRoachie:
Thanks! 💯

PR #1772, opened 2019-02-11
fix: Change merge library to deepmerge from lodash

### LINKED ISSUE #1735: [ThemeProvider] Themed.ListItem should have a unique "key" prop
**Explain what you did**
I'm using ThemeProvider with FlatList and ListItem

**Expected behaviour**
No warning message.

**Describe the bug**
When I use ThemeProvider with FlatList and ListItem the warning messages appears:
Warning: Each child in an array or iterator should have a unique "key" prop. See https://fb.me/react-warning-keys for more information.

 in ListItem (at withTheme.js:42)
 in Themed.ListItem (at Item.jsx:120)

**To Reproduce**
ThemeProvider  with FlatList and ListItem
And yes, I'am using keyExtractor in FlatList.

| react-native-elements |   1.0.0      |
| react-native          |    0.57.7     |
| yarn           |    1.13.0     |

iRoachie:
While using ThemeProvider, some errors would be shown for some
components for arrays not having keys. This was due to the merge
algorithm in lodash.merge which happens to affect react's keys.

Switching to deepmerge instead of lodash.merge does not have this
issue but maintains expected behaviour.

Fixes #1735 #1517

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,8 +34,8 @@
   },
   "dependencies": {
     "color": "^3.1.0",
+    "deepmerge": "^3.1.0",
     "hoist-non-react-statics": "^3.1.0",
-    "lodash.merge": "^4.6.1",
     "opencollective-postinstall": "^2.0.0",
     "prop-types": "^15.5.8",
     "react-native-ratings": "^6.3.0",
--- a/src/config/ThemeProvider.js
+++ b/src/config/ThemeProvider.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import merge from 'lodash.merge';
+import deepmerge from 'deepmerge';
 
 import colors from './colors';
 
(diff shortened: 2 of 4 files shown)


M-Jas:
@iRoachie not sure if anyone else has comment yet, but deepmerge is causing some issues with React Native Navigation(or any objects that proto methods on it) on ```withStyles```. Specifically it's stripping the prototypes off of objects since is cloning and not copying the object. [deepmerge issue](https://github.com/TehShrike/deepmerge/issues/118). Wanted to drop a note before opening an issue. 

Why do we need to return a new object every single update?

Also should I toss up a PR to update how we are returning the value from the context? [Context Caveats](https://reactjs.org/docs/context.html#caveats)?

iRoachie:
Oh i was not aware of this @M-Jas. Definitely would love to open an issue where we could talk about it more.

PR #1781, opened 2019-02-14
Should show clear icon when value is not empty

vellengs:
should show clear icon when value is not empty when this.props.value;

### DIFF
--- a/src/searchbar/SearchBar-ios.js
+++ b/src/searchbar/SearchBar-ios.js
@@ -32,7 +32,6 @@ const defaultClearIcon = {
 class SearchBar extends Component {
   constructor(props) {
     super(props);
-    const { value } = props;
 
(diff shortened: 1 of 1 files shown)


iRoachie:
I'm not understanding the purpose of this PR. The iOS SearchBar already does this behaviour.

vellengs:
If I set the search bar value as program set. ( value change event will not trigger) then close icon will not show.(that isn't same with IOS search bar).

vellengs:
image
I want if search bar has value then always show the close icon .

iRoachie:
This is how the SearchBar functions right now. Once there is a value in the SearchBar the close icon shows. This is how the native iOS search bar behaves.

vellengs:
Might show close icon when focus on?

Issue #1787, opened 2019-02-18
 Button ripple effect overflows outside button container Even without borderRadius


aliassar:
**Explain what you did**
I wrote a simple button with react-native-element .
` <Button onPress={() => this.login()} title='login' containerStyle={styles.buttonContainer} />`

**Expected behaviour**
I expected ripple not to be borderless.

**Describe the bug**
ripple is borderless.

**To Reproduce**

Example:

https://snack.expo.io/@ali.assar/button-problem
# Steps

    1. Click on 'Button'

    2. See the overlay

**Possible Solution**
In line 65 of Button.js maybe changing the bloolean from true to false (borderless property) can fix it.

` if (Platform.Version >= 21) { attributes.background = TouchableNativeFeedback.Ripple( 'ThemeAttrAndroid', true ); }`
----------------------------------->>>>>>>> ` if (Platform.Version >= 21) { attributes.background = TouchableNativeFeedback.Ripple( 'ThemeAttrAndroid', false ); }`

**Screenshots**

![untitled](https://user-images.githubusercontent.com/40848213/52952444-cfc02600-3399-11e9-9bbd-4553712a31e2.png)

**My Environment**

| software | version |
| react-native-elements | 1.0.0|
| react-native | 0.58 |
| yarn | 1.13.0|

xavier-villelegier:
Hey @aliassar 👋 , sorry for the delay !
I just tested it on Snack, and apparently the solution you provided fixes the bug ([repro here](https://snack.expo.io/@xavier-villelegier/tactless-celery))

Would you mind opening a PR to fix it ? 💯

aldoreyes:
This is still happening in 3.1.0, was this rolled back for some reason? I don't see the merged change from PR #1799 on 3.1.0

andrei0807:
yes me too.
I can see same error.

FerminIriarte-0324:
I also same issue but update android version from 9 to 10 with API 29. this issue disappears.
Development environment is geomotion/win10.

HugoGresse:
Just updated to 3.3.0, ripple is crazy big :/

flyingcircle:
@HugoGresse just tested in snack and was not able to reproduce. Maybe create a snack showing the issue.

koaladlt:
I had a similar issue. My mistake was to use margin on "buttonStyle" prop. Using it on "containerStyle" fixed the problem.

mdovn:
[...]

PR #1799, opened 2019-02-23
Fix button out of border ripple

aliassar:
According to the #1787 the ripple of TouchableNativeFeedback in Button will overflow out of box due to value of borderless variable & it changed in this pull request.


iRoachie:
Thanks again 🎖

Issue #1755, opened 2019-02-06
Keep fixed height for errorMessage prop on Input


iRoachie:
**Is your feature request related to a problem? Please Describe.**
Currently, when the error message is shown on the Input component, it changes the layout of the screen. This is because currently the Text component displaying the error message is not rendered when the prop is falsy.

**Describe the solution you'd like**
Always render the error message, so the height is consistent but use another means to hide and show it like opacity.

**Additional context**
It behaved this way in 0.19.1 with FormValidationMessage

PR #1801, opened 2019-02-24
Fix: Height for errorMessage prop on Input

saxenanihal95:
According to the #1755 the component showing error message should always render.

### Before
bad

### After
good


iRoachie:
Hey @saxenanihal95 this doesn't actually fix the problem. Setting `display: none` will still cause other elements on the screen to shift. Display none is the same as not rendering.

saxenanihal95:
@iRoachie my bad i will update the PR

saxenanihal95:
@iRoachie updated PR used opacity instead of display

iRoachie:
@xavier-villelegier We have another problem that will come by using this. Components that use the Input component like the SearchBar will end up displaying with extra space.

image

Maybe instead should add a prop to always show error message ?

saxenanihal95:
I think one way is to separate `InputComponent` which can be used by `Input` and others components like `SearchBar`.

iRoachie:
What I think we can do is add a prop on the input maybe called `renderErrorMessage` that the SearchBar component could use to achieve the styling we have atm.

This will be a breaking change anyway since it affects layout. Let’s do this for 2.0.0

saxenanihal95:
ok @iRoachie so we will add a prop `renderErrorMessage` which will be `true` by default and when `Search` uses Input it will make `renderErrorMessage`  `false` ?

iRoachie:
Yea that's what I was thinking

saxenanihal95:
ok then will update the PR according to that

iRoachie:
Also you need to rebase your branch from `next`. Some conflicts there

saxenanihal95:
ok i will rebase

saxenanihal95:
I have fixed the conflicts now will make changes according to the discussion we have

iRoachie:
There's no changes now in the PR?

saxenanihal95:
yes not right now

saxenanihal95:
@iRoachie Updated the PR

iRoachie:
The default would be true, also you need to add to the typescript definitions.

Also the search bars would need to be updated to set this property to false.

Think we should leave this to 2.0.0 cause it’ll break some layouts that people have in their apps

saxenanihal95:
ok then i will wait for 2.0.0

iRoachie:
@saxenanihal95 Are you still interested in making this feature?

saxenanihal95:
@iRoachie yes

src/input/Input.js
```diff
@@ -138,15 +139,15 @@ class Input extends React.Component {
...
+        {renderErrorMessage && (
           <Text
             {...errorProps}
             style={StyleSheet.flatten([
               styles.error(theme),
               errorStyle && errorStyle,
             ])}
           >
-            {errorMessage}
+            {errorMessage || ''}
```
iRoachie:
```suggestion
            {errorMessage}
```

saxenanihal95: @iRoachie but In all SearchBar renderErrorMessage is already false ?

src/searchbar/SearchBar-android.js
```diff
@@ -142,6 +142,7 @@ class SearchBar extends Component {
             styles.rightIconContainerStyle,
             rightIconContainerStyle,
           ])}
+          renderErrorMessage={false}
```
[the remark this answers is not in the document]
iRoachie: Sorry I missed that.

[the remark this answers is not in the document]
iRoachie: Just the docs and types then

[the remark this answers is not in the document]
saxenanihal95: ok sure

docs/input.md
```diff
@@ -243,6 +243,14 @@ styling for right Icon Component container
 | :-----------------: | :-----: |
 | View style (object) |  none   |
 
+### `renderErrorMessage`
+
+displays errorMessage
```
iRoachie:
```suggestion
If the error message container should be rendered (take up vertical space). If `false`, when showing errorMessage, the layout will shift to add it at that time.
```

src/index.d.ts
```diff
@@ -879,6 +879,12 @@ export interface InputProps extends TextInputProperties {
    *  props to be passed to the React Native Text component used to display the label (optional)
    */
   labelProps?: TextProps;
+
+  /**
+   *  displays error message
+   */
+
```
iRoachie:
```suggestion
```

iRoachie:
You need to update the snapshots for the ci to pass. https://react-native-elements.github.io/react-native-elements/docs/testing.html#snapshot-testing

saxenanihal95:
> You need to update the snapshots for the ci to pass. https://react-native-elements.github.io/react-native-elements/docs/testing.html#snapshot-testing

 @iRoachie Done

iRoachie:
Okay, we need to make some last-minute changes. Right now if `renderErrorMessage={false}` and `errorMessage="Test"` nothing shows at all. Which means right now users would be unable to get back the previous behaviour.

I made a table to explain the different states:

| renderErrorMessage | errorMessage | message | take up space (render)
| -- | -- | -- | -- |
| true | true | yes | yes |
| true | false | no | yes |
| false | false | no | no |
| false | true | yes | yes | 

I've managed to get the correct behaviour by making a constant:

```jsx
const hideErrorMessage = !renderErrorMessage && !errorMessage;


<Text
  {...errorProps}
  style={StyleSheet.flatten([
    styles.error(theme),
    errorStyle && errorStyle,
    hideErrorMessage && {
      height: 0,
      margin: 0,
      padding: 0
    },
  ])}
>
  {errorMessage}
</Text>
```

saxenanihal95:
@iRoachie test is getting passed locally but failing here

saxenanihal95:
@iRoachie any update on this ?

saxenanihal95:
@iRoachie I have updated the snapshots and now tests are passing

flyingcircle:
@saxenanihal95 can you squash your changes into 1 or 2 commits?

flyingcircle:
@saxenanihal95 I tested it out, but I don't see it working. Are you able to share an example of this change working?

saxenanihal95:
@flyingcircle are you talking about jest test ?

flyingcircle:
@saxenanihal95 No, I mean I imported your changes into an app and still saw the error causing a change in rendering margin:
Here was my component:
```
<Input
  containerStyle={[styles.inputContainerStyle]}
  placeholder="Shake input"
  ref={ref => (this.shakeInput = ref)}
  rightIcon={
    <Button
      title="Shake"
      onPress={() => {
          this.shakeInput && this.shakeInput.shake()
          if (this.state.shakeError) {
            this.setState({shakeError: ""});
          } else {
              this.setState({shakeError: "Shake me on error !"});
          }
      }}
    />
  }
  errorMessage={this.state.shakeError}
/>

```

app-error

src/input/Input.js
```diff
@@ -172,6 +178,11 @@ Input.propTypes = {
   labelStyle: TextPropTypes.style,
   labelProps: PropTypes.object,
   theme: PropTypes.object,
+  renderErrorMessage: PropTypes.bool,
+};
+
+Input.defaultProps = {
```
saxenanihal95, line 184: @flyingcircle thanks for pointing out actually defaultProps was already defined and my defined default props was overriding with that

Will update the PR

saxenanihal95: @flyingcircle I have updated the PR now it will work

flyingcircle:
@saxenanihal95 thanks for your work!

AzzouDuGhetto:
What about the TypeScript definition ? It seems like @iRoachie delete it but why ?

flyingcircle:
@AzzouDuGhetto this looks like it was a merging error on my part. Somehow the removal made it into my PR. Should be fixed.

AzzouDuGhetto:
Forgive my lack of knowledge but isn't #2323 remove it once again ? Or the change will be available later ?

flyingcircle:
@AzzouDuGhetto I pushed a fix to the `next` branch. It will be included in 2.0.1.

PR #1803, opened 2019-02-25
Should show clear icon  onFocus

vellengs:
Should show clear icon when onFocus value not empty

### DIFF
--- a/src/searchbar/SearchBar-ios.js
+++ b/src/searchbar/SearchBar-ios.js
@@ -63,7 +63,8 @@ class SearchBar extends Component {
   onFocus = () => {
     this.props.onFocus();
     UIManager.configureNextLayoutAnimation && LayoutAnimation.easeInEaseOut();
-    this.setState({ isEmpty: text === '' });
+    const { value } = props;
+    this.setState({ isEmpty: value === '' });
     this.setState({
       hasFocus: true,
     });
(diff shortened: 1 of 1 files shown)


vellengs:
#1781 I checked with ios native components like safari search bar,  focus on should show that icon.

xavier-villelegier:
It's already the case. If the searchbar isn't empty, we show the clear icon (it doesn't depend on the focus)
image

Repro https://snack.expo.io/@xavier-villelegier/searchbar-clear

vellengs:
@xavier-villelegier did you make sure the value set program set not initiate value or type to input?

Lythenas:
I can confirm that it is not automatically displayed when changing the value programmatically (at least on android). See  #1846

vellengs:
@Lythenas I did a PR before as check it in render, but that 's real different behavior, sometimes we need not show clear button, but if on editing should be shows that icon button.

Lythenas:
@vellengs currently the clear icon only shows up when you type in the search bar and it stays visible even if it looses focus and disappears when the search bar is empty. But it does not show up when you set the value without the user typing anything.

That would be fixed if instead of having an `isEmpty` state that is set when the value changes, you just check if the value is empty in render. That wouldn't change anything else.

PR #1811, opened 2019-03-02
The value of keyExtractor must be a string

YIPG:
The value of keyExtractor must be a string. -[related issue](https://github.com/facebook/react-native/issues/18291)

Now the warning like below appeared.
```
Warning: Failed child context type: Invalid child context `virtualizedCell.cellKey` of type `number` supplied to `CellRenderer`, expected `string`.
```

### DIFF
--- a/docs/listitem.md
+++ b/docs/listitem.md
@@ -93,7 +93,7 @@ const list = [
   ... // more items
 ]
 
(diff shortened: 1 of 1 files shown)


iRoachie:
Thanks 💯

PR #1818, opened 2019-03-07
Update social_icons.md

TeruyaHaroldo:
It is possible to use the Google logo (G) and not only the Google Plus logo (G+).


iRoachie:
We should probably allow Google button here as well in the source

PR #1825, opened 2019-03-12
fix(Button-android): remove wrong color argument

ceyhuno:
The first argument of TouchableNativeFeedback.Ripple function takes a color as a string.
https://github.com/facebook/react-native/blob/master/Libraries/Components/Touchable/TouchableNativeFeedback.android.js#L169

### DIFF
--- a/src/buttons/Button.js
+++ b/src/buttons/Button.js
@@ -64,7 +64,7 @@ class Button extends Component {
     ) {
       if (Platform.Version >= 21) {
         attributes.background = TouchableNativeFeedback.Ripple(
-          'ThemeAttrAndroid',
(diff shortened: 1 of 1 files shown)


iRoachie:
What problem is this solving?

ceyhuno:
It shouldn’t change the behaviour. It’s just that ‘TouchableNativeFeedback.Ripple('ThemeAttrAndroid')’ is wrong, ‘ThemeAttrAndroid’ is not a color.

iRoachie:
Gotcha!

iRoachie:
Thanks for PR and explanation 🥇

PR #1830, opened 2019-03-13
fix: make ImageComponent a ComponentType

SimenB:
This allows use to use a `FunctionComponent` and not just a class


iRoachie:
Thanks 💯

PR #1835, opened 2019-03-20
feat: Update Card to use new Image Component

### LINKED ISSUE #1687: Update Card to use new Image Component
Now that we have the  [Image component that allows gradual image loading](https://github.com/react-native-training/react-native-elements/pull/1582)  we should use it in the Card component.

haruelrovix:
Close #1687

### DIFF
--- a/src/card/Card.js
+++ b/src/card/Card.js
@@ -1,18 +1,19 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { View, Platform, Image, StyleSheet } from 'react-native';
+import {
+  ActivityIndicator,
+  View,
+  Platform,
+  Image as ImageNative,
+  StyleSheet,
+} from 'react-native';
 
 import normalize from '../helpers/normalizeText';
-import {
-  BackgroundImage,
-  fonts,
-  TextPropTypes,
-  ViewPropTypes,
-  withTheme,
-} from '../config';
+import { fonts, TextPropTypes, ViewPropTypes, withTheme } from '../config';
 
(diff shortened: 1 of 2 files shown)


src/card/Card.js
```diff
@@ -80,9 +81,10 @@ const Card = props => {
 
         {image && (
           <View style={imageWrapperStyle && imageWrapperStyle}>
-            <BackgroundImage
+            <Image
               style={[{ width: null, height: 150 }, imageStyle && imageStyle]}
               source={image}
+              PlaceholderContent={<ActivityIndicator />}
```
iRoachie: I'd say remove `PlaceholderContent` just to be consistent with the other components that use Image

iRoachie:
I also updated your branch from next, so be aware of that when you start working locally again.

iRoachie:
Also this will fail when it has children.

Simulator Screen Shot - iPhone 7 - 2019-08-01 at 08 59 46

iRoachie:
#1992 Made the change so that the Image component uses ImageBackground so this problem should be solved now. Once you do a rebase we should be good to go

src/card/Card.js
haruelrovix: @iRoachie updated via [f8858bc](https://github.com/react-native-training/react-native-elements/pull/1835/commits/f8858bcd36ad9b2494c0f9a3e0e062e60009cb29).

iRoachie:
Thanks again @haruelrovix real mvp 💯

PR #1642, opened 2018-12-05
[Card.js] onPress & onPressColor added on image

achillelrc:
PR branch rebased to next [#1584](https://github.com/react-native-training/react-native-elements/pull/1584)
Issue related [#814](https://github.com/react-native-training/react-native-elements/issues/814)

+optional prop imageOnPress
trigger when image header is pressed

+optional prop  imageOnPressProps
change properties of TouchableHighlight element

PR #1841, opened 2019-03-25
[Card.js] imageOnPress & imageOnPressProps

rub1e:
Hi, apologies if this isn't the right way to submit this, but I saw PR #1642 was nearly there, and I very much need this functionality, so I just finished it off with some tests.

Issue: #814 

* Finish off PR #1642 
* Add tests 
* (Tried and failed to fix the issue with the `tooltip` tests - Jest won't recognise `**` as valid javascript, but linting forbids `Math.pow()` - so I left it)

### DIFF
--- a/src/card/Card.js
+++ b/src/card/Card.js
@@ -1,6 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { View, Platform, Image, StyleSheet } from 'react-native';
+import {
+  View,
+  Platform,
+  Image,
+  TouchableHighlight,
+  StyleSheet,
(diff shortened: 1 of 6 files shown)


iRoachie:
Can you fix the conflicts here?

src/card/Card.js
```diff
@@ -17,6 +23,7 @@ import Divider from '../divider/Divider';
 const Card = props => {
   const {
     children,
+    flexDirection,
```
iRoachie: Why did you add these props in?

`flexDirection`, and `fontFamily`

Issue #1827, opened 2019-03-13
Change overlay color in Tooltip by props


cleoreo:
First of all, thanks of this awesome library!

Can you add change overlay color props to Tooltip component? I found no way to change the color.

I hoped I can change the overlay color like this.

```
<Tooltip popover={<Text>Info here</Text>} overlayColor={'rgba(0, 0, 0, 0.6)'}>
  <Text>Press me</Text>
</Tooltip>
```

Added a pull request to add this ( #1828 )

iRoachie:
You can use the `windowBackgroundColor` prop https://snack.expo.io/@roach_iam/bad-sandwich

cleoreo:
Thank you for your reply!

I wanna change the tooltip overlay to something like below, and your example cannot help.
If #1828 pull request is not okay, can you suggest another way to achieve the following please?

Thanks again!


![Screen Shot 2019-03-18 at 10 36 17 AM](https://user-images.githubusercontent.com/22050666/54503391-002fbb80-496a-11e9-97c0-fcf8d9728010.png)

iRoachie:
Sorry about that, I saw overlay and thought you were referring to the Overlay component

PR #1868, opened 2019-04-18
Add isEmpty check to onFocus of the Search Bar

M-Jas:
## What does this PR do?
- Updates the ```onFocus``` function in the Search Bar(iOS & Android) to set the state of the ```isEmpty``` prop to show the the Clear icon when the Search Bar becomes focused when an initial value is given.

## Any background or context you want to provide?
- The search bar only sets the ```isEmpty``` state in the constructor and in the```onChangeText``` function. We ran into an issue were a value is given to the Search Bar(via redux/prop passing) after the Search Bars has already mounted, the user would focused the Search Bar and the clear icon would not be displayed until they changed the text in the input.

## Expected vs Actual
- Expected: As a user when I'm taken to a new screen with the search text/value is presented in the new Search Bar, and when I focus the search bar and I can press the the clear icon to start a fresh search.
-Actual: The user focuses the search bar and no clear icon is present.

## Images & Gifs
- Before focus on a new search bar with a value give via props
Screen Shot 2019-04-18 at 9 08 38 AM

- After focus on a new search bar with a value give via props(w/o pr changes)
Screen Shot 2019-04-18 at 9 09 03 AM

- Expected After focus on a new search bar with a value give via props
Screen Shot 2019-04-18 at 9 10 01 AM

Fixes #1846

PR #1869, opened 2019-04-18
Add overlayColor prop to toolTip for users to set the background color

M-Jas:
## What does this PR do?
- Adds a prop to the toolTip to change the overlay background color. If there is a prop to add an overlay, there should be a prop to change the color of that overlay. #1827

## Any background or context you want to provide?
- We are using the tool tip as a floating action button, on an app that our customers can theme(with their branding). Currently the users would only see the default white and this would cause issues. 

- Users can still change the opacity if they want

## Expected vs Actual
- Expected: If I change the color of the overlay color to red the user will see a red background/shadow to the overlay.
- Actual: We(library users) can't change the color so we are stuck with the default white.

## Images & Gifs
Simulator Screen Shot - iPhone XS - 2019-04-18 at 10 40 57

### DIFF
--- a/docs/tooltip.md
+++ b/docs/tooltip.md
@@ -29,12 +29,12 @@ import { Tooltip, Text } from 'react-native-elements';
 - [`highlightColor`](#highlightColor)
 - [`onClose`](#onClose)
 - [`onOpen`](#onOpen)
+- [`overlayColor`](#overlaycolor)
 - [`pointerColor`](#pointerColor)
 - [`popover`](#popover)
 - [`toggleOnPress`](#toggleOnPress)
 - [`width`](#width)
 - [`withOverlay`](#withoverlay)
-- [`overlaycolor`](#overlaycolor)
 - [`withPointer`](#withpointer)
 
 ---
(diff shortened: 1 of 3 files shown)


ronffy:
I need this feature,
I hope the author can handle it as soon as possible.

iRoachie:
Awesome PR 👏 Can you also add this to the typescript definitions?

docs/tooltip.md
```diff
@@ -34,6 +34,7 @@ import { Tooltip, Text } from 'react-native-elements';
 - [`toggleOnPress`](#toggleOnPress)
 - [`width`](#width)
 - [`withOverlay`](#withoverlay)
+- [`overlayColor`](#overlayColor)
```
iRoachie:
```suggestion
- [`overlayColor`](#overlaycolor)
```

M-Jas:
@iRoachie I fixed prop naming per your request. I always forget about the typescript definitions, sorry about that.

iRoachie:
Hey @M-Jas thanks for your work in the last release. We have an opencollective in which developers and companies contribute finances to help push forward the development of this library.

Can you submit an open collective expense of $150? This would've been for your work here as well as https://github.com/react-native-training/react-native-elements/pull/1868.

You can put the description as `Tooltip and SearchBar enhancements for 1.2.0`. Here's the link - https://opencollective.com/react-native-elements/expenses/new

Issue #1850, opened 2019-03-29
Tile image resizeMode


jcroft:
**Explain what you did**
Wanted to change my image in a Tile to resizeMode: contain

**Expected behaviour**
Expected there to be some prop to take this option

**Describe the bug**
It seems to me that resizeMode is always set to cover and cannot be changed.

Is there a reason for this, or a way to do it that I'm not seeing?

Thanks for such a great library!

jcroft:
Thanks so much! :)

ethanyuwang:
having the same issue

iRoachie:
@ethanyuwang Tile now has an `imageProps` prop in which you can pass in the values

PR #1873, opened 2019-04-21
feat(Tile): Add imageProps prop

iRoachie:
Allow customizing additional props on the image component.
Implements #1850


deadcoder0904:
@iRoachie when will this release? I need `imageProps` too :)

PR #1894, opened 2019-05-20
[TS] export type definition RecursivePartial<T>

mjmaix:
Need to have RecursivePartial to extend FullTheme for adding variables on colors. For example: adding `colors. backgroundColor`

```
interface ExtendedFullTheme extends FullTheme {
  colors: RecursivePartial<CustomColors>;
}

export const themes: { [key in ThemeName]: Partial<ExtendedFullTheme> } = {
  'JUST WHITE': {
    colors: {
      primary: '#000000',
      backgroundColor: '#fffff',
    },
  },

...

```

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -2014,7 +2014,7 @@ export function normalize(size: number): number;
  */
 export function registerCustomIconType(id: string, font: any): void;
 
-type RecursivePartial<T> = { [P in keyof T]?: RecursivePartial<T[P]> };
+export type RecursivePartial<T> = { [P in keyof T]?: RecursivePartial<T[P]> };
 
 export interface FullTheme {
   Avatar: Partial<AvatarProps>;


iRoachie:
I think you could do this without making the change:

```ts
import {
  Theme,
  ThemeProvider,
  Colors,
  ThemeConsumer,
} from 'react-native-elements';

interface MyTheme extends Theme {
  colors: {
    backgroundColor: string;
  } & Partial<Colors>;
}

const theme: MyTheme = {
  colors: {
    backgroundColor: 'pink',
  },
};

  render() {
    return (
      <ThemeProvider theme={theme}>
          <ThemeConsumer<MyTheme>>
            {({ theme }) => (
              <Text style={{ color: theme.colors.backgroundColor }}>
                Open up App.js to start working on your app!
              </Text>
            )}
          </ThemeConsumer>
        </View>
      </ThemeProvider>
    );
  }
}
```

mjmaix:
Thanks @iRoachie I will try

Issue #1816, opened 2019-03-05
Android: Unexpected token '.' (http://10.0.2.2:8081/index.delta?platform=android&dev=true&minify=false:88794) at


duycoder:
**What i did**
Upgrade from version **0.19.0** to latest **1.1.0** using "yarn" to use attribute **rightElement** of **ListItem** which is not available in version **0.19.0**

**Describe the bug**
The app got the red screen

**Possible Solution**
I had to revert back to version **0.19.0**, and the app worked fine.

**Screenshots**
![40263567-81247a02-5b3e-11e8-8d53-58b093d4e586](https://user-images.githubusercontent.com/16200019/53804963-7cde9500-3f7b-11e9-9708-2a7c13035918.png)

**Your Environment (please complete the following information):**

| software              | version |
| --------------------- | ------- |
| react-native-elements |    1.1.0    |
| react         |  16.3.1 |
| react-native          |  0.55.3       |
| yarn           | 1.13.0 |

pilot4u:
I have the version except RN 0.55.4
and getting error: "Unexpected token *" on app launch
![photo_2019-03-06_01-11-54](https://user-images.githubusercontent.com/28202403/53843967-e0cd8180-3fac-11e9-9ffe-8b808f7d3df2.jpg)

xavier-villelegier:
Mmh I see you also commented on https://github.com/facebook/react-native/issues/19350, which has no link with RNE, so I'm not really sure that this issue is strictly because of RNE (we would have heard about this issue sooner)

Kinda mainstream but did you try to clean install again your node_modules, and restart your RN server ? Be sure to remove any `yarn.lock` file before

duycoder:
Can i ask why `yarn.lock` related to this issues?

xavier-villelegier:
It doesn't, it's just to be sure to not take old versions of packages or dependencies of RNE. It sometimes has done the job for some reasons. As it's probably not related to RNE itself, and I'm unable to reproduce this locally, making a clean install is still a good start

duycoder:
Well, i did clear node_modules, remove `yarn.lock` and re-install every dependencies still cost me the same error, here is my `package.json`
{
"lodash": "^4.17.10",
"moment": "^2.22.1",
"native-base": "^2.4.3",
"npm": "^6.1.0",
"react": "16.3.1",
"react-native": "0.55.3",
"react-native-calendars": "^1.22.0",
"react-native-dash": "^0.0.8",
[...]

PR #1898, opened 2019-05-22
Switch from `**` to `Math.pow` for getTooltipCoordinate

dominikeinkemmer:
On my Android phone (Google Pixel 3) I received an error after running `react-native run-android` saying `Unexpected token "*" at ...`. (same as in this closed issue: https://github.com/react-native-training/react-native-elements/issues/1816#issuecomment-473521739)

After doing some painfully long research I found that using `**` instead of `Math.pow` can cause issues in that regard so I when I found that it is used in this project I changed this and my App ran smoothly again.

This PR switches is here to sync this change to `react-native-elements` and changes from `**` to `Math.pow` for `getTooltipCoordinate`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,16 +150,60 @@ See you on the other side! 👋😃
 
 ## Backers
 
-Support us with a monthly donation and help us continue our activities.
-[[Become a backer](https://opencollective.com/react-native-elements#backer)]
+[Become a backer](https://opencollective.com/react-native-elements#backer) and show your support for React Native Elements.
 
-<a href="https://opencollective.com/react-native-elements#backers" target="_blank"><img src="https://opencollective.com/react-native-elements/backers.svg?width=890"></a>
-
+[![React Native Elements Backer](https://opencollective.com/react-native-elements/backer/0/avatar)](https://opencollective.com/react-native-elements/backer/0/website)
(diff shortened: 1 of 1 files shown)


dominikeinkemmer:
Seems like the build fails because of the `no-restricted-properties` es-lint rule. If you are interested in merging that change I think that eslint rule has to be removed or adjusted for Math.pow?

iRoachie:
> After doing some painfully long research I found that using ** instead of Math.pow can cause issues in that regard so I when I found that it is used in this project I changed this and my App ran smoothly again.

Where'd you see this issue? Math.pow should function the same as **

dominikeinkemmer:
@iRoachie It happens in my application, when I use the latest version of react-native-elements I always get the Unexpected token `*` error. Once i found the `**` usage in `src/tooltip/getTooltipCoordinate.js` and replaced it with Math.pow instead, the error is gone and things work again.

There is also this issue on the react-native page:
https://github.com/facebook/react-native/issues/11371

The solution proposed by some people to use the `babel-plugin-transform-exponentiation-operator` didn't work for me however.

alex4dev:
I have exactly the same problem since i have updated from 1.0.0beta7 to 1.1.0, replace ** by Math.pow resolve the problem too.
I've seen this issue on my old ipad with iOS9 and on my galaxy S6  on Android 7

Issue #1905, opened 2019-05-26
pricing missing style for button text


aliazizi:
**Is your feature request related to a problem? Please Describe.**
I want to set style for button text in pricing table

**Describe the solution you'd like**
In button props I want to set `titleStyle`

iRoachie:
Added in https://github.com/react-native-training/react-native-elements/pull/1906. Will be available in 1.2.0

PR #1906, opened 2019-05-26
Add title style to pricing table button

aliazizi:
Adding title style to pricing table button
I add an issue here https://github.com/react-native-training/react-native-elements/issues/1905


aliazizi:
I add it

Issue #1565, opened 2018-11-08
How to preserve ListItem styling when upgrading to v1?


lukewlms:
After upgrading to v1, ListItem styling has disappeared in our app.  It appears this is intended (the styling from `<List>` is removed).

To make the upgrade path much cleaner, can you give us a simple way to reproduce the style we used to get from List?  We don't want our app to change so it should be identical, but reverse engineering it will be pretty annoying (and would have to be repeated by anyone doing the upgrade).

Thanks!

lukewlms:
Solved by creating ListItem and ListContainer files to map to legacy ListItem and List, with styles.  Files below.

I would strongly suggest having a clear upgrade path that will allow users to retain their styles without needing to figure out how to rebuild them like this, if only a note under ListItem in the documentation!

The documentation also just appears incorrect - it doesn't list any styles in the code, but styles are applied in the screenshots which is why I spent 2-3 hours trying to figure out why I wasn't getting the default styles.

Thanks, still love the library.

ListContainer.tsx:
```
import * as React from "react";
import { StyleSheet, View } from "react-native";
import { legacyRNElementsColors } from "../Styles";

const styles = StyleSheet.create({
  listContainer: {
    marginTop: 20,
    borderTopWidth: 1,
    borderColor: legacyRNElementsColors.greyOutline,
    backgroundColor: legacyRNElementsColors.white,
  },
});

const ListContainer: React.SFC<{ children: React.ReactNode }> = ({
  children,
}) => <View style={styles.listContainer}>{children}</View>;

export default ListContainer;
```

ListItem.tsx:
```
import * as React from "react";
import { StyleSheet } from "react-native";
import { ListItem as RneListItem, ListItemProps } from "react-native-elements";
import {legacyRNElementsColors} from "../Styles";

const styles = StyleSheet.create({
  listItemContainer: {
    paddingTop: 10,
    paddingRight: 10,
    paddingBottom: 10,
    borderBottomColor: legacyRNElementsColors.greyOutline,
    borderBottomWidth: 1,
    backgroundColor: "transparent",
  },
});

const ListItem: React.SFC<ListItemProps> = props => (
  <RneListItem
    containerStyle={[styles.listItemContainer, props.containerStyle]}
    {...props}
  />
);

export default ListItem;
```

In Styles.ts:
```
export const legacyRNElementsColors = {
  primary: "#9E9E9E",
  primary1: "#4d86f7",
  primary2: "#6296f9",
[...]

PR #1915, opened 2019-06-06
Add information about styling ListItems

lukewlms:
This seems like an important change to me because the writing here doesn't match the images.  Adding the example code will _not_ produce the pictured results, because there is no styling in the example code.  The code in this comment does produce nice results.

See one frustrated user here: https://github.com/react-native-training/react-native-elements/issues/1565#issuecomment-499344338 - we also hit this issue and spent probably a couple hours trying to figure out why styling had died after an upgrade.

I believe this issue will be a significant hurdle for new users, and hurts the initial experience quite a lot.   It's just not possible to use ListItems without styling (they will look terrible) and this is a library that usually provides at least reasonable styling right away, so at minimum an example seems called for.  This PR is one way to do it, ofc more docs could be written as well.

Thanks for considering - we use this lib heavily and want it to be a success!

### DIFF
--- a/docs/listitem.md
+++ b/docs/listitem.md
@@ -7,6 +7,8 @@ ListItems are used to display rows of information, such as a contact list,
 playlist, or menu. They are very customizeable and can contain switches,
 avatars, badges, icons, and more.
 
+ListItems are not styled by default.  For an example of styling lists, [see this comment](https://github.com/react-native-training/react-native-elements/issues/1565#issuecomment-436982251).
+
 ![Lists](/react-native-elements/img/lists.png)
 
 ## Usage


iRoachie:
The issue you linked was referencing upgrading from 0.19.1 to v1.0.0 of the library, in which the styles the author previously added on the list item didn't have the same effect in v1.0.0.

The link to the comment you want to add to the docs, will not give the same styling as the images. The images only show what's possible with customization. 

I think we can add a different message and maybe put it under the Image on the ListItem page. https://react-native-training.github.io/react-native-elements/docs/listitem.html

What do you think?

PR #1920, opened 2019-06-15
Added replaceTheme function

kuket15:
to merge the theme passed in with the default theme.

The standard updateTheme function only allows to merge the passed in theme with the current theme.

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -230,7 +230,7 @@ const theme = {
 
 You may want to make use of the theming utilities in your own components. For
 this you can use the `withTheme` HOC exported from this library. It adds two
-props to the component it wraps - `theme` and `updateTheme`.
+props to the component it wraps - `theme`, `updateTheme` and `replaceTheme`.
 
 ```jsx
 import React from 'react';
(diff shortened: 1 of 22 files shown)


iRoachie:
What’s the benefit of adding this? Since it does a merge can’t you specify the properties you want to override?

kuket15:
In my app I need to completely replace the current theme with a new one.
The standard updateTheme does a merge, so it keeps the existing properties that I don't overwrite.

E.g.

```
var currentTheme = {
    Button: {
      containerStyle: {
        marginHorizontal: 20,
        marginVertical: 10,
      },
      titleStyle: {
        color: 'rgb(255,255,255)',
        fontFamily: 'regular',
        fontSize: 16,
      },
    },
}

var newTheme = {
    Button: {
      containerStyle: {
        marginHorizontal: 50,
        marginVertical: 20,
      },
    },
}
```

In  this case the updateTheme function would keep the titleStyle properties in the updated theme. I don't want to manually set the titleStyle properties to the default values for the Button component.

PR #1875, opened 2019-04-21
Tooltip: add new prop skipAndroidStatusBar for using inside RN Modal

oxyii:
Added new bool prop `skipAndroidStatusBar` with `false` as default value.

Related #1871

PR #1930, opened 2019-06-26
Fix Android Tooltip position

ddereszewski:
Issue is reproduced for example on: Android 9, RN 0.59.8, RNE 1.1.0

Displayed element wrapped in Tooltip is shown shifted up from the original element. See image below

### Current behavior:
65609483_354296048567220_5352678562589573120_n

### After fix:
65317028_443290659784941_2625493953058701312_n


oxyii:
@ddereszewski not for all android-based devices. #1875

flyingcircle:
Fixed in #1875 will be released with 2.0.0

PR #1932, opened 2019-06-27
Tooltip trigger

ddereszewski:
This Allows to trigger the tooltip either by onPress or onLongPress

### DIFF
--- a/src/tooltip/Tooltip.js
+++ b/src/tooltip/Tooltip.js
@@ -35,8 +35,15 @@ class Tooltip extends React.PureComponent {
 
   wrapWithPress = (toggleOnPress, children) => {
     if (toggleOnPress) {
+      const { trigger } = this.props;
+      const onPress = trigger === 'onPress' ? this.toggleTooltip : null;
+      const onLongPress = trigger === 'onLongPress' ? this.toggleTooltip : null;
       return (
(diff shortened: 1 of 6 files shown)


src/tooltip/Tooltip.js
```diff
@@ -158,9 +166,7 @@ class Tooltip extends React.PureComponent {
         ) => {
           this.setState({
             xOffset: pageOffsetX,
-            yOffset: isIOS
-              ? pageOffsetY
-              : pageOffsetY - StatusBar.currentHeight,
+            yOffset: pageOffsetY,
```
iRoachie: Why'd you change this here?

Slapbox:
@iRoachie, thanks for reviewing this!

If the author doesn't use Typescript it might not be obvious what action is needed. I also need this feature (`onLongPress`) and I'd help get this PR completed, but I don't use Typescript so I'm not sure exactly what's needed.

Are all the Typescript definitions in this file, and do we just need to add `onPress` and `onLongPress` as `void`s? https://github.com/react-native-training/react-native-elements/blob/master/src/index.d.ts#L1389-L1455

iRoachie:
The typescript definitions mirror how the actual code would work. In this case you need to add a trigger property which type is either onPress | onLongPress

flyingcircle:
Closing in preference of #2406

PR #1948, opened 2019-07-08
Updated type definitions

renjfk:
* Fixed TooltipProps.toggleOnPress property
* Added toggleTooltip method to Tooltip

### DIFF
--- a/docs/tooltip.md
+++ b/docs/tooltip.md
@@ -158,3 +158,33 @@ Flag to determine whether or not dislay pointer.
 |  Type   | Default |
 | :-----: | :-----: |
 | boolean |  true   |
+
+## Interaction methods
+
+| method         | description                                     |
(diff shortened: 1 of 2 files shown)


iRoachie:
@AndreiCalazans Is the `toggleTooltip` method supposed to be public?

renjfk:
It should be public in fact I'm actually surprised why it's not even documented. Imagine a scenario like this https://snack.expo.io/@renjfk/rne-tooltip.

iRoachie:
Can you add this usage to the docs as well then? Seems pretty convenient.

See the Interaction Methods section of the SearchBar for example https://react-native-training.github.io/react-native-elements/docs/searchbar.html

renjfk:
This is the best I could come with, feel free to correct.

PR #1969, opened 2019-07-18
fix(Text): Reorder style declarations for h1, h2, h3, and h4

SteveAquino:
Reorders the style declarations for headers so that the fontWeight property can be overridden on Android.

### DIFF
--- a/src/text/__tests__/Text.android.js
+++ b/src/text/__tests__/Text.android.js
@@ -1,0 +1,34 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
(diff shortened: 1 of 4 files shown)


SteveAquino:
CI is failing on this branch with some interesting snapshot differences that I can't reproduce locally.  I suspect it has something to do with the platform specific tests. I'm open to suggestions on a better testing approach since this only effects Android, otherwise we can just remove the test.

SteveAquino:
Nevermind, this is already fixed `next` branch

PR #1981, opened 2019-07-29
Swiper component

oxyii:
I rewrote [my swiper](https://github.com/oxyii/react-native-web-swiper) with RNE code styling. All props are rewritten too. But I need help with docs.

[Actual demo](https://github.com/oxyii/rne-swiper-demo)


oxyii:
Ready for review I hope )

oxyii:
> [Actual demo](https://github.com/oxyii/rne-swiper-demo)

Updated demo. Now it's most actual

oxyii:
@iRoachie What exactly prop(s) we can exclude? ))) And please specify which props need to be renamed?

docs/swiper.md
```diff
@@ -0,0 +1,500 @@
...
+import { Swiper } from 'react-native-elements'
+
+<Swiper>
+  <View style={{ flex: 1 }}>{/* any content */}</View>
+  <View style={{ flex: 1 }}>{/* any content */}</View>
+  {/* ... */}
+</Swiper>
+
+<Swiper
+  vertical {/* slide up / down instead left / right */}
```
iRoachie: I would change this to `direction`, enum being either `horizontal` | `vertical`

docs/swiper.md
```diff
@@ -0,0 +1,500 @@
...
+  from={1} {/* initial slide is second */}
```
iRoachie: I'd say just call this `initialSlide` pretty self-explanatory

docs/swiper.md
```diff
@@ -0,0 +1,500 @@
...
+  loop {/* back to first slide after last */}
+  timeout={2} {/* autoplay duration (2 secs) */}
+  springConfig={{ speed: 11 }} {/* RN Animated.spring config */}
+  minDistanceForAction={0.15} {/* Swipe less that 15% keep active slide */}
+  positionFixed {/* Fix mobile safari vertical bounces */}
```
iRoachie: [Just for example] - Leave prop comments outside of the example, there's already a reference section. Also, remove any props that aren't necessary for it to work really.

So:
```
  timeout={2} {/* autoplay duration (2 secs) */}
  springConfig={{ speed: 11 }} {/* RN Animated.spring config */}
  minDistanceForAction={0.15} {/* Swipe less that 15% keep active slide */}
  positionFixed
```

docs/swiper.md
```diff
@@ -0,0 +1,500 @@
...
+- [`gesturesEnabled`](#gesturesenabled)
+- [`springConfig`](#springconfig)
+- [`minDistanceToCapture`](#mindistancetocapture)
+- [`minDistanceForAction`](#mindistanceforaction)
+- [`positionFixed`](#positionfixed)
+- [`containerStyle`](#containerstyle)
+- [`innerContainerStyle`](#innercontainerstyle)
+- [`swipeAreaStyle`](#swipeareastyle)
+- [`slideWrapperStyle`](#slidewrapperstyle)
+- [`controlsEnabled`](#controlsenabled)
```
iRoachie: I'd say maybe just `showControls` boolean

docs/swiper.md
```diff
@@ -0,0 +1,500 @@
...
+- [`loop`](#loop)
+- [`timeout`](#timeout)
+- [`gesturesEnabled`](#gesturesenabled)
+- [`springConfig`](#springconfig)
+- [`minDistanceToCapture`](#mindistancetocapture)
+- [`minDistanceForAction`](#mindistanceforaction)
+- [`positionFixed`](#positionfixed)
+- [`containerStyle`](#containerstyle)
+- [`innerContainerStyle`](#innercontainerstyle)
+- [`swipeAreaStyle`](#swipeareastyle)
```
iRoachie: Is there any way possible that this could be simplified? 3 layers of wrappers is a lot

containerStyle -> innerContainerStyle -> swipeAreaStyle

Based on previous experience, users always get confused when they're multiple wrappers.

oxyii:
My idea was to provide the ability to fine-tune the component. Therefore, I did not particularly monitor the number of props. Well, the component is able to work correctly without any props. In any case, I am waiting for specific proposals for the modification of the component

docs/swiper.md
```diff
@@ -0,0 +1,500 @@
...
+
+Over the swiper we need to create a controls layer. But this layer will block the possibility of swiper layer control.
+We created 9 controls placeholders to solve this problem:
+`top-left`, `top`, `top-right`, `left`, `center`, `right`, `bottom-left`, `bottom` and `bottom-right`.
+You can adjust controls position by placing into relevant placeholder:
+
+```js
+<Swiper
+  ...
+  controlsProps={{
```
iRoachie: Something like dots and buttons would benefit more from a renderProps approach.

e.g

```jsx
renderDot={ ({index, active}) => <MyDot />}
```

This would remove need for all these props

```
- [`dotsTouchable`](#dotstouchable)
- [`dotsWrapperStyle`](#dotswrapperstyle)
- [`dotProps`](#dotprops)
- [`dotActiveStyle`](#dotactivestyle)
- [`DotComponent`](#dotcomponent)
```

docs/swiper.md
```diff
@@ -0,0 +1,500 @@
...
+
+## `controlsProps`
+
+|   Type   | Default |
+| :------: | :-----: |
+|  object  |  below  |
+
+---
+
+### `cellsStyle`
```
iRoachie: Not sure the purpose of these two props `cellStyle` and `cellsContent`

docs/swiper.md
```diff
@@ -0,0 +1,500 @@
...
+
+Custom next button component
+
+|   Type    |                               Default                              |
+| :-------: | :----------------------------------------------------------------: |
+| component | [Button](/react-native-elements/docs/button.html) |
+
+---
+
+### `firstPrevElement`
```
iRoachie: I'd say leave out this and `lastNextElement`.  It's very unlikely this will be used

iRoachie:
Thanks for the great effort put in here @oxyii.

I'm very cautious about complicated code. A good example of complicated code is the Slider component (which is barely never touched cause the code is hard to follow, and the original author isn't here).

Fine-tuning is good, but too much will overbear the user. Remember RNE is mostly picked up by persons also learning react-native at the same time. Method names like `_fixState` and `_fixAndGo` don't exactly explain what's happening.

The first question I wanted to know is how do I stop the over bounce animation. I'd say maybe disable this by default cause it's not expected by carousel libraries.

Currently, none of the maintainers we have now were there when rne started. So the simplicity of code is really important for going forward with the project.

docs/swiper.md
oxyii: Because of web. CSS's `position` is not equal to RN `position`. Besides this, we cannot know the styles of the root container. Eg [expo makes `#root` rubber](https://github.com/expo/expo/issues/5081) by default.

I have been looking for a solution for a long time but cannot exclude innerContainer.

docs/swiper.md
oxyii: This is [from community](https://github.com/oxyii/react-native-web-swiper/issues/1)

docs/swiper.md
oxyii: `cellsStyle` allows to change placeholder's styles, eg position. `cellsContent` allows to render additional content inside controls placeholders. e.g. rel [this issue](https://github.com/oxyii/react-native-web-swiper/issues/12#issuecomment-520168203)

docs/swiper.md
oxyii: Ohhh... [It was so](https://github.com/oxyii/react-native-web-swiper/tree/3517a7f69be3bce835e75876004b520efbbefb77#properties) in the initial version. Ok, will change back

oxyii:
@iRoachie 

> Method names like _fixState and _fixAndGo don't exactly explain what's happening.

`_fixState` - will fix slider position on resizes, touch releases in case with keep current slide, change screen orientation etc.. what is the correct name for it?

`_fixAndGo` - will call `_fixState` and change active slide. Please provide more clear name

Issue #1901, opened 2019-05-24
inputComponent: propType not allows functions wrapped in React.forwardRef 


avasuro:
**Explain what you did**
Passed function wrapped in React.forwardRef in inputComponent property

**Expected behaviour**
No warning shown

**Describe the bug**
If I pass function wrapped in React.forwardRef in inputComponent property - I got a warning _"Warning: Failed prop type: Invalid prop \`inputComponent\` supplied to \`Input\`."._

**To Reproduce**

Snack:
https://snack.expo.io/HJmRSmSa4

Steps:
1. Run code presented in snack on your device/emulator, because snack web don't display warnings (see this issue https://github.com/expo/snack-web/issues/21)
2. After view will be rendered you will see a warning _"Warning: Failed prop type: Invalid prop \`inputComponent\` supplied to \`Input\`."_

**Possible Solution**
add PropTypes.elementType to list of possible types that can be passed in inputComponent

**Screenshots**
![image](https://user-images.githubusercontent.com/11969473/58312826-c6b97780-7e14-11e9-8bdc-9b5fe367bf90.png)

**Your Environment (please complete the following information):**

| software              | version |
| --------------------- | ------- |
| react-native-elements |    1.1.0     |
| react-native          |    0.59.8     |
| npm           |    6.9.0     |

iRoachie:
I didn’t get any warnings when i ran your snack https://snack.expo.io/HJmRSmSa4. Is this still occurring for you?

avasuro:
@iRoachie Yes, it is still occurring. If you are run my snack via expo web - please run it on your device or emulator instead, because as I said:

> ...snack web don't display warnings (see this issue expo/snack-web#21)

iRoachie:
Merged 🚀 will be live in 1.2.0

PR #1990, opened 2019-08-01
fix(Input): allow to pass React.forwardRef as inputComponent (#1901)

avasuro:
Fix for https://github.com/react-native-training/react-native-elements/issues/1901


iRoachie:
Thanks 💯

PR #1995, opened 2019-08-02
fix(SearchBar): Show clear based on empty check

### LINKED ISSUE #1846: Android SearchBar clear button now visible when changing value prop
**Explain what you did**

I'm setting the value of the search bar after it is first displayed.

**Expected behaviour**

The clear button should appear if the value is not empty.

**Describe the bug**

The clear button does only appear if the user changes the text but not when changing the text programmatically.

**To Reproduce**

https://snack.expo.io/B1fr1vquN

**Possible Solution**

This seems related to #1247 and looks like an easy fix. Don't only set `isEmpty` in the constructor and onChange. Check it dynamically in render.

| react-native-elements |    1.1.0     |
| react-native          |     0.59.1    |
| npm or yarn           |     yarn 1.12.3    |

iRoachie:
Changes logic of showing clear button from using state to one that simply checks the value prop. This ensures that when the component's value is updated programmatically that the clear button will show correctly.

Fixes #1846

<p float="left">

</p>

### DIFF
--- a/src/searchbar/SearchBar-android.js
+++ b/src/searchbar/SearchBar-android.js
@@ -63,19 +63,20 @@ class SearchBar extends Component {
 
   onChangeText = text => {
     this.props.onChangeText(text);
-    this.setState({ isEmpty: text === '' });
   };
 
   constructor(props) {
     super(props);
-    const { value } = props;
+
     this.state = {
       hasFocus: false,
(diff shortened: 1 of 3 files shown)


iRoachie:
Closing as this isn't the correct behaviour natively

PR #2008, opened 2019-08-09
Added React.ComponentElement to input label type

paulmathis:
Added React.ComponentElement to match the accepted input types as stated in the [docs](https://react-native-training.github.io/react-native-elements/docs/input.html#label).

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -873,7 +873,7 @@ export interface InputProps extends TextInputProperties {
   /**
    * 	Adds label (optional)
    */
-  label?: string  | React.ComponentElement;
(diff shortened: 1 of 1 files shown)


src/index.d.ts
```diff
@@ -873,7 +873,7 @@ export interface InputProps extends TextInputProperties {
   /**
    * 	Adds label (optional)
    */
-  label?: string;
+  label?: string  | React.ComponentElement;
```
iRoachie: Use this React.ReactElement instead. See https://github.com/react-native-training/react-native-elements/blob/e7ce592fa83a614b7cd1cce6f8a4e0795136775c/src/index.d.ts#L1100

Issue #2004, opened 2019-08-06
Avatar - Add option to disable fade animation


danwhite-ad:
Require the ability to disable fade between icon and source. I am setting the avatar source within componentDidMount (fetching from cache), but I am experiencing a fade between the set icon and the fetched source which is a jarring user experience.

iRoachie:
Fade in is the expected behaviour which is the direct opposite of jarring user experience. This can't be disabled at the moment. The avatar uses the RNE Image under the hood.

We may consider adding a prop to disable it if any more request come in

danwhite-ad:
The issue I am having is that I have an image in cache, but because of the fade it is making it appear that the image is being remotely fetched every time the component is loaded.

iRoachie:
Gotcha! That makes sense. It wouldn't be able to do in this next release though

Prateek93a:
Can I work on this one?

iRoachie:
Sure go ahead. This will need to be a prop on the Image component

Prateek93a:
Ok

Prateek93a:
Sir, thanks for providing me the chance to make my first coding open source contribution. Sir, according to my solution, there is a AllowFadeIn(a boolean) prop on Avatar component, and according to its value, fadeDuration prop of Image component imported from image folder, will be set to 0(if true) or 300(default value if false). Is it fine, or can we do something better?

Prateek93a:
Also kindly explain me why we are using onLoadEnd in the Image component

iRoachie:
The Image component has 2 parts, the underlying image and a placeholder view above it. The `onLoadEnd` function is called when the Image loads then we gradually fade the above placeholder.

PR #2012, opened 2019-08-13
Added AllowFadeIn prop as requested in #2004 

Prateek93a:
It is a boolean prop, which will make fadeDuration of Image prop=0 if true and 300(default value) if false. This feature will cater to the need mentioned in #2004. I hope, I made something useful.


Prateek93a:
Can someone guide me why the test is failing. I do not have experience with Travis CI tool.

iRoachie:
This doesn't actually change anything.

Prateek93a:
Extremely sorry for the last PR, that was a silly mistake

Issue #1717, opened 2019-01-15
ListItem Inset borders


SDxBacon:
Hi,

Currently, I am trying react-native-element's example code that shared by Monte9 on Github, [here](https://github.com/Monte9/react-native-elements-app). I am wondering is it possible to remove bottom border under leftIcon in ListItem without using SectionList? For example: 
![1fngiy8](https://user-images.githubusercontent.com/10906745/51187839-a97d1580-1917-11e9-870c-0f8f13229dcb.png)

[Monte9's example code](https://github.com/Monte9/react-native-elements-app/blob/master/src/views/settings.js#L145) using SectionList's props `ItemSeparatorComponent` to reach the goal. Can I reach the same effect without using SectionList but just configuring the ListItem's props? 

Thanks!

iRoachie:
There's no way to do this right now just with the props of the ListItem. The bottomBorder is actually set on the container and isn't a separate component that an inset (margin left) can be configured. However this maybe a cool feature to have. cc @martinezguillaume

flyingcircle:
Closing due to inactivity.

Brightness99:
We can try this.
`
     <ListItem
       ...
        Component={(props) =>
          <View>
            {props.children}
            <Divider style={{backgroundColor: '#aaa', marginLeft: 65 }}/>
          </View>
        }
      />
`

PR #2019, opened 2019-08-14
 #1717 react-native-training#1717 Added dividerWithMargin feature 

Prateek93a:
As per mentioned in the issue #1717, I added the feature for implementing borderBottom with nothing below the leftAvatars. I hope this would help

### DIFF
--- a/src/avatar/Avatar.js
+++ b/src/avatar/Avatar.js
@@ -53,7 +53,6 @@ const Avatar = ({
   placeholderStyle,
   renderPlaceholderContent,
   ImageComponent,
-  AllowFadeIn,
   ...attributes
 }) => {
   const width =
@@ -138,7 +137,6 @@ const Avatar = ({
           overlayContainerStyle,
           rounded && { borderRadius: width / 2, overflow: 'hidden' },
         ])}
-        fadeDuration={AllowFadeIn === true ? 300 : 0}
         source={source}
         {...imageProps}
         style={StyleSheet.flatten([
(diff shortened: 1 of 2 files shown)


Prateek93a:
Sorry for making so many PRs, but finally things worked out.
This dividerWithMargin feature will provider the border asked without using sectionlist.
It took a great deal of time to get it accepted by travis ci.
Hope this feature helps

iRoachie:
Thanks for making at attempt at this feature. However, this isn't as simple as you think. The ListItem would need to make much more significant changes to be able to support this properly.

See case with ListItem where the avatar size is large:

```jsx
<ListItem
  leftAvatar={{ source: { uri }, size: 'large' }}
  title="Nah girl"
  dividerWithMargin
/>
```

Simulator Screen Shot - iPhone 7 - 2019-08-14 at 21 24 45

Prateek93a:
I see, so this divider feature should cater to all the icon sizes and have different margins accordingly.
I will make the necessary changes.

Prateek93a:
Finally pulled it off!!
Finally got through Travis ci. Now I have implemented dividerWithMarginLeft and dividerWithMarginRight, to handle both left and right avatars of all the sizes small, medium, large and xlarge with default value corresponding to small. I hope this time it works for all test cases.

flyingcircle:
there has been recognition lately that we want to limit the number of props added to components, and this seems like it would contribute to "prop bloat". If there's another way to implement this without adding these props, you can open up a new PR

PR #2037, opened 2019-08-23
Update react-native-vector-icons linking

v1b3m:
For react-native versions from 0.6.0 and above, linking is done automatically and shouldn't be done manually.


docs/getting_started.md
```diff
@@ -73,6 +73,8 @@ npm i --save react-native-vector-icons
 react-native link react-native-vector-icons
 ```
 
+_Manual linking of react-native-vector-icons is not necessary if you're using react-native@0.6.0 or above since it is done automatically. This will throw an error though it won't prevent the application from running. To fix this you'll simply have to run `react-native unlink react-native-vector-icons` and the process will run as expected._
```
iRoachie: Should be 0.60

Issue #1784, opened 2019-02-17
ButtonGroup - containerBorderRadius doesn't work


trane294:
Version 1.0.0

Works:

```
<ButtonGroup
	containerStyle={{borderRadius: 0}}
</ButtonGroup>
```
Doesn't work even mentioned in documentation

```
<ButtonGroup
	containerBorderRadius={0}
</ButtonGroup>
```

iRoachie:
Correct. I think we can remove this prop altogether but this would be a breaking change. Let's put this back for a bit.

trane294:
Please just remove it from documentation. Thank you

PR #2068, opened 2019-09-16
Remove containerBorderRadius props

Naturalclar:
This PR addresses #1784 

I've removed containerBorderRadius from ButtonGroup docs,
and removed the containerBorderRadius from ButtonGroup component.

### DIFF
--- a/src/buttons/ButtonGroup.js
+++ b/src/buttons/ButtonGroup.js
@@ -45,6 +45,7 @@ const ButtonGroup = props => {
   } = rest;
 
   let innerBorderWidth = 1;
+  const defaultBorderRadius = 3;
 
   if (
     innerBorderStyle &&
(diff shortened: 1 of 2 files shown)


src/buttons/ButtonGroup.js
```diff
@@ -89,12 +89,12 @@ const ButtonGroup = props => {
               },
               i === buttons.length - 1 && {
                 ...lastBorderStyle,
-                borderTopRightRadius: containerBorderRadius,
-                borderBottomRightRadius: containerBorderRadius,
+                borderBottomRightRadius: 3,
```
iRoachie: Maybe we can have this in a variable, instead of just having 3's everywhere

Naturalclar: Thanks! I've set `defaultBorderRadius` variable as 3 👌

Issue #2091, opened 2019-09-26
testId no longer working on Image Tag.


TomJKlug:
> Before filling out an issue, be sure to check the version of react-native-elements you are using and the docs for that version over [here](https://react-native-training.github.io/react-native-elements/versions.html).

Version 1.2.0 and 1.2.1

**Explain what you did**
TestId on Image is no longer being honored. It is putting in the static TestId. 

**Expected behaviour**
When I put a testId on the image I expect it to be applied to the component. 

**Describe the bug**
<Image testId='testId'/> no longer gets the passed testId.

**To Reproduce**
We highly recommend that you re-create the bug on [Snack](https://snack.expo.io). If not, list the steps that a reviewer can take to reproduce the behaviour:

**Possible Solution**
Can you suggest a possible fix that would address this issue?

It should take the testId prop and apply it to the imagecomponent if present instead of just using a static string. See below. This is the image.js.
```
render() {
    const {
      placeholderStyle,
      PlaceholderContent,
      containerStyle,
      style,
      ImageComponent,
      children,
      ...attributes
    } = this.props;
    const hasImage = Boolean(attributes.source);

    return (
      <View
        accessibilityIgnoresInvertColors={true}
        style={StyleSheet.flatten([styles.container, containerStyle])}
      >
        <ImageComponent
          {...attributes}
          onLoad={this.onLoad}
          style={[
            StyleSheet.absoluteFill,
            {
              width: style.width,
              height: style.height,
            },
          ]}
          testID="RNE__Image"  /* THIS IS WHERE*/
        />
```

iRoachie:
Ah sorry about that. Things changed slightly when the component was refactored in https://github.com/react-native-training/react-native-elements/commit/e2795821fa85bbce48282b9c92d2260658d5c35e#diff-db91516ae3fd33db904a11400a5e19e4

iRoachie:
Can you make a PR to fix this? Base it from the `patch` branch

TomJKlug:
#2092

TomJKlug:
Sorry at work, so just a quick one. If I missed something on the contributing guide let me know. Added the testID, and added two tests to ensure it works.

iRoachie:
Thanks @TomJKlug  💯. There seems to be a prob with npm atm. So will publish as soon as it's back up

PR #2092, opened 2019-09-26
#2091 Honor testID if passed.

TomJKlug:
Honors the Test Id if passed on the image component.

### DIFF
--- a/src/image/Image.js
+++ b/src/image/Image.js
@@ -38,6 +38,7 @@ class Image extends React.Component {
       PlaceholderContent,
       containerStyle,
       style,
+      testID,
       ImageComponent,
       children,
       ...attributes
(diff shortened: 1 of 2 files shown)


iRoachie:
Instead just move the line https://github.com/react-native-training/react-native-elements/blob/next/src/image/Image.js#L62 above `{...attributes}`

TomJKlug:
Doing that the test does not pass? Am I missing something?

iRoachie:
The test passes, i'm just suggesting a better way to fix the issue

TomJKlug:
I'm sorry. I am not saying that my current implementation passes. I am saying I applied your fix and the tests no longer pass. 
```
 <ImageComponent
          testID="RNE__Image"
          {...attributes}
          onLoad={this.onLoad}
          style={[
            StyleSheet.absoluteFill,
            {
              width: style.width,
              height: style.height,
            },
          ]}
        />
```

This does not pass.

Issue #1938, opened 2019-07-03
Header crashes on Android. 


emomooney:
**Explain what you did**
Use header component. Used most basic example. Experiencing instant force close on Android, no error. Just closes. Works on ios. 

**Expected behaviour**
That I can use the example header component. 
`<Header
  leftComponent={{ icon: 'menu', color: '#fff' }}
  centerComponent={{ text: 'MY TITLE', style: { color: '#fff' } }}
  rightComponent={{ icon: 'home', color: '#fff' }}
/>`

**Describe the bug**
App crashes instantly. 
I've also attempted to use an example I found in a previous issue from here. 
`<Header
          innerContainerStyles={{flexDirection: 'row'}}
          backgroundColor='#f40057'
        >
          <Avatar
            small
            rounded
            source={{uri: "http://www.usanetwork.com/sites/usanetwork/files/styles/629x720/public/2016/07/mrrobot_s2_cast_rami-malek2.jpg"}}
            onPress={() => this.goToProfileDetail()}
            activeOpacity={0.7}
          />
          <View style={{alignItems: 'center'}}>
            <Text style={{justifyContent: 'center', color: '#fff', alignContent:'center'}}>Eliot Alderson</Text>
            <Text style={{justifyContent: 'center', color: '#fff'}}>eliot@food.com.br</Text>
          </View>
          <Icon
            name={'settings-applications'}
            color={'#fff'}
            onPress={() => this.goToProfileDetail()}
          />
        </Header>`

Results in the same crash. If I take the header component out everything runs properly. Relatively fresh project build. I have the following packages : 
`"react": "16.8.3",
    "react-native": "0.59.9",
    "react-native-clean-project": "^3.2.1",
    "react-native-elements": "^1.1.0",
    "react-native-firebase": "^5.5.4",
    "react-native-gesture-handler": "^1.3.0",
    "react-native-vector-icons": "^6.6.0",
    "react-navigation": "^3.11.0"`

**Your Environment (please complete the following information):**

| software              | version |
| --------------------- | ------- |
| react-native-elements |    1.1.0     |
| react-native          |    0.59.9     |
| npm or yarn           |     6.9.0    |

iRoachie:
Could not replicate this. See https://snack.expo.io/@roach_iam/authentic-soda. Do you have a repo or example you can provide that shows this bug?

devangjob:
> **Explain what you did**
[...]

PR #2095, opened 2019-09-29
Use ImageBackground only when backgroundImage is provided for Header

RobertSasak:
Otherwise use simple View. This commit is to address an issue on old Android phones. BackgroundImage without source will crash on Android.
Discussion and workaround can be found here https://github.com/react-native-training/react-native-elements/issues/1938

### DIFF
--- a/src/header/Header.js
+++ b/src/header/Header.js
@@ -77,7 +77,9 @@ class Header extends Component {
       linearGradientProps,
       ViewComponent = linearGradientProps && global.Expo
         ? global.Expo.LinearGradient
-        : ImageBackground,
+        : backgroundImage
+        ? ImageBackground
+        : View,
       theme,
       ...attributes
     } = this.props;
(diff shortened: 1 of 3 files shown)


RobertSasak:
I submitted it as an issue to react-native as well. https://github.com/facebook/react-native/issues/26637

iRoachie:
I really think we should get this fix this upstream, and not here. We'd have to change every component that uses ImageBackground. I'll see if I can message someone from the rn team on twitter

RobertSasak:
Here is a react native source code for ImageBackground https://github.com/facebook/react-native/blob/master/Libraries/Image/ImageBackground.js
It consist of Image inside of View. 
```jsx
<View  ...>
  <Image .../>
</View>
```
React native made a clear decision when they separated View and ImageBackground into two separated component. It is due performance. I understand that this is not a big deal in a header which is only once per screen, but it might have bigger impact in other components.

In my opinion is a bad coding practice to use ImageBackground instead of View in addition to performance loss. It would be cool to hear opinion of internal react native developer. Let me know if you get somebody to comment on it.

PR #2114, opened 2019-10-06
feat: Add disabled style for title, subtitle, left and right icon

### LINKED ISSUE #2111: ListItem: Add a "disabled" prop for every style prop

Currently there's no way to style a title, container etc.. according to `disabled` status unless we use the disabled status itself on the styling, eg:

```jsx
              <ListItem
                leftIcon={{ name: 'autorenew' }}
                titleStyle={{ color: resetModuleEnabled ? 'white' : 'grey' }}
                title={translate('screens.settings.reset')}
                disabled={!resetModuleEnabled}
                onPress={sendReset}
              />
```

Here I'm changing `titleStyle` according to `resetModuleEnabled`; but I'd rather there be some way to pass the disabled title style similar to how `disabledStyle` works for the listitem itself.

While the current solution *works*, it means I can't use `ThemeProvider` to globally set the disabled text style.

I'd like to pass an alternative set of styles according to disabled status for *every* part of the component. either this:

```jsx
<ListItem {...} titleStyle={{color: 'white', disabledStyle: {color: 'grey'} }} />
```

or this:

```jsx
<ListItem ... titleStyle={{color:'white'}} disabledStyles={{title: {color: 'grey'} }} />
```

Note that the current `disabledStyle` works only for the underlying view component, but I'd like to have a way to style every sub-part of a ListItem(ie. the container, the right/left icons, title & subtitle etc), and so I'd be able to use `ThemeProvider` to determine how a disabled ListItem should behave

right now I'm passing the disabled property to determine the styling, but IMO that's not ideal.

haruelrovix:
Add 'disabledStyle' for sub-part of a ListItem. Close #2111.


haruelrovix:
@iRoachie @barbiero Am I on the right track?

Use it like so:

```
const theme = {
  ListItem: {
    disabledTitleStyle: {
      color: 'grey'
    },
    disabledSubtitleStyle: {
      color: 'grey',
      fontStyle: 'italic'
    },
    disabledLeftIconStyle: {
      containerStyle: {
        backgroundColor: 'grey'
      }
    }
  }
};
```

in [React Native Elements App](https://github.com/react-native-elements/react-native-elements-app)'s [settings](https://github.com/react-native-elements/react-native-elements-app/blob/master/src/views/settings.js). Result:



In screenshot above, `Wi-Fi` and `Personal Hotspot` are disabled.

haruelrovix:
@Barbiero @iRoachie 

> ... but I'd like to have a way to style every sub-part of a ListItem

From my list, here is the **sub-part of a ListItem**:

- [ ] leftElement
- [x] leftIcon
- [ ] leftAvatar
- [ ] contentContainer
- [x] title
- [x] subtitle
- [ ] rightContentContainer
- [ ] rightTitle
- [ ] rightSubtitle
- [ ] input
- [ ] switchProps
- [ ] checkBox
- [ ] badge
- [ ] buttonGroup
- [x] rightIcon
- [ ] rightAvatar
- [ ] rightElement
- [ ] checkmark
- [ ] chevron

It's gonna be big if we include all of them in one PR. I think I'll stop this PR here, only for **four** components.

Also @ganes1410 seems interested to the related issue as well. Feel free to work for the other parts 🤝

iRoachie:
We should think about this more. The sub elements of the ListItem already each have a disabled prop and associated styles. So if we can already set these on the ThemeProvider and just pass the disabled prop onto these elements

iRoachie:
I don’t really like the idea of all these props being added, this is becoming to look like the ListItem from before v1. Maybe we can rethink the api a bit. 

What about if each prop took in a function that then returned an object?

```jsx
<ListItem titleStyle={(disabled) => { color: disabled ? ‘red’ : ‘blue’}}
```

Thought?

haruelrovix:
@Barbiero @iRoachie please check this commit, [8215ffa](https://github.com/react-native-elements/react-native-elements/pull/2114/commits/8215ffa641dbad2fdf7dbda1a1e9e01b2ffc203d).

Do we need to change `containerStyle` [L115](https://github.com/react-native-elements/react-native-elements/blob/3e567deb1083e41b81209e8034540383f8951691/src/list/ListItem.js#L115) to make it consistent with the new approach?

iRoachie:
Yea @haruelrovix let's do it. Let's roll it out for this component only for 2.0. After we get feedback from users, we'll know how to move ahead with the other components. But I personally believe this is the best way to go forward

iRoachie:
Also maybe we should make this an object too.

```jsx
<ListItem titleStyle={({ disabled }) => { color: disabled ? ‘red’ : ‘blue’}} />
```

That way for other components, we can support things like `checked` etc

PR #2117, opened 2019-10-07
Update listitem.md

ridvanaltun:
Example old, below issue fixed.

`LinearGradient 'end' property should be an object with fields 'x' and 'y', Array type is deprecated.`


iRoachie:
Seems like both versions work. https://snack.expo.io/@roach_iam/quiet-graham-crackers

I think expo has one set of instructions that are different from the ones on react-native-linear-gradient.

Expo - https://docs.expo.io/versions/latest/sdk/linear-gradient/#props
react-native-linear-gradient - https://github.com/react-native-community/react-native-linear-gradient#additional-props

Should we change?

PR #2119, opened 2019-10-07
chore: update and move types for react-native-vector-icons

Naturalclar:
updated `@types/react-native-vector-icons` to current latest, which fixes typings for `Icon` component, and moved it from `dependencies` to `devDependencies`.

I believe `@types` packages should be in `devDependencies` as having them in `dependencies` will cause typings to lock-in with the library.


iRoachie:
It's supposed to be a dependency. It's because cause we are shipping their types as apart of our types.

As a dev dependency If you try to use one of our components that extends from Icon, the typings will error as they don't exist as a dependency. See the snapshot below using your branch.

Screenshot 2019-10-07 at 10 01 42 AM

iRoachie:
Might I ask what problems you were having with Icon that you needed to update the dependency? Might've been easier if you made an issue first to discuss what was the error.

Naturalclar:
@iRoachie Thanks for the response! I realized after sending a PR that `@types/react-native-vector-icons` should be in `dependencies` so that people wont need to install `@types/react-native-vector-icons` while using `react-native-elements`. I've made a new issue explaining the problem I was having and possible solution in #2120 :+1:

iRoachie:
Thanks 🎉

iRoachie:
Live in [v1.2.6](https://github.com/react-native-elements/react-native-elements/releases/tag/v1.2.6)

PR #2124, opened 2019-10-10
feat(Icon): Use platform Touchable

### LINKED ISSUE #2027: Wrap type=Android SearchBar in TouchableNativeFeedback
Pressing the buttons in SearchBar on Android with platform=Android is not showing ripple on buttons, it is doing opacity. 

If Android, and platform=Android on Searchbar, we should wrap the buttons in TouchableNativeFeedback.

react-native-paper does it well - https://callstack.github.io/react-native-paper/searchbar.html

However I prefer react-native-elements, because of the loading indicator, which paper lacks.

**Additional context**

iRoachie:
Allows using TouchableNativeFeedback on android, and TouchableHighlight on web and iOS.

None of the breaking changes below break your app, but the look might change depending on the props used.

BREAKING CHANGES:
- No longer uses TouchableHighlight on android by default.
- Default underlay color is now transparent instead of white
- activeOpacity previously not set is now 0.3

Resolves #2027

### DIFF
--- a/src/icons/Icon.js
+++ b/src/icons/Icon.js
@@ -6,7 +6,9 @@ import {
   View,
   StyleSheet,
   Text as NativeText,
+  TouchableNativeFeedback,
 } from 'react-native';
+import Color from 'color';
 
 import getIconType from '../helpers/getIconType';
 import { ViewPropTypes, withTheme } from '../config';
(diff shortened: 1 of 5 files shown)


iRoachie:
Reason why https://github.com/react-native-elements/react-native-elements/issues/2027#issuecomment-538776577 didn't work is because TouchableNativeFeedback needs a child View element in which to calculate the ripple. It's because of this that we had to move around the styles a bit.

ganes1410:
Hey @iRoachie. I can confirm this works after testing with `Pixel 2 API_28` emulator as mentioned in this issue #2027.

PR #2135, opened 2019-10-14
Change TouchableHighlight to TouchableOpacity

filipkastovsky:
TouchableHighlight is overly aggressive, causes entire element to turn black, TouchableOpacity would be preferable


iRoachie:
TouchableHighlight is the default touchable for iOS for List Items. Not changing this. If you want to change it in your app you can by setting the Component prop.

PR #2137, opened 2019-10-15
Few fixes and Improvements for Button

iRoachie:
Some nice improvements for button, mainly on android. Other non-visible changes are improvements to accessibility and a ripple fix for border-radius

### Ripple Color
The Ripple color now correctly matches the background and foreground text like it should.

**Solid (current)**
solid-current

**Solid (after)**
solid-after

**Clear (current)**
clear-current

**Clear (after)** 
clear-after

### DIFF
--- a/src/buttons/Button.js
+++ b/src/buttons/Button.js
@@ -9,6 +9,7 @@ import {
   Platform,
   StyleSheet,
 } from 'react-native';
+import Color from 'color';
 
 import { withTheme, ViewPropTypes } from '../config';
 import { renderNode, nodeType, conditionalStyle, color } from '../helpers';
@@ -41,7 +42,7 @@ class Button extends Component {
       loadingProps: passedLoadingProps,
       title,
       titleProps,
-      titleStyle,
+      titleStyle: passedTitleStyle,
       icon,
       iconContainerStyle,
       iconRight,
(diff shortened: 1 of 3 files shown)


Monte9:
Solid! Thanks for continually improving the library and making the elements even more polished! 💯 
Something I was curious about is measuring impact of changes like this? Button is one of the most commonly used RNE component.

Is there a way for us to track how many installs we have on the lastest version & what components are being used in the apps? Knowing the usage of each component will also help us prioritize work in the future.

iRoachie:
Only thing I can think of is the used by statistic from github Screenshot 2019-10-15 at 3 32 26 PM

https://github.com/react-native-elements/react-native-elements/network/dependents

iRoachie:
Also the changes can be undone if the user wants to preserve the old behaviour. The release notes for 2.0.0, outline breaking changes as well as what to change if a user wants to revert that behaviour

Monte9:
Yeah, that github "used by" def gives us a based metric. Was wondering if there could be a Google Analytics like tool that sends metrics back each time a user import Button from RNE in their project etc.

Also these changes looks like improvements, so not too concerned about users wanting to revert back to older version. But release notes helps with that in case they want to. 👍

Monte9:
:shipit:

PR #2144, opened 2019-10-20
Check whether loading is false before call onPress

imcvampire:
Implement for https://github.com/react-native-elements/react-native-elements/issues/2141

### DIFF
--- a/src/buttons/Button.js
+++ b/src/buttons/Button.js
@@ -42,6 +42,7 @@ class Button extends Component {
     const {
       TouchableComponent,
       containerStyle,
+      onPress,
       buttonStyle,
       type,
       loading,
--- a/src/buttons/Button.js
+++ b/src/buttons/Button.js
@@ -29,11 +29,18 @@ class Button extends Component {
     }
   }
 
+  handleOnPress = () => {
+    const { loading, onPress } = this.props;
+
+    if (!loading) {
+      onPress();
+    }
+  };
+
   render() {
     const {
       TouchableComponent,
       containerStyle,
-      onPress,
       buttonStyle,
       type,
       loading,
@@ -89,7 +96,7 @@ class Button extends Component {
         ])}
       >
         <TouchableComponent
-          onPress={onPress}
+          onPress={this.handleOnPress}
           activeOpacity={0.3}
           disabled={disabled}
           {...attributes}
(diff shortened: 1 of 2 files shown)


imcvampire:
> Going to need add a test for this.
> 
> Basically the test should shallow render the button and pass in a stub onPress. After clicking on the button, assert that the stub hasn't been called

Added!

src/buttons/__tests__/Button.js
```diff
@@ -43,6 +43,19 @@ describe('Button Component', () => {
     expect(onPress).toHaveBeenCalled();
   });
 
+  it('should not be call onPress events when loading is true', () => {
+    const onPress = jest.fn();
+    console.log = jest.fn();
+    const wrapper = shallow(<Button theme={theme} />);
+
+    wrapper.setProps({ onPress, loading: true });
```
iRoachie: No need to put this here just set it on the button directly. Also, remove the console.log mock

src/buttons/__tests__/Button.js
```diff
@@ -43,6 +43,18 @@ describe('Button Component', () => {
     expect(onPress).toHaveBeenCalled();
   });
 
+  it('should not be call onPress events when loading is true', () => {
+    const onPress = jest.fn();
+    const wrapper = shallow(<Button theme={theme} loading={true} />);
+
+    wrapper.setProps({ onPress });
```
iRoachie: The onPress as well. Just it on the button.

The only time you'll need to use setProps is when you're testing before and after states.

src/buttons/__tests__/Button.js
```diff
@@ -43,6 +43,18 @@ describe('Button Component', () => {
     expect(onPress).toHaveBeenCalled();
   });
 
+  it('should not be call onPress events when loading is true', () => {
+    const onPress = jest.fn();
+    const wrapper = shallow(<Button theme={theme} loading={true} />);
```
iRoachie:
```suggestion
    const wrapper = shallow(<Button theme={theme} loading />);
```

src/buttons/Button.js
```diff
@@ -30,11 +30,18 @@ class Button extends Component {
...
+    if (!loading) {
+      onPress();
+    }
+  };
+
   render() {
     const {
       TouchableComponent,
       containerStyle,
-      onPress,
```
iRoachie: This won't work. Removing onPress from here means it'll be in `...attributes`. Which means your onPress handler will be overridden and `handleOnPress` will never run

imcvampire: I have corrected it

iRoachie:
Thanks again 🚀

PR #2147, opened 2019-10-23
fixed unnecessary ThemProps in withTheme hoc

roadmanfong:
when I use withTheme

```js
import React from 'react'
import {ThemeProps, withTheme } from 'react-native-elements'

interface Props extends ThemeProps<any>{
   onPress: () => void
}

class Button extends React.Component {
   // ... rest of all
}

const ThemeButton = withTheme(Button)
```

```js
<ThemeButton 
  onPress={this.onPress}
/>
```

keep showing error about not providing required props `theme` `updateTheme` `replaceTheme`

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -2070,4 +2070,4 @@ export const ThemeContext: React.Context<ThemeProps<{}>>;
 
 export function withTheme<P = {}, T = {}>(
   component: React.ComponentType<P & ThemeProps<T>>
-): React.ComponentClass<P>;
+): React.ComponentClass<Omit<P, keyof ThemeProps<T>>>;


iRoachie:
What are you trying to do? I'm not understanding what bug you're having

roadmanfong:
I think `withTheme` is an injector according this article https://medium.com/@jrwebdev/react-higher-order-component-patterns-in-typescript-42278f7590fb

```js
const ThemedComponent = withTheme(Component)
```

`theme` `updateTheme` `replaceTheme`  are already injected by `withTheme` should not be a required props of `ThemedComponent`, but currently the definition treat these as required props

iRoachie:
That's not true. withTheme works as expected.

```tsx
interface MyTextProps {
  name: string;
}

const MyText = withTheme<MyTextProps>(({ theme, name }) => (
  <Text style={{ color: theme.colors.primary }}>{name}</Text>
));

const Main = () => {
  return (
      <MyText name="Mary" />
    </View>
  );
};
```

Example without component props:

```tsx
const MyText = withTheme(({ theme }) => (
  <Text style={{ color: theme.colors.primary }}>Mary</Text>
));

const Main = () => {
  return (
      <MyText />
    </View>
  );
};
```

roadmanfong:
Yes it works well in function component, but class component shows error

```js

interface Props extends ThemeProps<any> {
  name: string
}

class MyText2 extends Component<Props> {
  public render() {
    const {
      name
      theme,
    } = this.props
    return (
      <Text style={{ color: theme.colors.primary }}>{name}</Text>
    )
  }
}

const ThemedMyText2 = withTheme(MyText2)

const Main2 = () => {
  return (
      <ThemedMyText2 name="Mary" />
    </View>
  );
}

```

iRoachie:
Ahh okay, it'll fail because you don't use the generic on `withTheme`. But I can still accept this 👍.

iRoachie:
Thanks again 💯

iRoachie:
Live in [1.2.7](https://github.com/react-native-elements/react-native-elements/releases/tag/v1.2.7)

roadmanfong:
Thanks

PR #2152, opened 2019-10-24
chore: 🤖 git-cz integration

oxyii:
I like that RNE repo uses `husky`. Also I noticed that someone is trying to create semantic messages of git commits.
I would like to suggest integrating an interactive assistant [`git-cz`](https://github.com/streamich/git-cz) into RNE repo.
Initially, `git-cz` requires a global installation or `commitizen` installed globally. But after [little changes](https://github.com/streamich/git-cz/issues/79) this lib can work fine as `husky` hook executor.


package.json
```diff
@@ -113,7 +113,8 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged"
+      "pre-commit": "lint-staged",
+      "prepare-commit-msg": "exec < /dev/tty && node ./.git-cz/script.js || true"
```
oxyii: I'm not sure about windows users )) but linux and macOS ready

iRoachie:
How does this work with GUI based git programs? Say like committing via VScode Git panel?

oxyii:
@iRoachie `git-cz` is just a helper in creating a commit message. So `prepare-commit-msg` hook should return zero code anyway. That's why I wrote `... || true` in `package.json`. Without `... || true` commit failed with error `/bin/sh: /dev/tty: Device not configured` in WebStorm git GUI.

If hook returns non-zero exit-code then pre-passed commit message will pass to commit. Even in terminal, if you wrote `git commit -m "some prepared message"` but then you interrupt (`ctrl+c`) inside `git-cz` interactive shell, your commit will pass with message "some prepared message"

iRoachie:
By looks of the config and script, this looks like something we'll have to maintain over time. Maybe we could look into [commitlint](https://github.com/conventional-changelog/commitlint?) instead.

oxyii:
@iRoachie maintain? What exactly? Config looks simple. One-time configure `list`, `questions` and `types`... Then only manage `scopes` related to RNE components. 
`commitlint` is something else... message checker... this PR about interactive helper.
Okay, let close this PR if there are no other opinions

iRoachie:
I'm referring to the 42k file .git-cz/script.js

oxyii:
@iRoachie It's just build of `git-cz` with [this changes](https://github.com/streamich/git-cz/issues/79)

iRoachie:
Really not a fan of having this in source control. Especially at that size. If you can find a npm package that abstracts having to have this file then sure.

oxyii:
it remains to configure

oxyii:
git-cz now [supports](https://github.com/streamich/git-cz/pull/106) husky. I will work on this PR in few days

PR #2160, opened 2019-10-28
fix(test): Add coveragePathIgnorePatterns property

haruelrovix:
### Problem

Since Jest ignores `searchbar/__tests__/common.js` [#L74](https://github.com/react-native-elements/react-native-elements/blob/64315d6469cc34b67393e8c192d581118ed341c0/package.json#L74) we also need to ignore coverage for it as well.

image

### After the changes

image


iRoachie:
Good catch 💯

PR #2163, opened 2019-10-29
Changes marginVertical to 5

wdevon99:
Changes margin vertical to 5 since the current value is too less


iRoachie:
Why do you think it is too less?

PR #2166, opened 2019-10-30
support styles for Image's ImageComponent

kpalanick:
ability to pass styles into ImageComponent. An example is to pass an a `borderRadious` to an NativeImage component to have the image rounted


iRoachie:
Just use the containerStyle prop. https://snack.expo.io/@roach_iam/trusting-scones

kpalanick:
@iRoachie thanks for the example. `overflow: hidden` was the missing part

PR #2169, opened 2019-11-03
Update divider.md

ifty-rahman:
React components shouldn't end with ; as it counts as a character and cause an exception "Invariant Violation: Text strings must be rendered within a <Text> component"


flyingcircle:
I'll give you credit that you made this PR first, but PR #2186 which was a duplicate of this one was already merged. Thank you for the PR.

Issue #2089, opened 2019-09-26
Input: Change inputComponent to InputComponent


iRoachie:
This is to standardize the props for all component that accept components (e.g. ViewComponent for Button).

This is a breaking change.

Change can be made here https://github.com/react-native-training/react-native-elements/blob/next/src/input/Input.js#L167

jatinAroraGit:
Hey there,
I would like to work on this.
Can you assign this to me please ?
Thanks.

iRoachie:
Sure thing go ahead! Make your change against the `next` branch

iRoachie:
@jatinAroraGit Still interested in working on this?

PR #2176, opened 2019-11-06
change inputComponent to InputComponent

flyingcircle:
Fixes: #2089


src/input/Input.js
```diff
@@ -67,7 +67,7 @@ class Input extends React.Component {
       leftIconContainerStyle,
       rightIcon,
       rightIconContainerStyle,
-      inputComponent: InputComponent = TextInput,
+      InputComponent: InputComponent = TextInput,
```
iRoachie:
```suggestion
      InputComponent = TextInput,
```

flyingcircle: Thanks for the catch. I pushed another commit to resolve this.

docs/input.md
```diff
@@ -75,7 +75,7 @@ import { Input } from 'react-native-elements';
 - [`errorMessage`](#errormessage)
 - [`errorStyle`](#errorstyle)
 - [`errorProps`](#errorprops)
-- [`inputComponent`](#inputcomponent)
+- [`InputComponent`](#inputcomponent)
```
iRoachie: Can you move this to the bottom of the list?  See how Button does it.

flyingcircle: moved to the bottom

src/input/Input.js
iRoachie: You still need to change this

flyingcircle: I tried that change, but the tests fail. I'm not sure why though. I thought that should provide that variable as a default? But it doesn't seem to.

iRoachie: OKay I'd say put it as `InputComponent`,

Then add a default props section at the bottom.

```js
Input.defaultProps = {
  InputComponent: TextInput
}
```

iRoachie: The tests also use snapshot testing. So you need to update the snapshots. See https://react-native-elements.github.io/react-native-elements/docs/testing.html#snapshot-testing

flyingcircle: Alright got that piece working.

iRoachie:
Thank you for your work 🚀

PR #2178, opened 2019-11-08
fix(social): Adding this library did not add react-native-vector-icons

lustyrain:
Hi 
When the library was actually used in a project, an error occurred in the card component.
When I looked it up, it was registered in devDependencies.
Is this a mistake?

Please check.


iRoachie:
This wouldn’t change anything. Vector icons is a peer dependency of this project. Which means you bring your own version of vector icons.

Since it has native file(icon fonts) it’ll need to be installed in your project as a dependency and the files added.

PR #2181, opened 2019-11-12
Added onPressIn and onPressOut of touchable's

yoeven:


flyingcircle:
Not sure I understand what this prop means, but it would also require document updates, etc. Let me know if you're still interested in working on this PR, and I'll reopen.

PR #2186, opened 2019-11-17
Remove trailing semi-colon

conorgil:
I copy/pasted this example and it produced an error because of the trailing semi-colon. It was saying `Invariant Violation: Text strings must be rendered within a <Text> component` because it semi-colon is text. Took me a while to figure out what was going on there.


iRoachie:
Thanks 💯

PR #2193, opened 2019-11-25
Add prop to disable testID in Image to allow for react-native-fast-image on Android

jehartzog:
We are using [react-native-fast-image](https://github.com/DylanVann/react-native-fast-image) as the `ImageComponent` in our `Avatar`, but there is a [bug](https://github.com/DylanVann/react-native-fast-image/issues/221) in that lib that causes issues on Android when `testID` is set on an `Image`.

Since the `testID` is set by spreading the props, it doesn't allow us to override it with `undefined` or `null`, as it will revert to using the default `RNE__Image`.

I attempted using an empty string but that did not resolve the bug.

This PR adds an optional prop to completely allow disabling the setting of any `testID` on an `Image` component.

Our usage of this:
```js
import FastImage from 'react-native-fast-image'
import { Avatar } from 'react-native-elements';
...
    <Avatar
      // Use FastImage
      ImageComponent={FastImage}
      imageProps={{disableImageTestID: true}}
      {...this.avatarProps}
    />);
```

### DIFF
--- a/docs/divider.md
+++ b/docs/divider.md
@@ -11,7 +11,7 @@ distinction between sections of content.
 ```js
 import { Divider } from 'react-native-elements';
 
-<Divider style={{ backgroundColor: 'blue' }} />;
+<Divider style={{ backgroundColor: 'blue' }} />
 ```
 
 ---
--- a/src/image/Image.js
+++ b/src/image/Image.js
@@ -56,7 +56,7 @@ class Image extends React.Component {
         style={StyleSheet.flatten([styles.container, containerStyle])}
       >
         <ImageComponent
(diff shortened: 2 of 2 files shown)


jehartzog:
The CI check failures don't seem to be associated with these changes. Anything else I can do to help get this merged?

jehartzog:
Closing this. There is a much better way to fix this edge case without needing to add a special case prop: https://github.com/DylanVann/react-native-fast-image/issues/221#issuecomment-590833047.

I just implemented it and it works like a charm. It was kind of gnarly to get TS happy with it, but otherwise easy.

PR #2199, opened 2019-12-03
Update Header.js

25juan:
fixed oppo r9m crash


yosimasu:
@Monte9 

Here is a list of my clients.
They will crash with Header without fix.
After I apply this patch, the crash issue is gone.
image002

reallijie:
Thank you. It's work for me.

Issue #2198, opened 2019-12-02
Can't remove horizontal padding in SearchBar component


darrylyoung:
**Component**: SearchBar

**Explain what you did**

I'm trying to remove the padding in the `SearchBar` component.

**Expected behavior**

I would've expected to be able to remove the extra padding with `{{ padding: 0 }}`.

**Describe the bug**

**To Reproduce**

Wrap the `SearchBar` in a wrapper that gives `16px` padding on both sides. You can see this in the screenshot. It's a little hard to see with the background being white but know that the two blue cards are both `16px` from the outer edges of the screen.

I want the outer grey part of the `SearchBar` to also align with these cards. The component itself is aligning to them but the actual input inside the container seems to have padding pushing it in. I've given the container a red background to show what I mean. It looks like there's `8px` padding on the wrapper. Looking at the [code](https://github.com/react-native-elements/react-native-elements/blob/next/src/searchbar/SearchBar-default.js#L89) – I think this is the right code – it looks like there's padding set by the theme but that anything passed to `containerStyle` should work. In this case, I thought passing `padding: 0` to `containerStyle` would work but it doesn't appear to do anything.

![PNG image](https://user-images.githubusercontent.com/1460800/69978949-9ffb9c80-152d-11ea-8ed3-80ca186b2f7d.png)

```tsx
<SafeAreaView style={styles.container}>
  <SearchBar
    platform="ios"
    placeholder={translate('common.searchIn', { name: categoryName })}
    onChangeText={handleSearch}
    lightTheme={true}
    value={searchQuery}
    containerStyle={{ backgroundColor: 'red', padding: 0 }}
    inputStyle={{ color: color.palette.offBlack }}
    inputContainerStyle={{ backgroundColor: color.palette.iOSLightGrey }}
  />
```

I've also tried adding `margin: 0` and `padding: 0` to `containerStyle`, `inputStyle`, and `inputContainerStyle`.

**Screenshots**

**Your Environment (please complete the following information):**

| software              | version |
| --------------------- | ------- |
| react-native-elements |  1.2.7       |
| react-native          |     Expo sdk-35.0.0    |

darrylyoung:
I noticed that using works in `containerStyle={{ padding: 0 }}` to achieve what I want when **not** using `platform="ios"`. I would like to, however, use the iOS style. Is there a better workaround?

darrylyoung:
[...]

PR #2204, opened 2019-12-05
Allow right margin on SearchBar and still have it adjust when cancel slides in

darrylyoung:
For reference of how I started with this, and what I tried first, see #2198.

### Scenario

I want the input of the iOS `SearchBar` component to be flush with the edges of the wrapper. I've added a red background to show what I mean. This is default, without messing with any padding or margin. In this case, I want the input of the `SearchBar` to line up with the edges of the light blue cards below.

PNG image

### What I tried

To get the input right to the edges, I tried the following:

```tsx
containerStyle={{
  backgroundColor: 'red',
  paddingBottom: 0, // <- I just wanted to remove the vertical padding for the example
  paddingTop: 0 // <- I just wanted to remove the vertical padding for the example
}}
inputContainerStyle={{
  backgroundColor: color.palette.iOSLightGrey,
  marginLeft: 0 // <- this allows the input to start at the left edge of its container
  marginRight: 0 // <- here is the issue now (see below)
}}
```

This gives me the following:

PNG image

With this code, it does indeed remove the left margin so the input starts flush with the cards below and it looks good on the right side as well but by overriding `marginRight` as part of `inputContainerStyle`, it overrides the [styles you set](https://github.com/react-native-elements/react-native-elements/blob/next/src/searchbar/SearchBar-ios.js#L146) to adjust the right margin when the cancel button slides in (due to the order the CSS is applied). That means it ends up looking like this when the cancel button is shown:

PNG image 2

### Potential solution

Apply the user-supplied `inputContainerStyle` styles _before_ you apply the styles that adjust the margin when the cancel button slides in. For most people, nothing will change, but for the few that want to add their own right margin while still allowing it to slide in nicely with the cancel button, this seems like the way to go. With the change, I'd end up with the following as opposed to what's shown in the image above.

PNG image 3

### DIFF
--- a/src/searchbar/SearchBar-ios.js
+++ b/src/searchbar/SearchBar-ios.js
@@ -143,8 +143,8 @@ class SearchBar extends Component {
           }}
           inputContainerStyle={StyleSheet.flatten([
             styles.inputContainer,
-            hasFocus && { marginRight: this.state.cancelButtonWidth },
             inputContainerStyle,
+            hasFocus && { marginRight: this.state.cancelButtonWidth },
           ])}
           leftIcon={renderNode(Icon, searchIcon, defaultSearchIcon)}
           leftIconContainerStyle={StyleSheet.flatten([


flyingcircle:
So thinking about this a little bit more, I like the idea, but this then would mean that anyone who did anything with `marginRight` would become affected by this with no way to undo it since the custom style would be overwritten.

flyingcircle:
I came up with a different way that you could perform this task:

```js
<SearchBar
  platform="ios"
  onChangeText={this.updateSearch}
  showCancel={true}
  onFocus={() => this.setState({customSearchInputMargin: false})}
  onBlur={() => this.setState({customSearchInputMargin: true})}
  value={search}
  containerStyle={{
    backgroundColor: 'red',
    paddingBottom: 0,
    paddingTop: 0
  }}
  inputContainerStyle={[{
    marginLeft: 0,
  },
    this.state.customSearchInputMargin ? {marginRight: 0} : {}
  ]}
/>
```

And I think this is probably the better route than forcing a `marginRight` on the component that can't be overwritten.

PR #2211, opened 2019-12-10
Image Component shoule accept onLoad prop

Qiyu8:
fix issue 2206

### DIFF
--- a/docs/image.md
+++ b/docs/image.md
@@ -42,6 +42,7 @@ import { Image } from 'react-native-elements';
 - [`placeholderStyle`](#placeholderstyle)
 - [`transition`](#transition)
 - [`PlaceholderContent`](#placeholdercontent)
+- [`onLoad`](#onLoad)
 - [`ImageComponent`](#imagecomponent)
 
 ---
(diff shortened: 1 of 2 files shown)


TomJKlug:
For my own education. Won't this cause issues with the animation? Since you aren't running the onLoad that is provided with the RNE component? Should it not run the onLoad of the RNE and the onLoad passed?

iRoachie:
Correct @TomJKlug, this PR fixes one problem and breaks the Image component.

Qiyu8:
The purpose of RNE image is to ensure customized smooth image load transitioning, so the original onLayout、onLoad、onLoadStart and onLoadEnd function is abandoned. if someone wants to use these function, they should import image from react-native, is that correct?

iRoachie:
They might want to keep the same animation behaviour but have custom logic on the `onLoad`.

Qiyu8:
we can put the animation behaviour on **onLoadEnd**, which is not commonly used, while keep the important passed function **onLoad**, How about that?

iRoachie:
Already fixed by https://github.com/react-native-elements/react-native-elements/commit/69ea8819e9666157f48e756e272ba0396ea52c83

PR #2213, opened 2019-12-10
Update Badge.js

### LINKED ISSUE #2212: Badge component should expose textProps
When font scaling is enabled in a device there's no way to apply `allowFontScaling` property to `Badge` component.

Exposing `textProps` (of the `Text` component) would allow setting this prop (`allowFontScaling`) to `false` in turn resolving the problem.

I can apply `allowFontScaling` to static `defaultProps` of react-native `Text` component, but it doesn't play nice with TypeScript and latest React.

FDiskas:
Closes #2212


FDiskas:
Missing type definition updates - but... it's not my business :)

FDiskas:
would be nice any comments :D :cop:

flyingcircle:
1. It doesn't correctly solve the issue it claims to fix
2. You made a whole other series of changes that are outside of the original scope of the PR. I agree that the project needs to transition to typescript, but it should be in its own PR.

FDiskas:
So i need just to change title of this pull request?

flyingcircle:
And fix your changes, and likely update your PR to be current.

PR #2232, opened 2020-01-06
Feat: add imageContainerStyle to Card

diamont1001:
I need to do some styling to the Image of Card.


flyingcircle:
You can already do this with:
```js
imageProps={{containerStyle: imageContainerStyle}
```

PR #2239, opened 2020-01-15
Update index.d.ts

amiralidev:


flyingcircle:
@amiralidev type is already defined in `IconObject`, so no need to add it there.

PR #2247, opened 2020-01-27
fixes Avatar when rounded is true

omid-ebrahimi:
The rounded prop doesn't work in some cases. I've fixed it by injecting a borderRadius prop to the Image component.

### DIFF
--- a/src/avatar/Avatar.js
+++ b/src/avatar/Avatar.js
@@ -138,6 +138,7 @@ const Avatar = ({
           overlayContainerStyle,
         ])}
         source={source}
+        borderRadius={rounded ? width / 2 : undefined}
         {...imageProps}
         style={StyleSheet.flatten([
           styles.avatar,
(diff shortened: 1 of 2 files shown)


iRoachie:
Can you specify in which cases it doesn't work? Can you provide an example?

omid-ebrahimi:
I have a chat screen that shows profile pictures as a rounded image (All messages are rendering in a FlatList). I remember when I was scrolling the screen, rounded images that become near to the top or bottom of the screen, were loosing their rounded corners. So I could avoid this behavior by passing an imageProps as follow:
```
const imageProps = { borderRadius: 35.5 };

function Picture({ username, picture }): JSX.Element {
    return (
        <Avatar
            rounded
            size={73}
            source={picture ? { uri: picture.url } : undefined}
            title={username[0].toUpperCase()}
            imageProps={imageProps}
        />
    );
}
```
Passing such props makes the same effect as this pull request does.

PR #2249, opened 2020-01-28
Add card's clickable image

morteify:
This PR is related to #814

First part, before the Expo's "refreshing", presents component without passing _onPressImage_ prop, hence the TouchableOpacity inactive. After passing proper function, TouchableOpacity becomes responsive and reacts to consecutive interactions with defined behavior. 

TRIM_20200128_214125

Usage in basic code example:
```
import React, { useState } from "react";
import { StyleSheet, View } from "react-native";
import { Card, Icon, Button, Text } from "react-native-elements";

export default function App() {
  const [isTextShown, setTextShown] = useState(true);

  const handlePressImage = () => {
    setTextShown(!isTextShown);
  };

  return (
      {isTextShown && <Text h4>HELLO WORLD</Text>}
      <Card
        title="Default title"
        image={require("./assets/big-whale.jpeg")}
        onPressImage={handlePressImage}
      >
        <Text style={{ marginBottom: 10 }}>
          The idea with React Native Elements is more about component structure
          than actual design.
        </Text>
        <Button
          icon={<Icon name="code" color="#ffffff" />}
          buttonStyle={{
            borderRadius: 0,
            marginLeft: 0,
            marginRight: 0,
            marginBottom: 0
          }}
          title="VIEW NOW"
        />
      </Card>
    </View>
  );
}

  container: {
    flex: 1,
    backgroundColor: "#fff",
    alignItems: "center",
    justifyContent: "center"
  }
});
```

### DIFF
--- a/src/card/Card.js
+++ b/src/card/Card.js
@@ -1,6 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { View, Platform, Image as ImageNative, StyleSheet } from 'react-native';
+import {
+  View,
+  Platform,
+  Image as ImageNative,
+  StyleSheet,
+  TouchableOpacity,
+} from 'react-native';
 
 import normalize from '../helpers/normalizeText';
 import { fonts, TextPropTypes, ViewPropTypes, withTheme } from '../config';
@@ -15,6 +21,7 @@ const Card = props => {
     containerStyle,
     wrapperStyle,
     imageWrapperStyle,
+    onPressImage,
     title,
     titleStyle,
     titleNumberOfLines,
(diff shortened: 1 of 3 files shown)


rosalynpoort:
Hello, will this be merged soon? would like to use this function :) Thanks!

flyingcircle:
I wonder if it might be better to make the general `Image` component to accept an `onPress` prop instead, and then from `Card` you'd be able to pass it through `ImageProps`. What do you think? Something like when you pass `onPress` to `Image` it uses a Touchable as its container rather than the standard `View`

morteify:
@flyingcircle I'm not sure if that'd be the best solution here, because there are already two convenient ways to create a standalone Image component handling presses: you can either wrap Image with TouchableOpacity (what I'm proposing) or pass the entire custom component as a Image's prop `ImageComponent` ([docs](https://react-native-elements.github.io/react-native-elements/docs/image.html#imagecomponent)). Maybe there's no need to introduce a third way and add an additional complexity to a fairly straightforward component :)

flyingcircle:
@morteify before I can merge this, this needs to have the docs and types file updated.

flyingcircle:
Closing in favor of #2506

PR #2253, opened 2020-02-03
Default Icon set for the Components

### LINKED ISSUE #2227: Default Icon set for the Components
I was linking react-native-vector-icons and was deleting some of the unused font files from build, and it gave me error saying MaterialIcons is required because SearchBar uses the default Icon search and clear.

When I went to your code the [getIconType.js](https://github.com/react-native-elements/react-native-elements/blob/next/src/helpers/getIconType.js), it has the switch case default registerd to MaterialIcons, and because of that some of my icons are IonIcons and some default ones are MaterialIcons.

I wanted my app's IOS to use IonIcons & android to use MaterialIcons. But there is no such way other than overriding with the components props (LeftIcon, RightIcon, CloseIcon etc...) with my wished icons may be based on OS

Can a registerHelper can be exposed to set defaultIcon set for the OS, in that way I can be specific to use one icon for ios and other for android if not set it can always use MaterialIcons for both

iqbalfaisal:
Fixed #2227 issue. Now default type of icon for ISO is `Ionicon` and `MaterialIcon` for Android. Happy Coding

### DIFF
--- a/src/helpers/getIconType.js
+++ b/src/helpers/getIconType.js
@@ -1,3 +1,4 @@
+import { Platform } from 'react-native';
 import ZocialIcon from 'react-native-vector-icons/Zocial';
 import OcticonIcon from 'react-native-vector-icons/Octicons';
 import MaterialIcon from 'react-native-vector-icons/MaterialIcons';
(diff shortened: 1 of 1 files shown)


iRoachie:
This isn't the way to fix this. Let's discuss more on the issue

PR #2255, opened 2020-02-03
Add ability to overwrite Image onLoad function

### LINKED ISSUE #2206: Image Component doesn't honor onLoad passed to it.
**Explain what you did**

Image Component onLoad is not being fired. When given a function through props.

**Expected behavior**

I expect the onLoad I supply to the Image Component to run.

**Describe the bug**

The Image component doesn't honor my onLoad function and instead only runs its own onLoad method.

**To Reproduce**
1. Create and Image component
2. Give it an onLoad function that console logs
3. Have it run
4. It won't console log.

`<ImageComponent
          testID="RNE__Image"
          {...attributes}
          onLoad={this.onLoad}`

The problem is here in Image.js

react-native-elements 1.2.7
react-native 0.61.5

flyingcircle:
Fixes #2206

### DIFF
--- a/docs/image.md
+++ b/docs/image.md
@@ -43,6 +43,7 @@ import { Image } from 'react-native-elements';
 - [`transition`](#transition)
 - [`PlaceholderContent`](#placeholdercontent)
 - [`ImageComponent`](#imagecomponent)
+- [`onLoad`](#onload)
 
 ---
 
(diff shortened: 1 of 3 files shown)


flyingcircle:
This import is currently causing an error in react-native-web v0.12 since they upgraded to using RN-0.61. This import is no longer available from RN.

xavier-villelegier:
Hey @flyingcircle ! 👋 Thanks for your PR ! 💯 

This issue was fixed in https://github.com/react-native-elements/react-native-elements/commit/69ea8819e9666157f48e756e272ba0396ea52c83 by @fwuensche during a meetup where we take simple issues and onboard people on how to contribute to an open source project.

The idea was not to override completely the `onLoad` function because we still want the the loading thing to work, but just trigger the onLoad props at the end of the custom `onLoad` function if there is one.
We also didn't need to update the docs as it's already specified at the beginning of the page that the `Image` component takes all the React Native `Image` props (including `onLoad`)

PR #2272, opened 2020-02-26
fix(CheckBox): Allow default checkbox testID to be overwritten

Asvarox:
Allows the default `testID` prop value to be overwritten when such prop is provided.


iRoachie:
Sure!

PR #2273, opened 2020-02-27
ADA (Americans with Disabilities Act) Compliance

akshaypunchh:
As TalkBack (Android) / Voice Over (iOS) is not working properly.

### DIFF
--- a/src/buttons/ButtonGroup.js
+++ b/src/buttons/ButtonGroup.js
@@ -71,6 +71,8 @@ const ButtonGroup = props => {
         return (
           <View
             key={i}
+            accessible={true}
+						accessibilityLabel={`${buttons[i]} ${isSelected ? ' is selected' : ' double tap to select'}`}
             style={StyleSheet.flatten([
               // FIXME: This is a workaround to the borderColor and borderRadius bug
               // react-native ref: https://github.com/facebook/react-native/issues/8236


akshaypunchh:
@iRoachie Please review this.

iRoachie:
How can I test this using the simulator?

akshaypunchh:
@iRoachie to test on simulator, you can use Accessibility Inspector of Xcode. 
Screenshot 2020-03-02 at 10 55 03 AM . 

if you have real device you can test it by enable talkback (for android) and voice over (for iOS). 

**How to activate VoiceOver on iPhone and iPad**

1. Launch the Settings app from your Home screen.
2. Tap on General.
3. Tap on Accessibility.
4. Tap on VoiceOver under the Vision category at the top.
5. Tap the VoiceOver switch to enable it.

akshaypunchh:
@iRoachie Hi.. This will affect current functionality. It is useful for blind person or low-vision person.

flyingcircle:
@akshaypunchh I have some experience in making apps accessible. Have you done a review to find where we aren't in compliance? Can you create a spreadsheet on which elements are not working as intended?

akshaypunchh:
@flyingcircle Thanks for response. As you know, ButtonGroup is used to create Tabs.  Now enable talkback/voice over , you can't determine which tab is selected or which have to select.  
A basic knowledge sharing about talkback/voice over, It just speak out the text for the user who have vision disability so he can use the app.

flyingcircle:
@akshaypunchh I think I placed that comment in the wrong thread. I think I mixed it up with another issue. So reviewing the code. I agree that this needs to be fixed, but changing the accessibilityLabel is not the correct way to do it. The `accessibilityState` needs to be set instead. Otherwise, a user would be unable to set their own label in the button group component.

Issue #2312, opened 2020-04-16
Image component tintColor style does not work


bscaspar:
The Image component style prop `tintColor` does not change the color for non-transparent pixels which would be the expected behavior according to the [react native docs](https://reactnative.dev/docs/0.36/image-style-props#tintcolor) and the [RNE docs](https://react-native-elements.github.io/react-native-elements/docs/image.html) that mention this component "Also receives all React Native Image props"

Example code which works if Image is imported from react-native instead:
```js
import { Image } from 'react-native-elements'

export const ColoredImage = () => (<Image style={{ height: 40, width: 40, tintColor: '#ff0000' }} /> )

```

[Snack](https://snack.expo.io/@bcaspar/rne-image-tint-color)

oliverdolgener:
Can confirm. Same problem with Avatar imageProps

cooperwalter:
Here's a simple Snack demonstrating the bug. It shows the expected behavior on RN's Image component and the current behavior on RNE's Image component.

https://snack.expo.io/@cooperawalter/react-native-elements-image-tintcolor-bug

oliverdolgener:
It's because `containerStyle` does not apply to the Image component but to the View around it.
https://github.com/react-native-elements/react-native-elements/blob/next/src/image/Image.js#L59

There should be a tweak in this line: https://github.com/react-native-elements/react-native-elements/blob/next/src/image/Image.js#L65 to support `tintColor`

bscaspar:
Looks like we arrived at the same conclusion! I submitted a PR (linked above) with a fix that applies all styles to the base Image component at that line.

PR #2313, opened 2020-04-16
fix: Image component not applying style prop other than width, height

bscaspar:
Applies additional style props provided to `<Image />` to the `<ImageComponent />` and flattens all the styles on this component.
Addresses #2312

### DIFF
--- a/src/image/Image.js
+++ b/src/image/Image.js
@@ -52,6 +52,7 @@ class Image extends React.Component {
       ...attributes
     } = this.props;
     const hasImage = Boolean(attributes.source);
+    const { width, height, ...styleProps } = style;
 
     return (
       <View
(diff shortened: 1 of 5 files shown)


flyingcircle:
Thanks @bscaspar this will make it into 2.0.0

src/image/Image.js
```diff
@@ -62,13 +63,14 @@ class Image extends React.Component {
...
             StyleSheet.absoluteFill,
             {
-              width: style.width,
-              height: style.height,
+              width: width,
+              height: height,
             },
-          ]}
+            styleProps,
+          ])}
```
iRoachie, line 73: @flyingcircle was just looking at this PR you merged. Any reason you went this route?

Could've simply done:

```jsx
style={StyleSheet.flatten([
  StyleSheet.absoluteFill,
  style
])}
```

flyingcircle: I did merge this, though I'm not the author. It didn't look too different from the previous commit, so I guess I didn't consider how to simplify it from even its previous version. I can say that I am now read up on the Stylesheet documentation ;P.

PR #2319, opened 2020-04-20
Tool Tip RTL Support added - Bug Fix

### LINKED ISSUE #2125: Tooltip: Icon and pointer shown at the wrong place when using RTL and showing the tooltip
Using version 1.1

**Explain what you did**
Create a view using the Tooltip component. Sets the RTL to true `I18nManager.forceRTL(true)`

The tooltip icon is positioned on the top right of the view as on screenshot below.
When using RTL, react-native does its magic and flip the whole UI. The tooltip icon is then on on the top left of the screen as expecetd

**Expected behaviour**
When using RTL, I am expected the tooltip icon to be on the left when the tooltip is visible. It is not the case. For some reasons, it flips back to it's non-RTL position

**Describe the bug**
Everything works great with the tooltip component until I set the RTL property to true in the I18nManager. 

Example: 

* Using a non-RTL Language (English)

Everyting works as expected. Tooltip icon is on the left
image

* Using an RTL Language such as Arabic or Hebrew (the text is not translated fully but it does not matter for this problem)

image

image

You can see that the actual tooltip icon in on the left. But when toggling the tooltip, it is suddenly on the right again. The same goes with the pointer that is also not located at the expected position

**To Reproduce**
Set RTL to true on any example using the tooltip (provided that the tooltip is not centered) and you should be able to reproduce

abhigna-shah:
#2125 

This PR includes fix for "Tooltip: Icon and pointer shown at the wrong place when using RTL and showing the tooltip "
Now, Icon positioning will be now displayed at a proper place for RTL.


flyingcircle:
I think instead of having copied code for right and left it may be better to do something like this:

```js
rtlDirection = I18nManager.isRTL ? 'right': 'left';

{
  [rtlDirection]: x
}
```
Also a unit test to make sure this is working (and doesn't break in the future) would be good

flyingcircle:
Closing in preference of #2340

PR #2316, opened 2020-04-19
Typo correction

dsouzaedison:
Typo correction on the contributing.md

PR #2321, opened 2020-04-20
[Docs]: Minor typographical fix

jamesgeorge007:
Follow up of #2316


flyingcircle:
Having 'Once' two sentences in a row sounds a bit odd to me. Maybe you can change the verbiage in the second sentence to use 'if' or something else other than 'once'?

flyingcircle:
Thanks @jamesgeorge007

Issue #1266, opened 2018-06-17
Bottom Sheet component


iRoachie:
Think it would be really cool if we could have a component like this. 

<img src="https://user-images.githubusercontent.com/5962998/41509798-511a12cc-7227-11e8-8612-cb573e73fcc0.gif" width="400" />

### Specs
Props could be:
- list of items with their icon
- title
- visible
- onRequestClose
- onSelect

saxenanihal95:
I have created this : 
![Screen-Recording-2020-04-21-at-11 22 27-PM](https://user-images.githubusercontent.com/17348595/79897531-8220cc80-8427-11ea-8f57-b1dc2649690a.gif)

flyingcircle:
@saxenanihal95 I don't think anyone else has started on it. It would be good for it to have default styling that more or less matches the other components as well as the ability to inject a custom component. It'll be easiest to discuss with an API doc as well. When you feel like you have something a bit more solid, open a PR and we can discuss tweaks from there.

PR #2325, opened 2020-04-24
Created BottomSheet component

saxenanihal95:
PR for #1266 (Bottom Sheet Component)

Right Now i have added three props:

buttonProps: which will take all props of Button Component of React Native Elements

cancelButtonIndex : index of Button which will be the index of list of Bottom Sheet and will close on Click

list: Array of list item props of ListItem Component which will show in list of ButtomSheet component


flyingcircle:
I'm not sure I understand the `buttonProps` field.  There's a cancel button that usually goes at the bottom, and I can understand that there could be a list of buttons, but `buttonProps` makes it sound like there's only one button. It might be good to have an example with pictures as well with the documentation.

saxenanihal95:
@flyingcircle will add example with pictures 

and `buttonProps` is for the button which shows the ActionSheet Component

saxenanihal95:
@flyingcircle updated PR with examples

flyingcircle:
@saxenanihal95 We are ready to start finishing up this component and getting it into RNE. I recently simplified the react-native-elements-app project. So please add a tab for showing an example of how to use this.

src/bottomSheet/BottomSheet.js
```diff
@@ -0,0 +1,94 @@
...
+
+const styles = StyleSheet.create({
+  safeAreaView: { flex: 1, backgroundColor: 'rgba(0,0,0,0.2)' },
+  modalView: {
+    flex: 1,
+    flexDirection: 'column-reverse',
+  },
+  listContainer: { backgroundColor: 'white' },
+});
+
```
flyingcircle, line 85: To follow the pattern, `BottomSheet.propTypes` needs to be defined here.

airen29:
Could we maybe add a functionality that will close the BottomSheet when button is pressed? Not that the only last one?

For example:
```
onPressClose: () => {
              this.setVisible(false);
              onPress && onPress();
            },
```

and `onPressClose` will be by default set to `false`.

PR #2328, opened 2020-04-26
A prop to allow control over button / icon size ratio

whileMeAwakeMeCode:
I propose to add a 'buttonStyle' prop to control the sizes of 'raised' or 'reversed' icons : ratio btw the icon itself and its circular container.


flyingcircle:
This is already possible using the `containerStyle`

Issue #2290, opened 2020-03-05
Pricing with configurable button


remithomas:
I would like to have a Pricing component with a disabled button, for example this pricing will be available next month...

My solutions are :
- Provide the button props to the pricing (this includes also other props of the button)
- Or provide the component button, this unlocks also the fact we can have animated icon or custom image.

I have considered to redesign myself in my app the pricing component.

thanks

PR #2332, opened 2020-04-27
improve PricingCard button props

flyingcircle:
Fixes: #2290

Changed pricing button so that it can accept all Button props. I also discovered that it would be possible for a developer to make a light themed button with a dark icon. So I improved the `button.icon` prop so that it can accept a string (and maintains old format), an object of Icon props, or a custom Icon.

Looking forward I think we could potentially deprecate the `onButtonPress` prop since it can be passed directly through button props instead.

### DIFF
--- a/src/pricing/PricingCard.js
+++ b/src/pricing/PricingCard.js
@@ -99,9 +99,15 @@ const PricingButton = props => {
       ])}
       titleStyle={titleStyle}
       onPress={onButtonPress}
-      icon={React.isValidElement(icon) ? icon : (
-        typeof icon === 'string' ? <Icon name={icon} size={15} color="white" /> :
-        <Icon {...icon}/>)}
+      icon={
+        React.isValidElement(icon) ? (
+          icon
+        ) : typeof icon === 'string' ? (
+          <Icon name={icon} size={15} color="white" />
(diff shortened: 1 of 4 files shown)


MegaMaddin:
Nice, that's exactly what I have been waiting for. Is there an ETA when this will be released?
Thanks!

flyingcircle:
@MegaMaddin We're planning a patch release for later this week, early next. I've recently been trying to think through how to refactor the code/ rethink how these components are actually built. But I can see about doing a 2.1 release in the next month to get some of these additions out there.

src/pricing/PricingCard.js
```diff
@@ -65,29 +65,61 @@ const PricingCard = props => {
...
+      titleStyle={titleStyle}
+      onPress={onButtonPress}
+      icon={
+        React.isValidElement(icon) ? (
+          icon
+        ) : typeof icon === 'string' ? (
+          <Icon name={icon} size={15} color="white" />
+        ) : (
+          <Icon {...icon} />
+        )
```
iRoachie, line 109: Could've used https://github.com/react-native-elements/react-native-elements/blob/next/src/helpers/renderNode.js#L3 here

Issue #2330, opened 2020-04-27
RN 0.62 now requires `useNativeDriver` be specified with use of Animated.


dquessenberry:
**Explain what you did**
After upgrading to React Native v0.62.2, I am now seeing a warning when calling the animated shake() method of the Input component. Note: this is just a warning and the shake animation still works fine. I suppose there are may be other RNE components that use Animated.timing() that need to specify useNativeDriver as well.

**Expected behavior**
I expect no warnings when using a supported method.

**Describe the bug**
This is a minor issue, not a major bug. Apparently as of v0.62.0 React Native requires useNativeDriver to be specified with Animated functions (whether true or false), see a commit explanation [here](https://github.com/facebook/react-native/commit/5876052615f4858ed5fc32fa3da9b64695974238). 

**How to fix**
All that is required is to pass `useNativeDriver` to animated functions calls: For example: 
` // RNE Input.js, lines 53-58:`
`  Animated.timing(shakeAnimationValue, {`
`    duration: 375,`
`    toValue: 3,`
`    ease: Easing.bounce,`
`    useNativeDriver: true, // <-- **Add this**`
`  }).start();`
`};`

**To Reproduce**
Upgrade to RN v0.62.0 or higher, call the RNE Input's shake animation method, see warning

**Screenshots**
<img width="584" alt="Screen Shot 2020-04-26 at 6 40 12 PM" src="https://user-images.githubusercontent.com/10273674/80326410-86653500-87ed-11ea-81b2-5d4f401b3dea.png">

**Your Environment (please complete the following information):**

| software              | version |
| --------------------- | ------- |
| react-native-elements |  1.2.7  |
| react-native          |  0.62.2  |

dquessenberry:
I pulled down RNE and did a quick search, and found only two instance of Animated.timing(...): 
1) Image.js: onLoad function. This instance does include the useNativeDriver, however it is set to true for iOS and false for Android - though I'm not sure why.
2) Input.js: shake function. This instance is where I was encountering the warning. I tested with it set to true on both an iOS simulator and an Android emulator and they both worked, and no more warning. Since the Image.js onload function had useNativeDriver set to false on Android, I tested the same with the shake function, but it no longer worked, so I kept it as true.

I forked the project, created a branch, added the useNativeDriver to the shake function, tested, committed, and issued a pull request: [https://github.com/react-native-elements/react-native-elements/pull/2333](https://github.com/react-native-elements/react-native-elements/pull/2333)

[...]

PR #2333, opened 2020-04-27
Setting 'useNativeDriver' now required for Animated.timing(...) funct…

dquessenberry:
…ion call with RN 0.62.x, see: https://github.com/facebook/react-native/commit/5876052615f4858ed5fc32fa3da9b64695974238.

Did a search for all instances of Animated.timing(...), and only found one instance that was not setting the useNativeDriver property: the Input shake method. Added the useNativeDriver property and set it to true. Tested on iPhone 11 (iOS 13.3) simulator and Android Pixel 3 API 29 emulator and both passed. Since the only other instance of the Animated.timing(...) in RNE was the Image onLoad method and it had useNativeDriver set to false on Android and true on iOS, so I also tested useNativeDriver set to false for Android, but the shake animation no longer worked, so I reverted back to true for both Android and iOS.

### DIFF
--- a/src/input/Input.js
+++ b/src/input/Input.js
@@ -54,6 +54,7 @@ class Input extends React.Component {
       duration: 375,
       toValue: 3,
       ease: Easing.bounce,
+      useNativeDriver: true,
     }).start();
   };
 


flyingcircle:
Oops didn't see this before merging my own. For your ticket. I'll see it next time if you make sure to mention the issue number in your PR such as `Fixes: #2330 `

Issue #2257, opened 2020-02-06
latest version 1.2.7 not working with react-native-web 0.12.0


28harishkumar:
**Explain what you did**

- created a new project
- added react-native-elements (1.2.7)
- added react-native-web (0.12.0)
- write hello world code

**Expected behavior**

should print hello world

**Describe the bug**
```
 web  Failed to compile.
project_path/node_modules/react-native-elements/src/config/ViewPropTypes.js
Module not found: Can't resolve 'react-native-web/dist/exports/ViewPropTypes' in 'project_path/node_modules/react-native-elements/src/config'
```

**It is working with `react-native-web: 0.11.7`** 
I found that  0.12.0 is released just 9 days before. **[And they are not going to maintain ViewPropTypes](https://github.com/necolas/react-native-web/issues/1512#issuecomment-568180289).**

samereberlin:
+1... The latest RNE version is not working with RNW 0.12.1 as well (due to the same reason: removal of deprecated APIs). Thanks!

zamponotiropita:
PropStyles (e.g. Image.propTypes) are deprecated in the current RN version. Please remove them.
Otherwise compatibility with the current react-native-web (1.12.0) is broken as it has completely removed PropTypes

flyingcircle:
I believe this is also a requirement for compatibility with RN 62, they removed propTypes from RN core, and I don't think there's a community version quite yet. But it looks like RN is suggesting either using Flow or Typescript instead. Does anyone have experience or thoughts as to which direction we should go for this?

flyingcircle:
There's another issue is that `react-native-vector-icons` is also not at a version that can support rn-web 0.12. Even with this fix, I'm not sure that RN-elements will be able to run with RN-web 12 due to that dependency.  @iRoachie I think this issue will have to wait for beyond version 2.0.0. Do you agree?

Monte9:
@flyingcircle thanks for fixing this in `react-native-ratings`. I have published a new version - v7.0

flyingcircle:
@28harishkumar or @samereberlin can you test out my branch to see if it works for web 0.12 now? To double check the work.

samereberlin:
Hi @flyingcircle , yes it works :-) Thanks!

MahmoudShaeer:
:/ can any one help me  same issue here :/

PR #2338, opened 2020-04-30
fix: remove React.PropTypes since deprecated

flyingcircle:
Fixes: #2257 
Also depends on a fixed version of RN-ratings: https://github.com/Monte9/react-native-ratings/pull/91
@Monte9

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "hoist-non-react-statics": "^3.1.0",
     "opencollective-postinstall": "^2.0.0",
     "prop-types": "^15.7.2",
-    "react-native-ratings": "^6.5.0",
+    "react-native-ratings": "^7.2.0",
     "react-native-status-bar-height": "^2.2.0"
   },
   "devDependencies": {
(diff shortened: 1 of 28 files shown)


mafortier-nexapp:
@flyingcircle In react-native-web, classes in styles coming from a StyleSheet are exported as a number.

These style props like this `containerStyle: PropTypes.object` will throw and crash:

image

flyingcircle:
@mafortier-nexapp that's a warning, not a crash. Do you have info on what was causing the crash?

mafortier-nexapp:
@flyingcircle you're right, that was not the cause of our crash. It's just a warning, don't mind my comment 😄

AzzouDuGhetto:
@flyingcircle I get a new warning after upgrading to 2.0.1 because of this change in Avatar.js: 
```
 - source: RNImage.propTypes.source,
 + source: PropTypes.node,
```

The warning is 
image

flyingcircle:
@AzzouDuGhetto `RNImage.propTypes.source` was deprecated in RN 62 and was throwing an error when using RN-web 12. I'll open a ticket on how to better fix this for 2.0.2

PR #2355, opened 2020-05-11
add useNativeDriver arg to animated input

flyingcircle:
Fixes: #2330

### DIFF
--- a/src/input/Input.js
+++ b/src/input/Input.js
@@ -54,6 +54,7 @@ class Input extends React.Component {
       duration: 375,
       toValue: 3,
       ease: Easing.bounce,
+      useNativeDriver: true,
     }).start();
   };
 


arabold:
There's another case:
```diff
diff --git a/node_modules/react-native-elements/src/slider/Slider.js b/node_modules/react-native-elements/src/slider/Slider.js
index 66fb510..f8480a5 100644
--- a/node_modules/react-native-elements/src/slider/Slider.js
+++ b/node_modules/react-native-elements/src/slider/Slider.js
@@ -16,6 +16,7 @@ const DEFAULT_ANIMATION_CONFIGS = {
     duration: 150,
     easing: Easing.inOut(Easing.ease),
     delay: 0,
+    useNativeDriver: false
   },
 };
```

In this case I had to set `useNativeDriver` to `false` as I got an error about `width` not being supported in Android when using the `<Slider />` component. Are we sure that `true` works for Input fields in all cases?

PR #2371, opened 2020-05-30
move @types/react-native-vector-icons to dev deps

artashsanoyan:
I assumed that all `@types` packages should be installed as dev dependencies.


flyingcircle:
It's not necessarily 100% true that all `@types` packages should be in `devDependencies`. @iRoachie is there any reason why `@types/react-native-vector-icons` is a `dependency` and not a `devDependency`?

PR #2390, opened 2020-06-09
Stop triggering LayoutAnimation without setting state in SearchBar-ios

andrewbeckman:
I discovered a bug when using the iOS SearchBar where if the `showCancel` prop was set to `true`, then my keyboard was mysteriously reappearing and re-dismissing after navigating away from the screen with the search bar focused. After looking into it further, I found the culprit! It looks like when the `showCancel` prop was added to the SearchBar, the `setState` call in `onBlur` to set `hasFocus` to false was put behind a check to only set state when `showCancel` is falsy; however, the LayoutAnimation trigger that is meant to be coupled with the state update was left outside that conditional. This PR puts the LayoutAnimation trigger under the same conditional as setting state so that unexpected animations do not occur.

### DIFF
--- a/src/searchbar/SearchBar-ios.js
+++ b/src/searchbar/SearchBar-ios.js
@@ -80,9 +80,9 @@ class SearchBar extends Component {
 
   onBlur = event => {
     this.props.onBlur(event);
-    UIManager.configureNextLayoutAnimation && LayoutAnimation.easeInEaseOut();
 
     if (!this.props.showCancel) {
+      UIManager.configureNextLayoutAnimation && LayoutAnimation.easeInEaseOut();
       this.setState({
         hasFocus: false,
       });


andrewbeckman:
@flyingcircle is there any more info you need or anything I can do to expedite getting this in for a release asap?

flyingcircle:
@andrewbeckman I'm having difficulty reproducing your issue. Do you have any additional information on the steps you took to see this issue?

andrewbeckman:
> @andrewbeckman I'm having difficulty reproducing your issue. Do you have any additional information on the steps you took to see this issue?

@flyingcircle I am trying to put together an example project that showcases the issue, but as luck would have it, reproducing this is proving difficult. That said, always calling `LayoutAnimation.easeInEaseOut()` when state is set conditionally seems like it can only hurt. My understanding from reading the [docs on `LayoutAnimation`](https://reactnative.dev/docs/layoutanimation) is that you call something like `LayoutAnimation.easeInEaseOut()` right before setting state to automatically animate the visual changes caused by the state change. When state does not get set (namely, when the `showCancel` prop is set to true), it seems to me like there is uncertainty as to what state change will happen next and unexpectedly have an animation. When I ran into this issue, making this change fixed it. I realize that is unsatisfying and I wish I could provide a sample project to show off the issue; however, I do believe this change brings things in line with how `LayoutAnimation` is intended to be used with no risk of breaking existing functionality.

flyingcircle:
Thinking about it more and testing a few things out, I see what you were saying about this line of code. I'm not sure I like the prop name `showCancel`, I had to keep reminding myself what it meant as I was going through the code - lol.

andrewbeckman:
> Thinking about it more and testing a few things out, I see what you were saying about this line of code. I'm not sure I like the prop name `showCancel`, I had to keep reminding myself what it meant as I was going through the code - lol.

Haha yeah `showCancel` could be more descriptive.

Issue #2191, opened 2019-11-21
Custom fontFamily does not work with Text Heading (h1, h2, h3, h4)


bearbytes:
I'm loading a custom font using Expo as described [here](https://docs.expo.io/versions/latest/guides/using-custom-fonts/).

Setting the new font as `fontFamily` style works in regular Text components.

However, as soon as one of the heading props are set on the Text component, the fontFamily is no longer applied.

Example:
```
import { Text } from 'react-native-elements'

<Text
  h1
  h1Style={{ color: 'red', fontFamily: 'dosis-regular' }}
  style={{ color: 'blue', fontFamily: 'dosis-regular' }}
>This is my text</Text>
```

This text is large, red and uses the default font.
After removing the `h1` prop, the text is small, blue and using the [Dosis](https://fonts.google.com/specimen/Dosis) font.

Keeping `h1` but removing `h1Style`, the text is large, blue and using the default font.
Keeping `h1` but removing `style` does nothing.

**Environment:**

| software              | version |
| --------------------- | ------- |
| react-native-elements |   1.2.7      |
| react-native          |     0.59.10    |
| expo | 35.0.0|

oxyii:
@bearbytes I'm not sure about exactly **REGULAR** fonts for using with `h*` props...
https://github.com/react-native-elements/react-native-elements/blob/0f06b924fc7536f7599aca268864a75f33e13b1e/src/text/Text.js#L29-L32
Please try to use **BOLD** font.

bearbytes:
Setting `fontWeight: 'normal'` seems to fix the issue. Loading and setting the Bold variant of the font however doesn't seem to make any difference if the fontWeight is not changed.

Maybe the information about which "weight" the font has is getting lost when loaded by `expo-font`, and all fonts are handled as regular.

For now, I'll change the default fontWeight of all h-elements in the ThemeProvider, thanks!

oxyii:
@iRoachie or we can qualify this issue as a bug?

Usamaliaquat123:
@iRoachie  @oxyii  can i fix this issue... ?

ispulkit:
This is also happening with Button. Custom `fontfamily` and `fontWeight` are not working. 
```
<Button
                  title="Request OTP"
                  buttonStyle={{
                    borderRadius: 8,
                    padding: '4%',
                    backgroundColor: '#0E85DD',
                  }}
                  fontFamily="Nunito-Bold"
                  fontWeight="bold"
[...]

PR #2402, opened 2020-06-18
add regular font to all android fonts

flyingcircle:
fixes: #2191 

@bearbytes
@oxyii 
@seankovacs

I was not able to reproduce directly the issue you were talking about, but can one of you confirm if this fixes the issue?

### DIFF
--- a/src/config/fonts.js
+++ b/src/config/fonts.js
@@ -2,15 +2,15 @@ export default {
   android: {
     regular: {
       fontFamily: 'sans-serif',
-      fontWeight: 'regular',
+      fontWeight: 'normal',
     },
     light: {
       fontFamily: 'sans-serif-light',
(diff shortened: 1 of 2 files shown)


oxyii:
Hmmmmm..... I'm not sure if this needs any changes.

oxyii:
@flyingcircle I think headers must be bold by default. So if anyone want to use regular font, please set `regular` font in `h*Style`

flyingcircle:
@oxyii I think you may need to re-review. If an h*Style is set, then bold is set. The issue as discussed in the ticket is that `fontWeight` must be set for particular fonts in android. You can see at the bottom that `fonts.android.regular` is set by default and `fonts.android.bold` is set  if an h*Style is set.

PR #2411, opened 2020-06-22
Add overlayContainerStyle to Card

justinh00k:
Adding a featuredTitle or featuredSubtitle to a Card adds an overlayContainer with opinionated styling, including an rgba(0,0,0,.2) background. There should be a way for the user to override these styles.

### DIFF
--- a/src/card/Card.js
+++ b/src/card/Card.js
@@ -86,10 +86,12 @@ const Card = (props) => {
               {...imageProps}
             >
               {(featuredTitle || featuredSubtitle) && (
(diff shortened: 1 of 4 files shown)


flyingcircle:
@justinh00k can you fix the conflict in the docs file?

justinh00k:
@flyingcircle looks like a rebase did it

flyingcircle:
@justinh00k I wasn't able to get this to do anything. Is there some styling that you can suggest that makes this prop useful?

flyingcircle:
Obviated by #2518

PR #2436, opened 2020-07-13
stop creating component if string length == 0; https://github.com/rea…

himanishm07:
…ct-native-elements/react-native-elements/issues/2259

### DIFF
--- a/src/helpers/renderNode.js
+++ b/src/helpers/renderNode.js
@@ -14,7 +14,13 @@ const renderNode = (Component, content, defaultProps) => {
   if (content === true) {
     return <Component {...defaultProps} />;
   }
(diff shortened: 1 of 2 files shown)


flyingcircle:
1renderNode` is used in a lot of places throughout the code. I'm not sure what kind of effects this could have on any number of components nor how it might change the current behavior.

himanishm07:
I can make the changes for the specific view itself but as the view should not be created if string is empty, I made the given changes.

If required, can make the changes to specific module itself
Will create one more pr and tag it here

flyingcircle:
Alright tested this out and it looks like this should be good. Thanks for the fix.

PR #2437, opened 2020-07-13
Fixing RN 0.63 issue where iOS onClose does not fire

dorthwein:
Removing is iOS check to fix onClose not firing in RN 0.63


flyingcircle:
@dorthwein Please message me on the RNE slack. I'd like to send you a thank you token for your contribution to the 2.1.0 release.

PR #2444, opened 2020-07-20
Fix issue with avatar not displaying with imageProps containerStyle

danwhite-ipc:
Fix issue with avatar not displaying if passing containerStyle within imageProps


flyingcircle:
@danwhite-ipc Message me on the RNE slack channel. I'd like to send you a thank payment for your commit.

danwhite-ipc:
@flyingcircle I don't need payment, if you wish please donate it to a local hospice or children's charity

PR #2464, opened 2020-07-30
Fixed an important typo

GabrielStofel:
The prop for showing the Edit Button on the Avatar Component is not ```showEditButton```. It's actually ```showAccessory```.


flyingcircle:
Thank you for pointing this out. However, there were a few extra changes that were needed to that file. You can see the change that needed to be made here: https://github.com/react-native-elements/react-native-elements/commit/7cec623fe95b33a4e2eac41728129bd0fc93374c

PR #2469, opened 2020-07-31
Avatar rework

flyingcircle:
This moves `Accessory` out of `Avatar` into its own component. Still backwards compatible.


MikeyFriedChicken:
Are we missing a TypeScript definition for Accessory?

PR #2476, opened 2020-08-03
Modified versions.json to contain correct version number for 2.1.0

alexttyip:
`versions.js` lists `v2.1.0` as `2.1`, which does not match the tag of `v2.1.0`, breaking the link to `v2.1.0`'s release notes in https://react-native-elements.github.io/react-native-elements/versions/.


flyingcircle:
This would also require renaming all the files in the `versioned-sidebar.js` file. To properly change a version in the documents there's a `rename-version` command in the `website` folder. You can run that for this PR or I can do that later. I didn't realize that naming my version 2.1 would break that link.

alexttyip:
@flyingcircle Ah, I see. I can indeed locate the `rename-version` command, but I can't seem to find the corresponding command for docusaurus v2.

Issue #2459, opened 2020-07-29
[Suggestion] ListItem -> Add a new text field


SrBrahma:
**Is your feature request related to a problem? Please Describe.**
Whatsapp app have a ListItem feature that it doesn't seem possible to do with this lib.

![image](https://user-images.githubusercontent.com/28177550/88761693-ffa25a00-d145-11ea-8795-8f744879c4fb.png)

**Describe the solution you'd like**
It looks like it's an 'preTitle', or a subSubTitle (**more** likely to be a preTitle, as my name there is bold, and all titles in the app are bold). So, ListItem would need 2 new props, preTitle and preTitleStyle.

SrBrahma:
I think this got wrongly closed.

Also, came here to say that I implemented this "pretitle" as a title component, but, the icons were logically center aligned to the entire view unlike the post image, that the icons are certainly center aligned to the view of the pretitle and the title. It looks better in the image way.

Also, I can't understand the new ListItem.js file (https://github.com/react-native-elements/react-native-elements/blob/next/src/list/ListItem.js). Will title and subtitle (etc) be deprecated?

flyingcircle:
@SrBrahma Yes title and subtitle are getting deprecated and I'm planning on removing them for v3. With the new version You should be able to create a more custom `ListItem` as you want here. Check out the working blog article that I have on this change: https://react-native-elements.github.io/react-native-elements/blog

Hopefully after all the deprecated code is removed, the code will be much easier to read.

So I think you should be able to implement that view with the new `ListItem`. If you come up with a subcomponent for `ListItem` that ends up creating this look, I'm still up for adding it. But to be honest, to me, it looks quite specific to your use case. Let me know how trying out the new `ListItem` goes.

SrBrahma:
Looks really cool! Will port my current code to it and give you a feedback and post here my code. I was even going to create a new feature request to allow left icon styling (to add a padding between it and the texts), but seems it is already possible with this new feature.

flyingcircle:
`ListItem.Content` is to replace the box `View` that held both the `title` and `subtitle`. I agree, not the prettiest solution. But it was the easiest way to refactor out the old method and have a direct 1-to-1 from the old style to the new style. I'm hoping to do the same thing for `Card` and maybe `Tile` as well.

SrBrahma:
Can't LisItem have by default a ListItem.Content?

PR #2494, opened 2020-08-10
deprecate ListItem subcomponents and avatar.accessory

flyingcircle:
Fixes: #2459
Remove `ListItem` sub-components, deprecate props.
Also deprecates `Avatar` accessory props.

TODO: 
1. double check that example continues to work
2. write blog post detailing upgrade details

### DIFF
--- a/src/list/ListItem.js
+++ b/src/list/ListItem.js
@@ -13,33 +13,24 @@ import { withTheme } from '../config';
 
 import Avatar from '../avatar/Avatar';
 import Badge from '../badge/Badge';
-import CheckBox from '../checkbox/CheckBox';
 import Icon from '../icons/Icon';
 import Text from '../text/Text';
-import ButtonGroup from '../buttons/ButtonGroup';
-import Input from '../input/Input';
+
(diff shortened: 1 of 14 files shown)


src/list/ListItem.js
```diff
@@ -392,4 +391,13 @@ PadView.propTypes = {
...
-export default withTheme(ListItem, 'ListItem');
+
+const ThemedListItem = withTheme(ListItem, 'ListItem');
+ThemedListItem.Chevron = ListItemChevron;
+ThemedListItem.Content = ListItemContent;
+ThemedListItem.Input = ListItemInput;
+ThemedListItem.Title = ListItemTitle;
+ThemedListItem.Subtitle = ListItemSubtitle;
+ThemedListItem.CheckBox = ListItemCheckBox;
+ThemedListItem.ButtonGroup = ListItemButtonGroup;
```
SimenB, line 402: These are not added to the TS types

SimenB: Ah: #2523

Issue #2505, opened 2020-08-15
Add a functionality that will close the BottomSheet when a list item is pressed


JoeDuncko:
**Is your feature request related to a problem? Please Describe.**

I'm trying to use the BottomSheet component to present a list of calendars to the user to add their event to. From what I can tell, there is no way to close the BottomSheet when the user has pressed one of the list items, which makes my use case difficult.

**Describe the solution you'd like**

Add a functionality that will close the BottomSheet when a list item is pressed.

As per a comment by @airen29 on https://github.com/react-native-elements/react-native-elements/pull/2325#issuecomment-671099539, it could look like this:

```javascript
onPressClose: () => {
              this.setVisible(false);
              onPress && onPress();
            },
```

**Describe alternatives you've considered**

There are some other ways this could be implemented. For example, having an `open` variable and a `onClose` function like [Material UI's modals](https://material-ui.com/components/modal/). Another option is making `cancelButtonIndex` into an array, but that doesn't seem super intuitive.

**Additional context**

JoeDuncko:
For conversation's sake, I think it's also a bit weird how coupled the BottomSheet is to the Button component... maybe decoupling them would be good (again, like MUI's modals)...

JoeDuncko:
I just realized that the Overlay component ( https://react-native-elements.github.io/react-native-elements/docs/overlay ) has a really similar API to what I'd like to see from the BottomSheet.

flyingcircle:
@JoeDuncko `BottomSheet` is a very newly added component, so I think making changes to make to API more flexible is a good idea. I think you're right that there shouldn't be a button attached to the component. Instead the Modal should be controlled by the user setting `isVisible` themselves on the coponent, and thus they'd be able to manage the cancel button much more easily a customizable. 
@saxenanihal95 was the creator of the component. His input might be good as well.

saxenanihal95:
@JoeDuncko @flyingcircle Yes i also think Button is tightly coupled to the ButtomSheet Component and also user should have a variable exposed to manage the visibility of the Modal so below the changes I am thinking to do: 

* Remove the Button Component from Bottom Sheet Component
* Expose Modal visible props and other modal props(https://reactnative.dev/docs/modal#props) to the user

flyingcircle:
@saxenanihal95 I think those look like good changes. I think we can justify this for a 2.3 release (we'll say BottomSheet was in "beta" for 2.2) - only change missing is the docs.
[...]

PR #2507, opened 2020-08-15
Fix: Functionality to close the BottomSheet

saxenanihal95:
Fixes: #2505

* Removed Button component as it is tightly coupled
to BottomSheet Component

* Exposed isVisible prop so that user can easily
manage BottomSheet Component visibililty

### DIFF
--- a/src/bottomSheet/BottomSheet.js
+++ b/src/bottomSheet/BottomSheet.js
@@ -7,7 +7,6 @@ import {
   ScrollView,
 } from 'react-native';
 import { withTheme } from '../config';
-import Button from '../buttons/Button';
 import ListItem from '../list/ListItem';
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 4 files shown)


JoeDuncko:
This looks like a much needed change! Question: is the `onClose` prop being used at all here? Should it be?

saxenanihal95:
@JoeDuncko Yes i was first thinking to provide `onClose` but that is already part of `modalProps`  (`onDismiss`) but that is only supported for ios right now.

src/bottomSheet/BottomSheet.js
```diff
@@ -86,14 +65,16 @@ const styles = StyleSheet.create({
...
+  onClose: () => {},
+  isVisible: false,
 };
 
 BottomSheet.propTypes = {
   list: PropTypes.array,
-  cancelButtonIndex: PropTypes.number,
-  buttonProps: PropTypes.object,
+  modalProps: PropTypes.object,
+  onClose: PropTypes.func,
```
flyingcircle: If `onClose` is already provided by `Modal` then I don't think it has to be included here.

JoeDuncko:
Cool, that makes more sense to me now. I think it's a good first step to a much more useful `BottomSheet` component. Thanks so much for getting this in so quickly!

PR #2508, opened 2020-08-17
Update input.md

indavidjool:
There  does  not seem  to be a value property to set on input.current.setNativeProps({ value: 'hello')}, there is a text property though?


indavidjool:
My apologies, seems like this is a version issue.  I forgot to update my packages to the latest version.

Issue #2522, opened 2020-08-27
Typescript build is failing with seemingly RNElements related errors


djMax:
**Explain what you did (Required)**

Upgraded a working Typescript RN project to latest RN elements and react native animated.

**Expected behavior (Required)**

Clean build

**Describe the bug (Required)**

```
../../node_modules/react-native-elements/src/index.d.ts:33:18 - error TS2320: Interface 'TouchableComponent' cannot simultaneously extend types 'TouchableHighlightProps' and 'PressableProps'.
  Named property 'delayLongPress' of types 'TouchableHighlightProps' and 'PressableProps' are not identical.

33 export interface TouchableComponent
                    ~~~~~~~~~~~~~~~~~~

../../node_modules/react-native-elements/src/index.d.ts:33:18 - error TS2320: Interface 'TouchableComponent' cannot simultaneously extend types 'TouchableHighlightProps' and 'PressableProps'.
  Named property 'hitSlop' of types 'TouchableHighlightProps' and 'PressableProps' are not identical.

33 export interface TouchableComponent
                    ~~~~~~~~~~~~~~~~~~

../../node_modules/react-native-elements/src/index.d.ts:33:18 - error TS2320: Interface 'TouchableComponent' cannot simultaneously extend types 'TouchableHighlightProps' and 'PressableProps'.
  Named property 'onLongPress' of types 'TouchableHighlightProps' and 'PressableProps' are not identical.

33 export interface TouchableComponent
                    ~~~~~~~~~~~~~~~~~~

../../node_modules/react-native-elements/src/index.d.ts:33:18 - error TS2320: Interface 'TouchableComponent' cannot simultaneously extend types 'TouchableHighlightProps' and 'PressableProps'.
  Named property 'onPress' of types 'TouchableHighlightProps' and 'PressableProps' are not identical.

33 export interface TouchableComponent
                    ~~~~~~~~~~~~~~~~~~

../../node_modules/react-native-elements/src/index.d.ts:33:18 - error TS2320: Interface 'TouchableComponent' cannot simultaneously extend types 'TouchableHighlightProps' and 'PressableProps'.
  Named property 'onPressIn' of types 'TouchableHighlightProps' and 'PressableProps' are not identical.

33 export interface TouchableComponent
                    ~~~~~~~~~~~~~~~~~~

[...]

Issue #2492, opened 2020-08-09
ListItem typescript error typo upgrade 2.2.0


hengkx:
**Explain what you did (Required)**

**Expected behavior (Required)**

**Describe the bug (Required)**
![image](https://user-images.githubusercontent.com/8358236/89738670-0e293380-daad-11ea-9d06-6bc23a8b03c7.png)

**To Reproduce (Required)**

We highly recommend that you re-create the bug on [Snack](https://snack.expo.io). If not, list the steps that a reviewer can take to reproduce the behaviour:

**Screenshots (Required)**

**Your Environment (Required):**

| software              | version |
| --------------------- | ------- |
| react-native-elements |         |
| react-native          |         |

SrBrahma:
Also happening here.

davidgovea:
Currently working around this using `<ListItem children={null} ...`

PR #2526, opened 2020-08-28
fix touchable type

flyingcircle:
Fixes: #2522 and #2492

@djMax please check this out and see if this fixes the issue.
@SrBrahma @hengkx please also check this out for the fix on the other ticket as well.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -30,9 +30,9 @@ import {
   IconProps as VectorIconProps,
 } from 'react-native-vector-icons/Icon';
 
-export interface TouchableComponent
-  extends TouchableHighlightProps,
-    PressableProps {}
+type Touchable = PressableProps & TouchableHighlightProps;
+
+export interface TouchableComponent extends Touchable {}
 
 /**
  * Supports auto complete for most used types as well as any other string type.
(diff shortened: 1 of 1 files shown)


flyingcircle:
I can't remember which ticket I added `PressableProps` in, but it was obviously wrong after reviewing. `Pressable` is a new Touchable component.

src/index.d.ts
```diff
@@ -22,17 +22,16 @@ import {
...
 import {
   IconButtonProps,
   IconProps as VectorIconProps,
 } from 'react-native-vector-icons/Icon';
 
-export interface TouchableComponent
-  extends TouchableHighlightProps,
-    PressableProps {}
+export interface TouchableComponent extends TouchableHighlightProps {
+  children?: React.ReactElement<any>;
```
SimenB, line 33: this gives a type error, unless there's only a single child (such as wrapping the children into `React.Fragment`). Just removing the `children` entry works.

(Happy you removed `PressableProps` though as we had to patch it (we're on RN 0.62 🙂 ))

SimenB: `children` is correctly inherited via https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6a47fa33adb7c55be63bbca45ce86d0ff4e59b18/types/react/index.d.ts#L501, so I don't think you need to specify this at all?

Issue #2523, opened 2020-08-27
ListItem typescript error with 2.3.0


ottob:
**Explain what you did (Required)**

Tried to follow the [upgrade guide for 2.3.0](https://react-native-elements.github.io/react-native-elements/blog/2020/08/10/2.3-upgrade-guide/). Im upgrading from 2.1.0.

Changed my old code:
```ts
<ListItem
    title="Edit message"
    leftIcon={{ name: 'edit' }}
    onPress={() => {
      setMenu(false)
      setEditing(true)
    }}
/>
```
to:
```ts
<ListItem
      onPress={() => {
        setMenu(false)
        setEditing(true)
      }}
        >
      <ListItem.Content>
        <Icon name="edit" />
        <ListItem.Title>Edit message</ListItem.Title>
      </ListItem.Content>
</ListItem>
```

**Expected behavior (Required)**

Successful typescript compile.

**Describe the bug (Required)**

Got a typescript error:

```
Property 'Content' does not exist on type 'typeof ListItem'.  TS2339

214 |           <ListItem.Content>
    |                     ^
```
I also get the same error for `ListItem.Title`:
```
Property 'Title' does not exist on type 'typeof ListItem'. TS2339
```

**To Reproduce (Required)**

We highly recommend that you re-create the bug on [Snack](https://snack.expo.io). If not, list the steps that a reviewer can take to reproduce the behaviour:

**Screenshots (Required)**

**Your Environment (Required):**

| software              | version |
| --------------------- | ------- |
| react-native-elements |  2.3.0       |
| react-native          |   0.63.2      |

SimenB:
I just added the stuff I needed myself, but we're applying this patch if it helps anybody get unblocked

```diff
diff --git a/node_modules/react-native-elements/src/index.d.ts b/node_modules/react-native-elements/src/index.d.ts
index 35fe68b..9099352 100644
--- a/node_modules/react-native-elements/src/index.d.ts
+++ b/node_modules/react-native-elements/src/index.d.ts
@@ -1199,7 +1197,10 @@ export interface ListItemProps extends TouchableComponent {
 /**
  * ListItem component
  */
-export class ListItem extends React.Component<ListItemProps, any> {}
+export class ListItem extends React.Component<ListItemProps, any> {
+  static Content: React.ComponentType<{right?: boolean}>;
+  static Title: React.ComponentType<TextProps & {right?: boolean, rightStyle?: StyleProp<ViewStyle>}>;
+}
[...]

PR #2527, opened 2020-08-28
fix new subcomponent types

flyingcircle:
Fixes: #2523

@ottob @SimenB please check this out and double check my work here before I submit it.


SimenB:
This works for me, thanks! 👍

MohammedFaragallah:
thank you for this it's a very important fix

some of these components have default props so it should be marked optional like `name` in ListItem.Chevron from `IconProps`

flyingcircle:
@MohammedFaragallah good catch! It actually is possible to overwrite the `name` prop. So I just changed it to being a `Partial` without excluding `name`.

PR #2556, opened 2020-09-04
Fix for: tap outside to close bottom sheet

snanthagopal:


src/bottomSheet/BottomSheet.js
```diff
@@ -16,34 +17,57 @@ class BottomSheet extends Component {
     super(props);
     this.state = {
       listHeight: undefined,
+      isVisible: this.props.isVisible,
     };
   }
 
+  componentDidUpdate(prevProps) {
```
flyingcircle, line 24: This function doesn't seem to do anything.

src/bottomSheet/BottomSheet.js
```diff
@@ -16,34 +17,57 @@ class BottomSheet extends Component {
...
-          <View style={styles.modalView}>
-            <View
-              style={([styles.listContainer], { maxHeight })}
-              onLayout={this.onLayout}
-            >
-              <ScrollView>{children}</ScrollView>
-            </View>
-          </View>
-        </SafeAreaView>
+        {isCancelable === true ? (
```
flyingcircle, line 64: Where is `isCancelable` coming from? You need to make updates to the docs if you're adding a new prop and an update to `index.d.ts`

flyingcircle (changes requested):

PR #2573, opened 2020-09-15
Fixes issue#2414 - Tooltip - Need KeyboardAvoidingView Support for tooltip

anandhakumarpalanisamy:
**Fixes issue#[2414](https://github.com/react-native-elements/react-native-elements/issues/2414) - Tooltip - Need KeyboardAvoidingView Support for tooltip**

- Added a configuration parameter "withKeyboardAvoidView" to the tooltip
- default value for "withKeyboardAvoidView" is false ->  keyboardAvoidView is not supported
- when set to true, KeyboardView is supported

**Working Demo :** 

keyboardavoid

### DIFF
--- a/src/tooltip/Tooltip.js
+++ b/src/tooltip/Tooltip.js
@@ -1,4 +1,4 @@
-import React, { Fragment } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import {
   TouchableOpacity,
@@ -6,6 +6,10 @@ import {
   View,
   StatusBar,
   I18nManager,
+  Platform,
+  ScrollView,
+  Dimensions,
+  KeyboardAvoidingView,
 } from 'react-native';
 
 import { withTheme } from '../config';
@@ -213,27 +217,35 @@ class Tooltip extends React.PureComponent {
   };
 
   renderStaticModalContent = () => {
-    const { withOverlay, overlayColor } = this.props;
+    const { withOverlay, overlayColor, skipAndroidStatusBar } = this.props;
 
     return (
-      <Fragment>
+      <View>
         <TouchableOpacity
-          style={styles.container(withOverlay, overlayColor)}
+          style={styles.container(
+            withOverlay,
+            overlayColor,
+            skipAndroidStatusBar
+          )}
           onPress={this.toggleTooltip}
           activeOpacity={1}
         />
         <View style={styles.closeOnlyOnBackdropPressViewWrapper}>
           {this.renderContent(true)}
         </View>
-      </Fragment>
+      </View>
     );
   };
   renderTogglingModalContent = () => {
-    const { withOverlay, overlayColor } = this.props;
+    const { withOverlay, overlayColor, skipAndroidStatusBar } = this.props;
 
     return (
       <TouchableOpacity
-        style={styles.container(withOverlay, overlayColor)}
+        style={styles.container(
+          withOverlay,
+          overlayColor,
+          skipAndroidStatusBar
+        )}
         onPress={this.toggleTooltip}
         activeOpacity={1}
       >
@@ -253,7 +265,12 @@ class Tooltip extends React.PureComponent {
 
   render() {
     const { isVisible } = this.state;
-    const { onClose, onOpen, ModalComponent } = this.props;
+    const {
+      onClose,
+      onOpen,
+      ModalComponent,
+      withKeyboardAvoidView,
+    } = this.props;
 
     return (
       <View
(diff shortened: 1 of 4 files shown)


src/tooltip/Tooltip.js
```diff
@@ -213,27 +217,35 @@ class Tooltip extends React.PureComponent {
   };
 
   renderStaticModalContent = () => {
-    const { withOverlay, overlayColor } = this.props;
+    const { withOverlay, overlayColor, skipAndroidStatusBar } = this.props;
 
     return (
-      <Fragment>
+      <View>
```
pranshuchittora, line 223: Can you please explain the reason for removing the `Fragment`

anandhakumarpalanisamy: @pranshuchittora  Thought of using native view instead of Fragment .  
If you feel  fragment has better advantage then view, let me know i will change it back to fragment.

pranshuchittora: In this scenario, you are not using `View` for any layout, it's been used as a container. The advantage of using `Fragment` is that it doesn't populate the actual UI. It gets omitted by the React reconcile while evaluating the tree. 
For more info -> https://reactjs.org/docs/fragments.html

src/tooltip/Tooltip.js
```diff
@@ -315,12 +348,17 @@ Tooltip.defaultProps = {
   skipAndroidStatusBar: false,
   ModalComponent: Modal,
   closeOnlyOnBackdropPress: false,
+  withKeyboardAvoidView: true,
```
pranshuchittora, line 351: IMO the default value should be `false` as the general use of `ToolTip` is to show info as compared to this specific use case where its been used to take the user input. 
Keeping it `false` will improve the performance by not adding those containers.

What's your view regarding this. Happy to discuss :)

src/tooltip/Tooltip.js
```diff
@@ -271,7 +288,22 @@ class Tooltip extends React.PureComponent {
...
           onRequestClose={onClose}
         >
-          {this.renderModalContent()}
+          {!withKeyboardAvoidView ? (
+            this.renderModalContent()
+          ) : (
+            <KeyboardAvoidingView
+              enabled={true}
+              keyboardVerticalOffset={StatusBar.currentHeight}
+              behavior={Platform.OS === 'android' ? undefined : 'position'}
```
pranshuchittora, line 297: I think the behaviour should be `behavior={Platform.OS == "ios" ? "padding" : "height"}` i.e. "height" for `Android`
https://reactnative.dev/docs/keyboardavoidingview#behavior

src/tooltip/Tooltip.js
```diff
@@ -271,7 +288,22 @@ class Tooltip extends React.PureComponent {
...
+            >
+              <ScrollView
```
pranshuchittora, line 299: The reason for using `ScrollView` ?

pranshuchittora (changes requested):

src/tooltip/Tooltip.js
anandhakumarpalanisamy: @pranshuchittora  Thank you very much for the suggestion . I have changed it back to Fragment :+1:

src/tooltip/Tooltip.js
anandhakumarpalanisamy: @pranshuchittora Thank you very much for the suggestion.
you are correct . "Keeping it false will improve the performance by not adding those containers." :+1: 
I have now changed the default value to "false"

src/tooltip/Tooltip.js
anandhakumarpalanisamy: @pranshuchittora  : KeyboardAvoidView support in Modals was kind of little bit tricky . The default example syntax from the react-native docs was not working when used inside Modals as it is.

I tried multiple suggestions from the internet.
The solution which worked for our case was :  https://github.com/react-native-community/react-native-modal/issues/32#issuecomment-562524623

src/tooltip/Tooltip.js
anandhakumarpalanisamy: @pranshuchittora  : KeyboardAvoidView support in Modals was kind of little bit tricky . The default example syntax from the react-native docs was not working when used inside Modals as it is.

I tried multiple suggestions from the internet.
The solution which worked for our case was :  https://github.com/react-native-community/react-native-modal/issues/32#issuecomment-562524623

react-native's keyboardAvoidView implementation handles four cases "'height', 'position', 'padding,default" : 
https://github.com/facebook/react-native/blob/master/Libraries/Components/Keyboard/KeyboardAvoidingView.js

passing the behaviour as "undefined" will trigger out the default case in keyboardAvoidView
As for as android is concerned,  "default" setting  was working fine for our case.

anandhakumarpalanisamy:
**Regarding test coverage :**  

I was not sure how to simulate the keyboard input event and check if the view is moving  when virtualkeyboard is overlapping with it . 

As of now, I have added test coverage to just check the default tooltip behaviour remains unaffected, when enabling or disabling "KeyboardAvoidView" support.

If anybody has any example on that , kindly  let me know .

PR #2587, opened 2020-09-18
fix(Button): Improved UX of the LinearGradient button

### LINKED ISSUE #2493: Gradient Button doesn't show visual onPress like regular button
**Explain what you did (Required)**

Created a gradient button. See example here:

```
import LinearGradient from 'react-native-web-linear-gradient'
import React from 'react'

<Button
  title={'TITLE'}
  onPress={onPress}
  ViewComponent={LinearGradient}
  linearGradientProps={{
    colors: ['#FFF', '#333'],
    start: { x: 0, y: 0 },
    end: { x: 0, y: 1 },
  }}
/>
```

**Expected behavior (Required)**

When pressing the button onPress is fired, but the user does not get any tactile feedback. See image. The Strava button works fine, but the Continue button does not give any feedback to the user whatsoever.

**Describe the bug (Required)**

It appears that once you add linearGradientProps the button does no longer provide feedback to the user.

**To Reproduce (Required)**
1. yarn add react-native-web-linear-gradient
2. copy above code in your project
3. Click the button

**Screenshots (Required)**

image

**Your Environment (Required):**

| react-native-elements |  ^0.18.0 |
| react-native          | 0.62.2 |

pranshuchittora:
Closes #2493 

If the button has linear-gradient prop then it will render `TouchableOpacity` in Android instead of `TouchableNativeFeedback` in order to provide the user with feedback.

Not using `Pressable` because of the following reasons:
- Very New.
- Needs `@latest` version of React Native, therefore won't be possible for all users to upgrade RN.
- As mentioned above, that it's very new so API may change in future to make it more stable.

### DIFF
--- a/src/buttons/Button.js
+++ b/src/buttons/Button.js
@@ -67,14 +67,14 @@ class Button extends Component {
     } = this.props;
 
     // Refactor to Pressable
-    const TouchableComponentInternal = TouchableComponent
-      ? TouchableComponent
-      : Platform.select({
-          android: linearGradientProps
-            ? TouchableOpacity
-            : TouchableNativeFeedback,
-          default: TouchableOpacity,
-        });
+    const TouchableComponentInternal =
+      TouchableComponent ||
+      Platform.select({
+        android: linearGradientProps
+          ? TouchableOpacity
+          : TouchableNativeFeedback,
+        default: TouchableOpacity,
+      });
 
     const titleStyle = StyleSheet.flatten([
       styles.title(type, theme),
--- a/src/buttons/Button.js
+++ b/src/buttons/Button.js
@@ -66,6 +66,16 @@ class Button extends Component {
       ...attributes
     } = this.props;
 
+    // Refactor to Pressable
+    const TouchableComponentInternal = TouchableComponent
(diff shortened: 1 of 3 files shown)


src/buttons/Button.js
```diff
@@ -66,6 +66,16 @@ class Button extends Component {
       ...attributes
     } = this.props;
 
+    // Refactor to Pressable
+    const TouchableComponentInternal = TouchableComponent
```
flyingcircle, line 70: I think this can be changed to:

```js
const TouchableComponentInternal = TouchableComponent || 
  Platform.select(...);
```

flyingcircle (changes requested):

src/buttons/Button.js
pranshuchittora: @flyingcircle  thanks for the feedback :)

flyingcircle (changes requested):
I forgot about this on the first review. But since this is changing the default value, then a note needs to be left on the Prop in the docs when the new default happens.

flyingcircle (approved):

Issue #2478, opened 2020-08-04
Chevron defaults to right arrow after upgrading to react-native-vector-icons 6.7 (iOS)


NathanNovak:
**Explain what you did**
After upgrading to 2.1.0 Chevron defaults to a right arrow on iOS. 

**Expected behavior**

Chevron remain as a chevron unless otherwise specified.

**Describe the bug**

After upgrading to 2.1.0 Chevron defaults to a right arrow on iOS.  did not affect Android.

BEFORE upgrade would display correctly.
`<ListItem 
     ...  
     chevron >`

AFTER upgrade I have to use icon library.
` <ListItem
          ...
          chevron={{
            name: 'right',
            type: 'antdesign',
            color: themes.LIST_CHEVRON_COLOR,
            size: 16,
          }}
        /> `

**Your Environment (please complete the following information):**

| software              | version |
| --------------------- | ------- |
| react-native-elements | 2.1.0 |
| react-native          |  0.6.3 |

flyingcircle:
You can double check the changes, but there were no changes made to `ListItem` between `2.0.4` and `2.1.0`. What version did you upgrade from and what version of icons are you using?

NathanNovak:
@flyingcircle We looked into Vector Icons and Ionicons more.  The code in RNE has the chevron for iOS as `arrow-forward`:  

https://github.com/react-native-elements/react-native-elements/blob/aed8d12d9eed90b8d0b9ec4863a3ad264f733562/src/list/ListItem.js#L27

In the new version of Ionicons (5.0.0) they renamed the arrow icon to `arrow-forward` and since Vector Icons is using the newer version of Ionicons the name `arrow-forward` is not the chevron anymore on iOS.  The correct new name for the ionicon chevron is `chevron-forward`.

**Vector Icons Package.json**
https://github.com/oblador/react-native-vector-icons/blob/32bbd538794a5835de685f54bf2dd6c053ede54c/package.json#L85

**Ionicons Change Log**
https://github.com/ionic-team/ionicons/blob/master/CHANGELOG.md

flyingcircle:
So then this has nothing to do with upgrading RNE to version `2.1.0`, but with you probably also updating `react-native-vector-icons` from `6.6.0` to >=`6.7.0`.

I'm open to updating the default if `chevron-forward` is backwards compatible with `6.6.0`, otherwise that change will have to wait for a major update since it will be changing the peer dependency requirement.

[...]

Issue #2423, opened 2020-06-29
Upgrade to 2.0.3 breaks IconButtonProps import


daviddelusenet:
When upgrading the `react-native-elements` dependency to `2.0.3` the following import doesn't work anymore in `index.d.ts`:

```tsx
import { IconButtonProps } from 'react-native-vector-icons/Icon';
```

This is causing TypeScrip errors in my custom theme since I'm overwriting the color of icons.

Screenshot 2.0.2:
<img width="568" alt="Screenshot 2020-06-29 at 09 10 36" src="https://user-images.githubusercontent.com/1398418/85983791-69fd9880-b9e8-11ea-8820-3e87207f8acd.png">

Screenshot 2.0.3:
<img width="1135" alt="Screenshot 2020-06-29 at 09 10 59" src="https://user-images.githubusercontent.com/1398418/85983823-784bb480-b9e8-11ea-84ac-2ade59f27349.png">

siddiboy:
The recent upgrade is breaking every react-native-vector-icons import. Below is a screenshot of the lock file:
![image](https://user-images.githubusercontent.com/20492995/86050936-38311400-ba72-11ea-92ff-2be5e9d570f5.png)

seems like the dependency of @types/react-native-vector-icons which was there for react-native-elements 2.0.0 but is no longer there.

flyingcircle:
Fixed with 2.0.4

PR #2596, opened 2020-09-22
Icons update

flyingcircle:
Fixes: #2478


package.json
```diff
@@ -33,7 +33,7 @@
     "logo": "https://opencollective.com/react-native-elements/logo.txt"
   },
   "dependencies": {
-    "@types/react-native-vector-icons": "^6.4.5",
+    "@types/react-native-vector-icons": "^6.4.6",
```
pranshuchittora, line 36: I think we should move this to `devDeps`

package.json
```diff
@@ -62,13 +62,13 @@
...
     "react-native-testing-library": "^2.1.0",
-    "react-native-vector-icons": "^6.6.0",
+    "react-native-vector-icons": "^7.0.0",
     "react-test-renderer": "^16.13.1",
     "rimraf": "^3.0.2",
     "typescript": "^3.9.5"
   },
   "peerDependencies": {
-    "react-native-vector-icons": ">6.6.0"
+    "react-native-vector-icons": ">7.0.0"
```
pranshuchittora, line 71: Let's add `react` too.

package.json
flyingcircle: I attempted this before. However, #2423 happened when I did this. I'm sure there's a better answer, but this may need to be handled in a separate ticket.

package.json
flyingcircle: I guess we can, though that's a little outside the scope of this PR.

hengkx:
any progress?

flyingcircle:
@hengkx you can see the comment from the maintainer of the expo ticket I linked. He plans to release the new version in December, so we'll plan to finish the v3 release then and merge in this ticket. He mentions using the fork, but that just isn't a viable solution for our library.

PR #2629, opened 2020-10-06
Docs: arrange props alphabetically in searchbar.md, listitem.md 

sanskritip:
Fixed the prop list for an easy understanding of the docs.


pranshuchittora (approved):
Thanks @sanskritip

PR #2641, opened 2020-10-15
Adds missing onFocus documentation for searchbar

chiraggshah:


flyingcircle:
The docs already mention that `Searchbar` includes all `TextInput` props: https://reactnative.dev/docs/textinput.html

PR #2645, opened 2020-10-18
avatar not showing placeholder if not source

jehielmartinez:
https://github.com/react-native-elements/react-native-elements/issues/2592

Fixed issue by adding a condition if `source.uri` doesn't exists
Added test for coverage

### DIFF
--- a/src/avatar/Avatar.js
+++ b/src/avatar/Avatar.js
@@ -107,7 +107,7 @@ const AvatarComponent = ({
       <Image
         placeholderStyle={StyleSheet.flatten([
           placeholderStyle,
-          hidePlaceholder && { backgroundColor: 'transparent' },
+          hidePlaceholder && styles.hiddenPlaceholderStyle,
         ])}
         PlaceholderContent={PlaceholderContent}
         containerStyle={imageContainerStyle}
@@ -142,6 +142,9 @@ const styles = StyleSheet.create({
     color: '#ffffff',
     backgroundColor: 'transparent',
     textAlign: 'center',
+  },
+  hiddenPlaceholderStyle: {
+    backgroundColor: 'transparent',
   },
 });
 
--- a/src/avatar/Avatar.js
+++ b/src/avatar/Avatar.js
@@ -78,7 +78,7 @@ const AvatarComponent = ({
     ));
 
   // Remove placeholder styling if we're not using image
-  const hidePlaceholder = source && source.uri;
+  const hidePlaceholder = !(source && source.uri);
 
   // Merge image container style
   const imageContainerStyle = StyleSheet.flatten([
@@ -107,7 +107,7 @@ const AvatarComponent = ({
       <Image
         placeholderStyle={StyleSheet.flatten([
           placeholderStyle,
-          !hidePlaceholder && styles.hiddenPlaceholderStyle,
+          hidePlaceholder && styles.hiddenPlaceholderStyle,
         ])}
         PlaceholderContent={PlaceholderContent}
         containerStyle={imageContainerStyle}
--- a/src/avatar/Avatar.js
+++ b/src/avatar/Avatar.js
@@ -78,7 +78,7 @@ const AvatarComponent = ({
     ));
 
   // Remove placeholder styling if we're not using image
-  const hidePlaceholder = !source;
+  const hidePlaceholder = !source || !source.uri;
 
   // Merge image container style
   const imageContainerStyle = StyleSheet.flatten([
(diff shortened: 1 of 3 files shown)


src/avatar/Avatar.js
```diff
@@ -78,7 +78,7 @@ const AvatarComponent = ({
     ));
 
   // Remove placeholder styling if we're not using image
-  const hidePlaceholder = !source;
+  const hidePlaceholder = !source || !source.uri;
```
pranshuchittora, line 81: I think using `&& ` instead of `||` would be a bit optimized as the engine will check the first argument and if it's false then it won't check for the second one (right of the operator).

src/avatar/__tests__/Avatar.js
```diff
@@ -204,5 +204,19 @@ describe('Avatar Component', () => {
...
+      const component = shallow(
+        <Avatar
+          size="medium"
+          overlayContainerStyle={{ backgroundColor: 'pink' }}
+          title="MD"
+          source={{}}
+        />
+      );
+
+      expect(component.props().style.backgroundColor).toBe('transparent');
```
pranshuchittora, line 218: Can you refactor the inline styling ` hidePlaceholder && { backgroundColor: 'transparent' },` to the stylesheet and then reference that style object here.

Why?
The styling may change might cause this test to fail. A better way is to use references to the style object directly.

pranshuchittora (changes requested):
Minor Changes :)

src/avatar/Avatar.js
jehielmartinez: Noted, but if `source` doesn't contain `uri`, the app crashes because it can't found the property `uri` to compare. That's why I set the `||` instead of `&&`. Let me know what you think.

src/avatar/__tests__/Avatar.js
jehielmartinez: Sure!

src/avatar/Avatar.js
pranshuchittora: Hi Jehiel, as mentioned in my previous comment that all JS engines come with a small hack/shortcut for optimizing such expressions.

```js
const x = null;

if (x && x.y) {
  console.log("x is !null and x.y as well");
} else {
  console.log("x is empty");
}

```

Just run the above snippet in Browser or Node.js and it won't fail because `x.y` will never be checked.

pranshuchittora: I recommend that you test the following changes that this PR brings locally as well. You can use this app -> https://github.com/react-native-elements/react-native-elements-app
and link the local project using `npm link`. :)

jehielmartinez: @pranshuchittora this is what I'm getting when changing the `||` to `&&` when testing
```
Avatar Component › Placeholders › shouldn't show placeholder if not using source

    TypeError: Cannot read property 'uri' of undefined

      79 | 
      80 |   // Remove placeholder styling if we're not using image
    > 81 |   const hidePlaceholder = !source && !source.uri;
```

Any help how to fix this?

pranshuchittora: Hi so as mentioned this works for expressions , so instead try this.

`  const hidePlaceholder = (source && source.uri) ;`

jehielmartinez:
@pranshuchittora can you check it again? Thank you

src/avatar/__tests__/Avatar.js
```diff
@@ -204,5 +204,20 @@ describe('Avatar Component', () => {
...
+          title="MD"
+          source={{}}
+        />
+      );
+      expect(
+        component.children().props().placeholderStyle.backgroundColor
+      ).toBe('transparent');
+      expect(toJson(component)).toMatchSnapshot();
+    });
   });
```
pranshuchittora, line 222: Hi @jehielmartinez this looks great. Can you pls add the test case when source.uri is undefined/.

`source = {{  uri : undefined }}`

pranshuchittora (changes requested):
LGTM, just one more edge case needs to be covered in the test suite :)

src/avatar/__tests__/Avatar.js
jehielmartinez: sure!

jehielmartinez:
@pranshuchittora test for `uri=undefined` added

pranshuchittora (approved):
LGTM

flyingcircle:
@jehielmartinez contact me on the RNE slack for a small token thank you for this PR.

PR #2650, opened 2020-10-25
fix(ThemeProvider): Custom theme lost on dark mode switch

vividh:
Fixes https://github.com/react-native-elements/react-native-elements/issues/2647.

Changelog:
- [x] Fix custom theme overwrite on dark mode switch
- [x] Add unit test

### DIFF
--- a/src/config/ThemeProvider.js
+++ b/src/config/ThemeProvider.js
@@ -34,9 +34,15 @@ export default class ThemeProvider extends React.Component {
     if (useDark !== state.useDark) {
       const defaultColors = useDark ? darkColors : colors;
       return {
-        theme: deepmerge(state.theme, {
-          colors: defaultColors,
-        }),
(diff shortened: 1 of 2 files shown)


src/config/ThemeProvider.js
```diff
@@ -34,9 +34,15 @@ export default class ThemeProvider extends React.Component {
     if (useDark !== state.useDark) {
       const defaultColors = useDark ? darkColors : colors;
       return {
-        theme: deepmerge(state.theme, {
-          colors: defaultColors,
-        }),
+        theme: deepmerge(
+          state.theme,
```
IlyaSemenov, line 38: Why is `state.theme` needed here? I think it's safer to omit it (merge default colors + props.theme).

flyingcircle:
@vividh contact me on the RNE slack for a small thank you token for this commit.

PR #2656, opened 2020-10-30
Add bottom sheet containerStyle prop

nopol10:
The bottom sheet has an underlay that currently cannot be customized. This adds the containerStyle prop that is present in other components so that style customization on the underlay is possible.


flyingcircle:
@nopol10 contact me on the RNE slack for a small token thank you for this PR.

PR #2658, opened 2020-10-30
Update input.md

levkovich2806:
Change links


flyingcircle:
@levkovich2806 can you update those urls globally throughout the whole project? I would prefer it happen all at once rather than in a bunch of small PR's.

flyingcircle (changes requested):
This needs to be applied more globally rather than just a single version of a single component.

pranshuchittora (changes requested):
Hi, Thanks for the changes. Can you please change the docs for all the versions.
/website/docs & versioned docs

Issue #2659, opened 2020-10-31
various style props using incorrect PropsType validation


carlost:
**Explain what you did (Required)**

Used a compiled react-native StyleSheet object to style a react-native-elements Button component.

**Expected behavior (Required)**

The Button component uses the stylesheet to control how it was displayed ... and it doesn't complain about it.

**Describe the bug (Required)**

The style prop was used by the component.  However, it logged a prop-type violation message:

> checkPropTypes.js:19 Warning: Failed prop type: Invalid prop `titleStyle` supplied to `Button`.

This occurs with just about any style prop i looked at in react-native-elements.

These error messages make me itchy!

**To Reproduce (Required)**

https://snack.expo.io/Zh6z6hZCT

This is a slight head scratcher ... I do not see the error message in snack.  I assume it is because the runtime checks performed by prop-types only occur in the dev env.

**Screenshots (Required)**

NA

**Your Environment (Required):**

| software              | version |
| --------------------- | ------- |
| react-native-elements |   v2.0.3      |
| react-native          |    v0.63.3     |
| react-native-web | v0.13.18|

I suspect that I am missing something here.  The type def consistently used in RNE for style props is:

```javascript
PropTypes.oneOfType([PropTypes.object, PropTypes.array])
```

It is my understanding that `StyleSheets.create` returns an object that has numbers as values.  RN basically memoizes these stylesheets for performance purposes. So if your `StyleSheet` looks like this:

```javascript
const styles = StyleSheet.create({
  button: {
    backgroundColor: 'grey'
  },
  buttonText: {
    color: 'red',
    fontSize: 18
  }
});
```

a call to `styles.button` will return a `number` and not an `array` or `object`.  You can wrap these numbers in arrays to bypass the prop-types checks or call `StyleSheet.flatten` to pass in the actual styles object instead of the number ID. However, those solutions are either confusing or a performance sinkhole.

If i am correct ... i am not entirely sure what the appropriate solution is.  

I punish myself with Flow, so I am not familiar with prop-types best practices.  RN has moved the various `StyleSheets` props into files named `react-native/Libraries/DeprecatedPropTypes/DeprecatedXXStylePropTypes.js`... so i am not sure if using those files is appropriate.

I suspect that widening the validation rule for the style props to:

```javascript
PropTypes.oneOfType([PropTypes.object, PropTypes.array, PropTypes.number])
```

isn't very useful because any ole number won't do ... it needs to be a StyleSheet ID.

[...]

PR #2660, opened 2020-10-31
Proper style props to Button

Karthik-B-06:
Added proper style props so that when using the component we will get suggestions. 

Fixing #2659 .

```
  titleStyle: ViewPropTypes.style,
  titleProps: Text.propTypes.style,
```

Screenshot 2020-10-31 at 2 21 37 PM
Screenshot 2020-10-31 at 2 22 08 PM

If this works we can go ahead adding it to other components.

Tests passing: 
Screenshot 2020-10-31 at 2 28 41 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "logo": "https://opencollective.com/react-native-elements/logo.txt"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


flyingcircle:
`ViewPropTypes` was removed in newer versions of react native. That is why this was removed.

PR #2667, opened 2020-11-04
fix(image): require assets/image doesnt work

infothien:
**Issue:**  it will never display the require asset/image (aim to fix this issue https://github.com/react-native-elements/react-native-elements/issues/2663 )

**Example:**
The code will looks sth like this
Welcome_tsx_—_dentsu

Currently, the result will be:
iPhone_8_Plus_–_14_1_and_iPhone_11_–_14_1

So the expectation will be:
Screenshot_2020_11_04_16_41


src/image/Image.js
```diff
@@ -51,7 +51,8 @@ class Image extends React.Component {
     } = this.props;
 
     const hasImage =
-      Boolean(attributes.source) && Boolean(attributes.source.uri);
+      typeof attributes.source === 'number' ||
```
infothien, line 54: here is checking for the case `ImageRequireSource` of Image from react-native

infothien:
I only change the code for Image.js, other changes come from `yarn prettify` and `yarn install`

infothien:
@flyingcircle , @pranshuchittora could you guys please give it a check !

src/image/Image.js
pranshuchittora: Can you explain the type `number` used here?
Source of the image can be a string || object :thinking:

pranshuchittora:
Pls refer this -> https://github.com/react-native-elements/react-native-elements/issues/2663#issuecomment-721852353

src/image/Image.js
infothien: Yes, I can @pranshuchittora. So in react native source will have type `ImageSourcePropType` like the image above
index_d_ts_—_dentsu

and this `ImageSourcePropType` will be a union of `ImageURISource | ImageURISource[] | ImageRequireSource;`
index_d_ts_—_dentsu

and `ImageRequireSource` is number. 
index_d_ts_—_dentsu

Since the source type comes from react-native offical so I think we should also support for ImageRequireSource otherwise, it seems like we are just dropping a part from it

infothien: @pranshuchittora could you please check this out to see does it make sense to you

pranshuchittora: Hi, I just closed the issue related to this PR. I think there's no issue at all, the usage was wrong pls refer the closing comment. :)

infothien: uhm, I am quite sad that you say that, since all the images I put above is clear enough for this case. The `source` for `Image` in `react-native-elements` got the type `ImageRequireSource` but seems like now we just dont allow user to use it anymore.

PR #2688, opened 2020-11-26
Simplify getIconStyle util return

reinaldosimoes:
Aggregates the several cases as they all return an empty object.

### DIFF
--- a/src/helpers/getIconStyle.js
+++ b/src/helpers/getIconStyle.js
@@ -1,37 +1,25 @@
 export default (type, extraProps) => {
   switch (type) {
     case 'zocial':
-      return {};
     case 'octicon':
-      return {};
     case 'material':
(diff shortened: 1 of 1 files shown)


pranshuchittora (commented):
I understand what you are trying to achieve with the following fix. This may make the app a bit slow.

Let me explain 
If the `type` is `zocial` then with the current code it will execute only the first case and performs early return.

But in the case of the changes proposed, the conditions will execute for each case after the matching case this approach is 

1. Slow
2. Not easy to infer easily (not readable)

reinaldosimoes:
@pranshuchittora fair points, it was less of an effort to optimize it and more of a streamlining the returns just for the sake of making sure they all return the expected. I assume adding unit tests might be better for this case in specific.

PR #2689, opened 2020-11-27
Avatar (only with title) doesn't displayed without source attribute

SamilMehdiyev:
bug fix related displaying avatar when source is not defined

### DIFF
--- a/src/avatar/Avatar.js
+++ b/src/avatar/Avatar.js
@@ -78,7 +78,7 @@ const AvatarComponent = ({
     ));
 
   // Remove placeholder styling if we're not using image
-  const hidePlaceholder = source != null ? true : false;
+  const hidePlaceholder = source !== null ? true : false;
 
   // Merge image container style
   const imageContainerStyle = StyleSheet.flatten([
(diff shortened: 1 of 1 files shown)


src/avatar/Avatar.js
```diff
@@ -78,7 +78,7 @@ const AvatarComponent = ({
     ));
 
   // Remove placeholder styling if we're not using image
-  const hidePlaceholder = !source;
+  const hidePlaceholder = source != null ? true : false;
```
reinaldosimoes, line 81:
```suggestion
  const hidePlaceholder = source !== null ? true : false;
```

reinaldosimoes:
Duplicate of https://github.com/react-native-elements/react-native-elements/pull/2645

src/avatar/Avatar.js
```diff
@@ -78,7 +78,7 @@ const AvatarComponent = ({
     ));
 
   // Remove placeholder styling if we're not using image
-  const hidePlaceholder = !source;
+  const hidePlaceholder = source !== null ? true : false;
```
pranshuchittora, line 81: `source` is an object and it must have a key of `uri` as well

PR #2695, opened 2020-12-10
fix(CheckBoxIcon): it always re-render even nothing change

daminhtung:
It perceives that `{ minWidth: size || 24 }` is different from `{ minWidth: size || 24 }` in style.
So `{ minWidth: size || 24 }`  is recognized as a new object. 
And then it always re-renders even though nothing has changed.

### DIFF
--- a/src/checkbox/CheckBoxIcon.js
+++ b/src/checkbox/CheckBoxIcon.js
@@ -25,12 +25,14 @@ const CheckBoxIcon = ({
 
   const VectorIcon = iconType ? getIconType(iconType) : FAIcon;
 
+  const vectorIconStyle = { minWidth: size || 24 }
+
   return (
     <VectorIcon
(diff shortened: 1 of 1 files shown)


pranshuchittora:
> It perceives that `{ minWidth: size || 24 }` is different from `{ minWidth: size || 24 }` in style.
> So `{ minWidth: size || 24 }` is recognized as a new object.
> And then it always re-renders even though nothing has changed.

React re-renders only when state or prop changes :question: 
IMO this is not related to any sort of re-rendering

pranshuchittora:
Apart from that react creates such an object once and used until the component unmounts, as objects in JS are reference type. So there won't be any memory overhead. 

`size` is a prop and changing prop must initiate a re-render call. That's the intended behaviour.

flyingcircle:
I agree with @pranshuchittora doesn't look like this will do anything.

PR #2706, opened 2020-12-15
fix: Header Text Alignment - iOS

pranshuchittora:


flyingcircle:
I tried this out in the example app and it causes the left icon spacing to get off. I'm wondering if we need to somehow include the `SafeAreaView` into the component? Still looking at this myself. Workaround though is to not include the `Header` in a `SafeAreaView`

flyingcircle (changes requested):
I think we need additional changes here:

I wrote this out in a sample file that I think does what we would want it to do:
```
<>
      <StatusBar barStyle={barStyle} translucent={true} {...statusBarProps} />
      <SafeAreaView style={styles.statusBar(theme)} />
      <SafeAreaView>
      <ViewComponent
      ...
```

```
const styles = {
  statusBar: (theme) => ({
    flex: 0,
    backgroundColor: theme.colors.primary,
  }),
```

src/header/Header.js
```diff
@@ -179,7 +179,10 @@ const styles = {
     borderBottomWidth: StyleSheet.hairlineWidth,
     paddingHorizontal: 10,
     backgroundColor: theme.colors.primary,
-    paddingTop: getStatusBarHeight(),
+    paddingTop: Platform.select({
```
flyingcircle, line 182: This change also needs to be made to the `height` style as well.

flyingcircle (changes requested):

flyingcircle:
I actually found a few additional items that needed changing. I think I'll just open my own PR

pranshuchittora:
Sure

Issue #2704, opened 2020-12-14
Header Text Alignment - IOS


natalyayyad:
I'm using the Header component, on Android its centered. However, On IOS seems like there's a top padding which I didn't add.
![image](https://user-images.githubusercontent.com/47296228/102147643-be4fe400-3e73-11eb-9837-166e0a7c5fd0.png)
![image](https://user-images.githubusercontent.com/47296228/102147680-cdcf2d00-3e73-11eb-91c4-995fb817f0aa.png)

code:
` <Header
        centerComponent={{
          text:'Header',
          style: {
            color: Theme.COLORS.TEXT,
            fontWeight: 'bold',
            fontSize: Theme.FONT.BIG,
          },
        }}
        containerStyle={{
          backgroundColor: Theme.COLORS.SECONDARY,
        }}
      />`

pranshuchittora:
Can you provide a snack demo to reproduce the same

natalyayyad:
Yes, here is the link:
[https://snack.expo.io/@natalyayyad/header-react-native-elements](url)

pranshuchittora:
Hi @natalyayyad thanks for the demo. I am able to repro this.
Raised the PR for the fix :)

flyingcircle:
I'm not sure that the issue is iOS here. You put the `Header` inside of a `SafeAreaView` component which is only applicable on iOS. If you put the `Header` outside of the `SafeAreaView`. I'll stick the rest of my comments in the PR.

natalyayyad:
@pranshuchittora Thank you. @flyingcircle  I tried removing SafeAreaView, or replace it with View but the issue remains. 
What's the default height of the header seems like it's 60, right?

otech47:
this is happening because the SafeAreaView is forced on all of the inner left/center/right `<Children>` components here: https://github.com/react-native-elements/react-native-elements/blob/next/packages/base/src/Header/Header.tsx#L149

would be nice to be able to pass down props to override the `edges` and `style` props on that inner container. it is just inside of the `containerStyle` so tweaking that doesn't help

![Screenshot 2023-03-09 at 2 16 49 PM](https://user-images.githubusercontent.com/4914611/224131192-1d218aa1-e172-446c-99d1-5cb1a369c93b.png)

otech47:
actually it seems like the ability to customize those props will soon be in an upcoming release: https://github.com/react-native-elements/react-native-elements/pull/3741

Issue #1791, opened 2019-02-19
Header height based on Orientation


kumaresan-cgvak:
**Is your feature request related to a problem? Please Describe.**
Header height is not changed according to the orientation of mobile device. So, header occupies more space in landscape mode.

**Describe the solution you'd like**
Header height can be reduced in Landscape mode of a mobile device.

dquessenberry:
Not sure why this issue was closed as it is a legit problem. I am able to reproduce the issue and I've looked through the Header source code, and it does not resize the header height for "notched" iPhones appropriately on orientation change. If the app is refreshed in landscape it does render the Header height appropriately, so I assume the Header component just needs an orientation change listener and to resize the height accordingly from portrait to landscape and vice versa.

In the meantime, I am writing a wrapper component to handle this, but it would be nice if the Header component did this automatically.

flyingcircle:
@dquessenberry please feel free to open a PR to fix this.

flyingcircle:
It looks like it's not an easy problem to solve. Might have to wait on a `react-native-status-bar-height` v3 https://github.com/ovr/react-native-status-bar-height/issues/13

PR #2707, opened 2020-12-15
update header component to handle more phone shapes/sizes

flyingcircle:
Fixes: #2704 and #1791

`Header` still needs to be outside of the `SafeAreaView`, but this integrates `SafeAreaView` into the component itself so that the left and right components have better placement in landscape orientation.

### DIFF
--- a/src/header/Header.js
+++ b/src/header/Header.js
@@ -87,7 +87,12 @@ class Header extends Component {
     return (
       <>
         <StatusBar barStyle={barStyle} translucent={true} {...statusBarProps} />
-        <SafeAreaView style={styles.statusBar(theme)} />
+        <SafeAreaView
+          style={StyleSheet.flatten([
+            styles.statusBar(theme),
+            backgroundColor && { backgroundColor },
+          ])}
+        />
         <ViewComponent
           testID="headerContainer"
           {...attributes}
--- a/src/header/Header.js
+++ b/src/header/Header.js
@@ -185,9 +185,6 @@ Header.defaultProps = {
 };
 
 const styles = {
-  statusBar: (theme) => ({
-    backgroundColor: theme.colors.primary,
-  }),
   container: (theme) => ({
     borderBottomColor: '#f2f2f2',
     borderBottomWidth: StyleSheet.hairlineWidth,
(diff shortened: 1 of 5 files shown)


flyingcircle:
@pranshuchittora and @natalyayyad check this out and see if this works. `Header` still needs to be outside of a `SafeAreaView` because the `Header` component draws itself into the unsafe areas.

src/header/Header.js
```diff
@@ -84,57 +85,66 @@ class Header extends Component {
...
-          containerStyle,
-        ])}
-        source={backgroundImage}
-        imageStyle={backgroundImageStyle}
-        {...linearGradientProps}
-      >
+      <>
         <StatusBar barStyle={barStyle} translucent={true} {...statusBarProps} />
-        <Children
+        <SafeAreaView
```
pranshuchittora, line 90: I think wrapping a reusable component with `SafeAreaView` is not recommended. It must be consumed from the user end. If someone wants to deliberately overflow the Header then, that behaviour can't be achieved.

flyingcircle: This specific one at least doesn't wrap anything. It has a flex of 0 and is used to work around some issues.
However, without the `SafeAreaView` below wrapping the `Children`, it is almost impossible to get the orientation working as one would expect. The left and right components end up showing up in the curves of the iPhone instead of in the safe area where you'd expect them. (Try removing it and trying it out in the demo app, and see where the left and right components end up)
I'm inclined to think that this covers 99% of use cases and the 1% that want to "overflow" the `Header` will likely be making their own component anyways.

pranshuchittora: I agree. My only concern is that we try to minimize the render tree. People generally use libraries like `react-navigation` which comes with default wrapping of SFA.
https://reactnavigation.org/docs/handling-safe-area/

1. IMO, specifying this workaround in the docs.
2. Or we can add a boolean prop `safeArea` which internally wraps iOS in SafeAreaView and Android in paddingTop status bar height

pranshuchittora: Or a new component `SafeArea ` which does this job. So that it can be used for non-header usage as well. Like apps with bottom nav etc.

flyingcircle: Alright! I replaced the `SafeAreaView` with a different library. This means that now everything shows up as expected. And Android and iOS look and behave the exact same!

Issue #2714, opened 2020-12-18
After upgrade to v3.0.0 all Image components turned to grey rectangulars 


Aparus:
Hi! Thank you for v3! 

After upgrade from v2.3.2 to v3.0.0, all Images (imported from RNE) turned to grey rectangulars. 

Here is my code:

```jsx
<Image
	source={image}
	style={[
		{ width: 100, height: 100, resizeMode: 'contain' },
		contentTypeStyle.image
	]}
/>
```

Importing Image from `react-native` instead of `react-native-elements` fixed the issue:  

import { Image } from 'react-native ~~-elements~~'

Honestly I don't know why I need RNE Image, but I just inform you. May be it will be a good bug report? 

**Your Environment (Required):**

| software              | version |
| --------------------- | ------- |
| react-native-elements |   3.0.0      |
| react-native          |   0.63.4      |
| expo |   40.0.0, web |

flyingcircle:
RNE `Image` has an animated transition for when an Image loads so it can provide a better aesthetic. I noticed this happening when I transitioned to Expo 40, but it only happened in the emulator environment. Is this happening still when you run it on a real device?

Aparus:
> RNE `Image` has an animated transition for when an Image loads so it can provide a better aesthetic. I noticed this happening when I transitioned to Expo 40, but it only happened in the emulator environment. Is this happening still when you run it on a real device?

Thank you for explanation about Image! 
Expo 40! I upgraded to it before RNE upgrade. 
My laptop with browser is a real device )) But, no. I used to use Expo client, but because of some issues (String.matchAll(regex) doesn't work in React-Native) and I didn't fix it yet...

flyingcircle:
The `Image` is still working in the Playground: https://react-native-elements.js.org/#/image . So I don't really have a way of reproducing what you're describing here.

Aparus:
> The `Image` is still working in the Playground: https://react-native-elements.js.org/#/image . So I don't really have a way of reproducing what you're describing here.

sorry , i didn't mention that I use `<Image source={require(file/local/path/)} /> `. I don't know how to reproduce it in playground. But on my laptop (expo  web), {uri} works, but {require} doesn't work. Here is an example with your image from playground: 

![localhost_19006_(iPhone 6_7_8)](https://user-images.githubusercontent.com/1222611/102643425-500b6a00-4170-11eb-9435-56f988e3db70.png)

flyingcircle:
[...]

PR #2717, opened 2020-12-18
fix image source for local images

flyingcircle:
Fixes:  #2714

@Aparus please check out this change and see if that fixes it.


Aparus:
> Fixes: #2714
> 
> @Aparus please check out this change and see if that fixes it.

Yes, it fixed it for <Image source={require} /> . I hope <Image source={{uri}} also is still working. 
Thank you!

PR #2723, opened 2020-12-22
Windows update

### LINKED ISSUE #2680: Add support for Windows
I want to use this library on Windows. Some components are fine, but components, that choose their style depending on platform do not work. This is because the library is calling react-native's _Platform.select_, which on Windows returns _undefined_.
So then `Platform.select(theme.colors.platform).grey` fails with an error _Cannot read property 'grey' of undefined_.

Add fallback colors when platform is undefined.

An obvious alternative would be React Native team making `Platform.select` work on Windows too.
My temporary fix is to define my own _Platform_ object with method _select_ that always returns '_android_'.

flyingcircle:
Fixes #2680

This actually wasn't too far reaching. I add a default option to colors. Most every other Platform.select already had a default option available.

@petr7555 can you check this out and see if it fixes the issue?


pranshuchittora:
@flyingcircle this will surely work. But for more native feel RNW has some other constants to make components look and feel like native apps. 
Have a look at this -> https://microsoft.github.io/react-native-windows/docs/windowsbrush-and-theme#using-theme-brushes-in-a-style and this -> https://microsoft.github.io/react-native-windows/docs/windowsbrush-and-theme#using-theme-brushes-in-a-style

pranshuchittora (approved):

pranshuchittora:
For more advanced usage we might need to modify the theme provider which internally handles those acrylic, ascent and other UI mods.

What's your thought on this?

flyingcircle:
I'm not sure how many RN Windows / MacOS developers there really are out there. It seems like a pretty niche area, and I don't want to bloat phone devs with another package. But I also don't know that having a "windows themed" app is a requirement for most Windows apps either. If they need to be that niche, they likely would have to make many updates in order to get that working anyways.

petr7555:
Thanks for these changes, I have been using it since and it works really nice.

PR #2737, opened 2020-12-30
Doc prop split

flyingcircle:
Fixes: #2719 

@pranshuchittora test this out and see if this lets you do what you want in the playground

PR #2738, opened 2020-12-30
separate out doc props sections for playground

flyingcircle:
Fixes: #2737 

@pranshuchittora alright. This should be clean. Try this out and see if it works the way you think it should


pranshuchittora (approved):
LGTM, thanks :+1:

Issue #2734, opened 2020-12-28
Tap outside to close bottom sheet


wangyi7099:
Is there a way to close buttom sheet by tapping the outside? Thanks!

flyingcircle:
Currently, no. I don't believe so.

freshomer:
Here is a pr to fix this:

https://github.com/react-native-elements/react-native-elements/pull/2740

migumelar:
Hi for temporary workaround I able to close the sheet by wrapping the children with some 

here is the snippet: 
```js

// to get screen height you can use https://github.com/facebook/react-native/issues/23693#issuecomment-523640102
const height = yourFunctionToGetScreenHeight()

// omitted code

<BottomSheet
                        isVisible={isVisible}
                        modalProps={{
                            onRequestClose: () => {
                                setIsVisible(false)
                            }
                        }
                        }
                    >
                            <TouchableWithoutFeedback onPress={() => setIsVisible(false)}>
                                <View style={{ flex: 1, height: height }}   > 
                                    <TouchableWithoutFeedback onPress={() => console.log('im also called')}>
                                          
                                          // Your bottom sheet content is here
                                          // example: 
                                        <View style={tailwind('absolute bottom-0 right-0 left-0 bg-white flex p-10 mx-4')}>
                                            <Text style={tailwind('text-black')}>Helloo theree</Text>
                                            <Button title="click me" onPress={() => { console.log('the child clicked and not triggering close') }}>
                                            </Button>
                                        </View>

                                    </TouchableWithoutFeedback>
                                </View>
                            </TouchableWithoutFeedback>
                    </BottomSheet>

```

PR #2740, opened 2020-12-31
fix(bottom_sheet): enable touch outside to close.

freshomer:
Fix issue #2734


src/bottomSheet/BottomSheet.js
```diff
@@ -40,6 +47,13 @@ class BottomSheet extends Component {
             containerStyle && containerStyle,
           ])}
         >
+          {touchOutsideToClose && (
+            <TouchableOpacity
```
flyingcircle, line 51: I think here `TouchableWithoutFeedback` would be a better choice.

website/docs/bottomsheet.md
```diff
@@ -42,9 +42,13 @@ const list = [
 
 ## Props
 
-- [`containerStyle`](#containerStyle)
-- [`isVisible`](#isvisible)
-- [`modalProps`](#modalprops)
+- [Usage](#usage)
```
flyingcircle, line 45: Not sure what happened here. These lines should not be added here.

website/docs/bottomsheet.md
```diff
@@ -42,9 +42,13 @@ const list = [
...
+- [Props](#props)
+- [Reference](#reference)
+  - [`containerStyle`](#containerstyle)
+  - [`isVisible`](#isvisible)
+  - [`touchOutsideToClose`](#touchoutsidetoclose)
```
flyingcircle, line 50: This shouldn't be so specific. This should instead match the `Overlay` component's prop `onBackdropPress` so that people can hook in their own function instead of being locked into only a single action.

flyingcircle (changes requested):
The typescript file also needs updated with the the new argument.

website/docs/bottomsheet.md
freshomer: Not sure what happened with my editor. It is automatically generated. And it has been reverted.

website/docs/bottomsheet.md
freshomer: Touching outside to close a bottom sheet is a very natural behavior. If someone has to hook this event, maybe another method could be added here just as you mentioned?

src/bottomSheet/BottomSheet.js
freshomer: Tried this choice. But `TouchableWithoutFeedback` doesn't allow empty child here.

website/docs/bottomsheet.md
flyingcircle: No, I do not agree with maintaining 2 props when 1 would be just fine. Writing a hook to close the sheet is a single line and not burdensome. And any developer should be very familiar with basic `state` variables in order to achieve it through a generic hook.

flyingcircle: Plus `isVisible` is a prop not a part of the state. So this method just won't make sense/won't work.

src/bottomSheet/BottomSheet.js
flyingcircle: Checkout here how the `Overlay` code is able to do it: https://github.com/react-native-elements/react-native-elements/blob/next/src/overlay/Overlay.js

website/docs/bottomsheet.md
freshomer: It was before I merged latest changes. BTW, I believe there would not be many develops who would like to hook this action in most cases and I saw several developers ask for this basic feature. It is different from `onBackdropPress`.

src/bottomSheet/BottomSheet.js
freshomer: I know how to do it. But I didn't see any benefits in that way.

jonaujo89:
did anybody fixed it

PR #2749, opened 2021-01-11
added onBackdropPress on BottomSheet

superdopey:
Added onBackdropPress like on the Overlay component. Users will be able to close the BottomSheet when clicking the backdrop just like how this is done on the Overlay component.

### DIFF
--- a/src/bottomSheet/BottomSheet.js
+++ b/src/bottomSheet/BottomSheet.js
@@ -20,18 +20,26 @@ function BottomSheet({
       {...modalProps}
     >
       <SafeAreaView
-        style={StyleSheet.flatten([
-          styles.safeAreaView,
-          containerStyle && containerStyle,
-        ])}
+        style={styles.safeAreaView}
         {...props}
       >
          <Pressable
-        onPress={onBackdropPress}
-        style={styles.pressable}>
-          <View>
+        onPress={()=> {
+          if(onBackdropPress)
+          {
+            onBackdropPress
+          }
+        }}
+        style={StyleSheet.flatten([
+          styles.pressable,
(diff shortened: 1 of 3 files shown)


flyingcircle:
Hey @superdopey thanks for the PR. The prop would also need to be added to the documentation in `website/docs` and also the type needs to be added in the `index.d.ts`. After you've added that, I can do a full code review.

MagicFame:
Looks good, I think the PR should be merged..

superdopey:
@flyingcircle I made the changes you suggested.

flyingcircle:
@superdopey alright I tried it out in the example app but I ran into this:
Current Version:


Branch Code:


There was also an issue that if the child view was displaying static (not-pressable) elements, the `onBackdropPress` was still being called when they were pressed.

flyingcircle (changes requested):
Changes requested as mentioned in the above comment.

superdopey:
> Changes requested as mentioned in the above comment.

@flyingcircle The changes are applied. Btw I am still new to doing pull-request like this, when I push changes do I need to comment here or do you see the changes regardless of the comment?

superdopey:
> > Changes requested as mentioned in the above comment.
> 
> @flyingcircle The changes are applied. Btw I am still new to doing pull-request like this, when I push changes do I need to comment here or do you see the changes regardless of the comment?

@flyingcircle Is there some else you need before we can continue this PR?

flyingcircle:
@superdopey I made some typescript changes. Hopefully it doesn't interfere with your changes here, but can you pull and merge the current `next` branch into your PR to resolve the "conflicting" files and then I'll test it out again.

Issue #15, opened 2016-09-13
Would love to see a Two-Marker Slider for Setting a Range


quadsurf:

dabit3:
@quadsurf This sounds like a good idea. Can you shoot over a link to an example of what you had in mind, maybe implemented somewhere else? Thanks

TheGermanGuy:
I think he is referring to a bidirectional related slider. 
Similar to the one on Tinder or something like the one shown in the link(http://www.tutorialspark.com/jqueryUI/jQuery_UI_Slider_Bidirectional_Related_Slider.php)
I would love to have this too.

Monte9:
@quadsurf Here are some sliders that I found.
1. https://github.com/jeanregisser/react-native-slider - I like that it's customizable & API looks pretty robust. 
1. https://github.com/JackDanielsAndCode/react-native-multi-slider - Has implementation of multi-selection slider
1. https://github.com/franfran/react-native-snap-slider - Love the functionality that this provides

I think adding a **custom slider API** (like 1) + **multi-selection** (like 2) + **snap functionality** (like 3) would be good addition to the elements. 

@dabit3 thoughts?

dabit3:
@Monte9 I like the idea of having the customizable api of 1 and the multi selection of 2, and the snap functionality looks cool as well. 

1. We can possibly add one of those existing projects as a dependency to get the component working out of the box.

2. or maybe you can even fork [#1](https://github.com/jeanregisser/react-native-slider) and implement the two slider feature into a pr or something and we can still add as a dependency

3. or maybe we can copy the existing component and add out tweaks as a new component to RN Elements?

I'm cool with any of these ideas, or anything along those lines really.

Monte9:
Update: Slider has been added to the latest release of RNE. Check out the documentation [here](https://github.com/react-native-community/react-native-elements#slider-component).

We went with Option 1. I'll leave this issue open as it would be cool to add on Option 2 & Option 3 to the existing Slider component.

If anyone has bandwidth and would like to submit a PR to enhance it, I am open to merging it.

Monte9:
In an effort to streamline all RNE issues such that it is easier to get to solutions faster/ submit a new issue we have decided to close any enhancement/roadmap features as "issues". 

Feel free to use this issue to talk about this feature and ping @binoy14 or me (@monte9) if you feel like you can submit a PR for this. Thanks in advance.

[...]

PR #2757, opened 2021-01-22
Added range selection mode to slider.

mt81:
Added the range selection to the slider #15. 
I have opened a draft pull request, to get initial comments and suggestions.


flyingcircle:
Looks promising. The only thing I'm thinking is that to reduce down the props, instead of adding `rangeSelection`, just determine if range should be enabled by whether a user only passed 1 value or 2 values.

PR #2760, opened 2021-01-25
Expo: remove global usage due to removal in SDK 41.

### LINKED ISSUE #2747: Button, Header and ListView are using depricated global.Expo object (SDK40)
After upgrading to expo SDK40, I started getting this warning:
```
Your project is accessing the following APIs from a deprecated global rather than a module import: LinearGradient (expo-linear-gradient).

The global "__expo" and "Expo" objects will be removed in SDK 41. Learn more about how to fix this warning: https://expo.fyi/deprecated-globals
```

After a bit of investigation I found that it is caused by `Button `property `linearGradientProps`, which uses `global.Expo` object.
`Header `and `ListView `is also affected by the same issue.
image

Could some one fix this and create PR?

sviande:
Remove usage of global.Expo due to deprecation in SDK 40 and removal in SDK 41.

Fix #2747

### DIFF
--- a/src/header/Header.js
+++ b/src/header/Header.js
@@ -50,7 +50,7 @@ Children.propTypes = {
 class Header extends Component {
   componentDidMount() {
     const { linearGradientProps, ViewComponent } = this.props;
-    if (linearGradientProps && !global.Expo && !ViewComponent) {
+    if (linearGradientProps && !ViewComponent) {
       console.error(
         "You need to pass a ViewComponent to use linearGradientProps !\nExample: ViewComponent={require('react-native-linear-gradient')}"
       );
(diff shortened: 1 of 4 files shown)


src/header/Header.js
```diff
@@ -74,8 +74,8 @@ class Header extends Component {
       barStyle,
       children,
       linearGradientProps,
-      ViewComponent = linearGradientProps && global.Expo
-        ? global.Expo.LinearGradient
+      ViewComponent = linearGradientProps
```
flyingcircle, line 77: I think this might be just slightly easier to read as:
```js
ViewComponent = (linearGradientProps || !backgroundImage) ? View : ImageBackground
```

flyingcircle (changes requested):

flyingcircle:
@sviande thank you very much. This looks good, just 1 small change, and then I can test it out this weekend.

flyingcircle (approved):

Issue #2761, opened 2021-01-26
App can't compile after adding a <Header> tag (crosspost from Slack)


lorecast162:
Hello. I am just learning how to do stuff with React Native and I'm developing my little app on Linux and testing on Android.
I wanted to have a little titlebar on the top and stumbled upon the ReactNative Elements `Header` component, so I installed it and the vector icons and there was no problem there.
But, after I put in a little Header tag with just a center component (`<Header centerComponent={<Text>5C</Text>}></Header>`) my app crashed saying `Invariant Violation: requireNativeComponent: "RNCSafeAreaView" was not found in the UIManager. `
So I looked that up on google and I only found issues regarding `react-native-navigation` using XCode and one SO answer that seemed like it would do the trick (https://stackoverflow.com/a/60319825). It says to  install `react-native-safe-area-context` and then do a pod install.
The issue now is that 1) I can't do pod install as I'm on Linux and that is a macOS-only thing and 2) Even that didn't fix anything.
So I commented out that Header tag and everything went back to normal.
I even tried the `Button` component to see if that was problematic too but that one worked fine.
Here is a screenshot taken from my phone (I took two and assembled them as one)
![img](https://i.imgur.com/AQZzEQi.png)

lorecast162:
I didn't flag this as a bug as I don't _know_ it is.

PR #2762, opened 2021-01-26
add note for rn-safe-area-context setup

flyingcircle:
Fixes: #2761


phillipb:
@flyingcircle heads up, this doesn't fix non-expo apps. There appears to be an issue with that library auto-linking. I had to manually like that dependency. https://github.com/th3rdwave/react-native-safe-area-context/issues/165

PR #2764, opened 2021-01-28
Themmyloluwaa/pull request

themmyloluwaa:
Hi, 

I noticed the checkbox component supports being disabled but the docs don't have this listed.

The rating component also supports setting the background colour but isn't available in the docs so I added them both. 

Thanks.


flyingcircle:
`Checkbox` doesn't have an explicit `disabled` prop. The default component of `Checkbox` is `TouchableOpacity` which has a `disabled` prop. So this doesn't need to be added to the docs.

PR #2769, opened 2021-02-05
change react-native-safe-area-context to a peer

walfly:
Having multiple versions of the react-native-safe-area-context causes an invariant violation when starting your app.  This fixes these two issues https://github.com/react-native-elements/react-native-elements/issues/2721 and https://github.com/react-native-elements/react-native-elements/issues/2768

I removed react-native-safe-area-context from dependencies and added it to peer and dev depencies.

### DIFF
--- a/website/docs/getting_started.md
+++ b/website/docs/getting_started.md
@@ -99,3 +99,14 @@ npm i --save react-native-safe-area-context
 # link
 react-native link react-native-safe-area-context
 ```
+
+It is required to add the `SafeAreaProvider` to the outside of the app. The suggested way to do this is
+the following:
+
+```js
+import { SafeAreaProvider } from 'react-native-safe-area-context';
+
+function App() {
+  return <SafeAreaProvider>...</SafeAreaProvider>;
+}
+```
--- a/website/docs/getting_started.md
+++ b/website/docs/getting_started.md
@@ -85,14 +85,17 @@ debug it using
 
 ### Step 3: Setup react-native-safe-area-context
 
-Some components such as `Header` or `BottomSheet` rely on using the **react-native-safe-area-context** library. React Native Elements already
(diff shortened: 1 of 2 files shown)


flyingcircle:
I agree this change needs to be made. Can you also change the docs here: https://reactnativeelements.com/docs/#step-3-setup-react-native-safe-area-context so that it reflects this change?

walfly:
Updated

website/docs/getting_started.md
```diff
@@ -85,14 +85,17 @@ debug it using
...
 
-Some components such as `Header` or `BottomSheet` rely on using the **react-native-safe-area-context** library. React Native Elements already
-lists the library as dependency, but it is required to add the `SafeAreaProvider` to the outside of the app. The suggested way to do this is
-the following:
+If you have already installed **react-native-safe-area-context** as a dependency for
+your project you can skip this step. Otherwise run the following command:
+
+> _Manual linking of react-native-safe-area-context is not necessary if you're using react-native@0.60.0 or above since it is done automatically. This will throw an error though it won't prevent the application from running. To fix this you'll simply have to run `react-native unlink react-native-safe-area-context` and the process will run as expected._
 
-```js
```
flyingcircle, line 92: Don't people still need to add this section to their app?

flyingcircle (changes requested):

website/docs/getting_started.md
walfly: Does that belong here, and not just in the safe-area-context readme? I'll add it back.

flyingcircle: I can agree with that, but a link to the setup guide would be good to let people know that there's more setup than just installing it

flyingcircle:
@walfly thanks! I'll test it out this weekend.

PR #2772, opened 2021-02-09
fix typescript Colors.platform type

SyMind:


flyingcircle:
This was fixed in #2781

PR #2791, opened 2021-03-02
Source prop missing in Image from Accessory.jsx

Brechard:
When Accessory was moved out of Avatar having a custom image as Accessory stopped working. 

See the issue https://github.com/react-native-elements/react-native-elements/issues/2787

### DIFF
--- a/src/avatar/Accessory.js
+++ b/src/avatar/Accessory.js
@@ -31,6 +31,7 @@ function Accessory({
       <View>
         {source ? (
           <Image
+            source={source}
             style={{
               width: size,
               height: size,


flyingcircle (changes requested):
I just pushed a big code change to convert the codebase to TS. If you could pull and fix your branch, I would appreciate it. Sorry for the inconvenience.

Thanks

Brechard:
I saw you did the changes in the branch 'next' instead of 'patch' so I created another pull request. I am sorry if this is not the way it should be done, this is my first time contributing to a public repo.

flyingcircle:
No worries. I think `patch` has become more of an artifact of the early dev work. I don't really keep that uptodate like I probably should

Issue #2802, opened 2021-03-09
export components props types


MohammedFaragallah:
current behavior
```jsx
import { AvatarProps } from 'react-native-elements/dist/avatar/Avatar';
```

expected behavior
```jsx
import { AvatarProps } from 'react-native-elements';
```

I can work on a PR if you agree

PR #2803, opened 2021-03-09
Exported AvatarProps for the issue #2802

hetanthakkar1:
Exported AvatarProps from index.ts to allow importing AvatarProps from `react-native-elements` instead of `react-native-elements/dist/avatar/Avatar`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -21,8 +21,7 @@ import Image from './image/Image';
 // Productivity
 import Card from './card/Card';
 import Tile from './tile/Tile';
-import Avatar from './avatar/Avatar';
-import { AvatarProps } from './avatar/Avatar';
+import Avatar, { AvatarProps } from './avatar/Avatar';
 import Accessory from './avatar/Accessory';
 import Header from './header/Header';
 import PricingCard from './pricing/PricingCard';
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -22,6 +22,7 @@ import Image from './image/Image';
 import Card from './card/Card';
 import Tile from './tile/Tile';
 import Avatar from './avatar/Avatar';
+import { AvatarProps } from './avatar/Avatar';
```
flyingcircle, line 25: This can be merged with the line above:
```js
import Avatar, {AvatarProps} from './avatar/Avatar';
```

flyingcircle (changes requested):
If we're going to export the Prop types, then let's add exports for all of the components. Add exports for all of the components.

jeeva2812 (commented):
LGTM. Works for me.

hetanthakkar1:
@flyingcircle  I have made the changes. Could you please review it

jeeva2812 (approved):

hetanthakkar1:
Okay I am on it. @flyingcircle I have a doubt. I can see d.ts files when I install react -native-elements in a project in node modules.
But I can't see those files in the repo. I want to solve an issue that requires me to change that d.ts files. How can I get that file?

Issue #2794, opened 2021-03-05
prop type validation in icon color property will be failed if use PlatformColor


mr-africa:
**Explain what you did (Required)**

I try to use [platform color](https://reactnative.dev/docs/platformcolor) as a color for icon

**Expected behavior (Required)**

no warning messages

**Describe the bug (Required)**
Everything is working, but I get warning message:

```
Warning: Failed prop type: Invalid prop `color` of type `object` supplied to `Icon`, expected `string`.
```

**To Reproduce (Required)**

for ios:
```
import { PlatformColor } from 'react-native'

<Icon name="home" color={PlatformColor('secondaryLabel')} />
```

for android:
```
import { PlatformColor } from 'react-native'

<Icon name="home" color={PlatformColor('@android:color/background_light')} />
```

**Your Environment (Required):**

| software              | version |
| --------------------- | ------- |
| react-native-elements |     3.2.0    |
| react-native          |    0.63.4     |

jeeva2812:
PR #2812 might fix this.

PR #2812, opened 2021-03-10
Modified the props of Icon color to incorporate PlatFormColor for issue  #2794

hetanthakkar1:
I am a bit skeptical about my approach as I have changed the type of color prop to any.  I tried using the type `color: string | object` but that resulted in the ts error saying that **object is not assignable to string**. 

One thing I am sure of is that we need to incorporate the object as one of the types for color prop as the PlatformColor returns an object. So I thought that otherwise also we are passing object and string so assigning type any to color prop might make sense

Secondly, the TouchableNativeFeedback ripple effect takes a color as an argument whose type must be a string. Changeing that to include the object  type will require a modification in the core package so when the color is passed as PlatformColor( which is an object) we won't be able to pass that into TouchableNativeFeedback ripple effect so as a default I passed  `theme.color.black `(we can change this to any color)

### DIFF
--- a/src/icons/Icon.tsx
+++ b/src/icons/Icon.tsx
@@ -37,7 +37,7 @@ export type IconType =
 
 export interface IconObject extends TouchableHighlightProps {
   name?: string;
-  color?: string;
+  color?: any;
   size?: number;
   type?: IconType;
   style?: StyleProp<TextStyle>;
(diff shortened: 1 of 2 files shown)


hetanthakkar1:
@flyingcircle  what are your thoughts about my approach

flyingcircle:
I can appreciate the attempt and I thought of `any` myself when trying to come up with a solution to it. There's a deeper problem here though that I think will require a better solution. Perhaps it might be worth it to do some investigation into the types from `vector-icons` and `colors` as it seems like there's something between those 2 that isn't working correctly. My hunch here is actually that the `vector-icons` types library might have a bug in it?

hetanthakkar1:
I'm investigating the issue. If we couldn't find the issue causing that type problem,can we then implement my solution as the last resort!?

flyingcircle:
Solved by https://github.com/react-native-elements/react-native-elements/pull/2847

Issue #2751, opened 2021-01-14
Make sure component is mounted before calling setState.


mvehar:
https://github.com/react-native-elements/react-native-elements/blob/0ae0852b48c62af30f6c53af48a4ed3fdb5b5751/src/tooltip/Tooltip.js#L202

Here the setState is called in callback (measure, requestAnimationFrame). In react code can happen that tooltip is removed before this code is executed (virtual lists) and throws an error.

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in Tooltip (at withTheme.js:42)
    in ThemedComponent (at withTheme.js:59)
    in ForwardRef(Themed.Tooltip) (at base.item.component.tsx:249)
```

flyingcircle:
You said it though. Tooltip has to do some measuring after layout in order to function properly. What is your suggestion here?

flyingcircle:
This also isn't an error. Just a warning.

mvehar:
True, its a warning, but with new LogBox in react native I gets annoying to see this warning box all the time. 

Solution could be using class property "isMounted" and setting the boolean value in componentDidMount, componentWillUnMOunt (or useEffect if using hooks).

Then any callback/promise resolve could check its value before applying changes/state.

The-Keshav-Agarwal:
@flyingcircle I would like to work on this issue but can you please tell me isMounted is already defined or I have to make a local variable?

flyingcircle:
@The-Keshav-Agarwal if you are new to React Native, I would suggest taking some time to learn how to run the app and work with the library first. `isMounted` is a React concept, so you'd have to look at the documentation there.

The-Keshav-Agarwal:
@flyingcircle can you assign this to me so that I can give it a try.

The-Keshav-Agarwal:
@flyingcircle i am getting this error while committing the changes and because of this I am not able to push my code.
![image](https://user-images.githubusercontent.com/56120769/110580456-c081c000-818e-11eb-9584-d3952f13ab88.png)
It is showing there is some error in  `withTheme.tsx`  but I didn't make any changes in that file. Can you please tell me what is going wrong so that I can push my changes.

PR #2821, opened 2021-03-10
fix: Make sure component is mounted before calling setState. #2751

namantayal:
Fixes  issue #2751 

No
No
**Summary**
Added _isMounted in tooltip.tsx, set it true on componentDidMount and false on componentWillUnmount. Then added condition on setState so that it works only when the component is mounted

No

**Other information**


Naman-1234 (changes requested):
@namantayal, Nice Work but I think "&&" is a better way to perform something rather than Checking through If condition.
Example: boolean_variable==true&&(SOMETHING_U_WANNA_Do).

namantayal:
@Naman-1234 Changed the if conditions, is it fine now?

PR #2830, opened 2021-03-11
Migrated button component to RN Functional Component

blackwolf08:
Migrated Button to a functional component.

Yes, but with a catch! useEffect hook cant be tested with shallow render of enzyme GitHub issue here https://github.com/enzymejs/enzyme/issues/2086, therefore useEffect hook test fails, but thoroughly tested on demo app.

**Summary**

Aiming to migrate all class-based components to functional components starting now! GSoC 2021

No

**Other information**


src/buttons/Button.tsx
```diff
@@ -53,173 +53,166 @@ export type ButtonProps = TouchableOpacityProps &
...
-      ...attributes
-    } = this.props;
+  const handleOnPress = useCallback(
+    (evt) => {
+      if (!loading) {
+        onPress(evt);
+      }
+    },
+    [loading, onPress]
+  );
```
blackwolf08, line 98: useCallback used instead of normal fat arrow function to optimize performance.

blackwolf08:
@flyingcircle review requested

flyingcircle:
Fixed in PR #2834 , closing this PR. Thank you for your contribution.

flyingcircle:
Actually rereviewing your code, I think changing the current code to use the `useEffect` statement that you have would be a welcome change. I'm not sure that the `useCallback` provides an actual improvement though. I'll reopen this if you want to pull in the latest from `next`.

blackwolf08:
@flyingcircle I have pulled from the next and implemented changes

arpitBhalla:
@blackwolf08 Without useEffect Dependency array, it will show error on each render,

blackwolf08:
@arpitBhalla please refer useEffect docs

arpitBhalla:
@arpitBhalla  I am saying about this line https://github.com/blackwolf08/react-native-elements/blob/3dfd4f8f4c653465176e5ff3d8ec355438db0ccb/src/buttons/Button.tsx#L59
It will show error on each render...?

sunnyscaler:
@arpitBhalla yup the point is, you can’t render a button component with linear gradient if you haven’t installed the appropriate package.

arpitBhalla:
@sunnyscaler  I got this, but what I want to say is on each re-render console.error will be called,

 ```javascript
useEffect(() => {
    if (props.linearGradientProps && !props.ViewComponent) {
      console.error(
        "You need to pass a ViewComponent to use linearGradientProps !\nExample: ViewComponent={require('react-native-linear-gradient')}"
      );
    }
  },[props.linearGradientProps,props.ViewComponent]);
```

Using this we can reduce console.error calls....?

blackwolf08:
@arpitBhalla using 'useMemo' would not respect new props if the component is controlled from the outside

blackwolf08:
I'm not sure what are you trying to say, using this code instead of without dependancy array would result in exactly same number of console errors if the 'if' condition  within the console log doesn't change

arpitBhalla:
Ohhk Sir, I got it Thanks

arpitBhalla:
@blackwolf08 Sir 

> I'm not sure what are you trying to say, using this code instead of without dependancy array would result in exactly same number of console errors if the 'if' condition within the useEffect returns true

@blackwolf08 Sir then why we use useEffect here 😅

arpitBhalla:
@blackwolf08, Sir Tests are failing(using useEffect).... as you said too , how to fix that???

blackwolf08:
@arpitBhalla I'm not quite sure, as per my research we could use mount instead of shallow in test, but that would be heavy, i tried to mimic useEffect through jest.spyOn but that didn't quite work, I also tried before and after methods of jest, but didn't work, I tried to update tue component with .update() method but it yielded same outputs. I'll try few more tomorrow...let me know if you find one! Thanks

arpitBhalla:
@blackwolf08 Sure SIr

Issue #2754, opened 2021-01-18
Ability to change theme for a component tree via state change


kdo1234:
**Is your feature request related to a problem? Please Describe.**
It seems that there is no way to dynamically change the theme for a ThemeProvider's tree directly via a state change. The theme for the ThemeProvider tree can be dynamically set before the first render, but can't be changed later because the tree does not start using the new theme. By "change the theme" I don't mean changing the properties of the theme (which would be expensive to detect) but changing the theme prop to refer to a different object.

**Describe the solution you'd like**
Changing the ThemeProvider's theme should update components within the ThemeProvider's tree.
```
export default class App extends React.Component {
  state = { theme: 0}

  theme() {
    return this.state.theme === 1 ? theme1 : theme0;
  }

  render() {
    return (
        <ThemeProvider
          theme={this.theme()}
          >
          <Button
            onPress={() => {this.setState((state) => {return {theme: (state.theme + 1) % 2 }})}}
            title={"Change Theme with state (does not work)"} 
          />
        </ThemeProvider>
    );
  }
}
```

**Describe alternatives you've considered**
I've implemented an alternative with a hidden component that uses getDerivedStateFromProps to trigger a call to `withTheme`, but it doesn't seem DRY to duplicate specifying the theme.

```
class ThemeChanger extends React.Component {
  state = {}

  static getDerivedStateFromProps(props, state) {
    const { realTheme } = props;
    if (realTheme !== state.realTheme) {
      if (state.realTheme !== undefined) {
        props.replaceTheme(realTheme);
      }
      return {
        realTheme,
      };
    }
    return null;
  }

  render() {
    return null;
  }
}

const ThemedThemeChanger = withTheme(ThemeChanger);

export default class App extends React.Component {
  state = { theme: 0}

  theme() {
    return this.state.theme === 1 ? theme1 : theme0;
  }

  render() {
    return (
        <ThemeProvider
          theme={this.theme()}
          >
          <ThemedThemeChanger realTheme={this.theme()}/>
          <Button
            onPress={() => {this.setState((state) => {return {theme: (state.theme + 1) % 2 }})}}
            title={"Change Theme with side effect (works)"} 
          />
        </ThemeProvider>
    );
  }
}
```

const ThemedThemeChanger = withTheme(ThemeChanger);

**Additional context**
Here is a snack that reproduces the current behavior and the alternative: https://snack.expo.io/DhVc8tX4z

flyingcircle:
[...]

PR #2833, opened 2021-03-12
Theme can now be changed on changing the state of component 

hetanthakkar1:

Added ability to change the theme for a component tree via state change
Fixed issue #2754

### DIFF
--- a/src/config/ThemeProvider.tsx
+++ b/src/config/ThemeProvider.tsx
@@ -74,8 +74,13 @@ export default class ThemeProvider extends React.Component<
         ),
         useDark,
       };
+    } else {
+      return {
+        theme: state.theme,
+        useDark,
+        update: false,
+      };
     }
-    return null;
   }
 
   updateTheme = (updates: RecursivePartial<FullTheme>) => {
--- a/src/config/ThemeProvider.tsx
+++ b/src/config/ThemeProvider.tsx
@@ -57,7 +57,10 @@ export default class ThemeProvider extends React.Component<
 
   static getDerivedStateFromProps(props, state) {
     const { useDark } = props;
-    if (useDark !== state.useDark && !state.update) {
+    if (
+      useDark !== state.useDark ||
+      (props.theme !== state.theme && !state.update)
+    ) {
       const defaultColors = useDark ? darkColors : colors;
       return {
         theme: deepmerge(
--- a/src/config/ThemeProvider.tsx
+++ b/src/config/ThemeProvider.tsx
@@ -56,9 +56,8 @@ export default class ThemeProvider extends React.Component<
   }
 
   static getDerivedStateFromProps(props, state) {
-    const { useDark, theme } = props;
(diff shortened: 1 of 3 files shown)


hetanthakkar1:
Sorry for the multiple commits that resulted because of multiple commits I did for testing my changes

hetanthakkar1:
Demo Code which can is now supported:

image



import { ThemeProvider, Button } from "react-native-elements";
import React, { useState } from "react";
const theme1 = {
  Button: {
    titleStyle: {
      color: "white",
    },
  },
};
const theme2 = {
  Button: {
    titleStyle: {
      color: "yellow",
    },
  },
};
export default function App() {
  const [theme, setTheme] = useState(theme1);
  return (
    <ThemeProvider theme={theme}>
      <Button
        title="My Button"
        onPress={() => {
          if (theme == theme1) {
            setTheme(theme2);
          } else {
            setTheme(theme1);
          }
        }}
      />
      <Button
        title="My 2nd Button"
        onPress={() => {
          if (theme == theme1) {
            setTheme(theme2);
          } else {
            setTheme(theme1);
          }
        }}
      />
    </ThemeProvider>
  );
}

hetanthakkar1:
@flyingcircle  You can use my above code to test my changes

hetanthakkar1:
@flyingcircle  Could you please review!?

src/config/ThemeProvider.tsx
```diff
@@ -76,6 +81,7 @@ export default class ThemeProvider extends React.Component<
   updateTheme = (updates: RecursivePartial<FullTheme>) => {
     this.setState(({ theme }) => ({
       theme: deepmerge(theme, updates),
+      update: true,
```
flyingcircle, line 84: I don't understand the use of update and setting it true here. Are we saying that if you run `updateTheme` then changing the `theme` prop will stop working? How about we just go without the update variable and just update the `theme` as it is passed in.

flyingcircle (changes requested):

hetanthakkar1:
@flyingcircle  So we are updating theme at 2 places in my code .One at the updateTheme method and other at getDerivedStateFromProps method so we dont want to execute both of them to do the same.So if updateTheme function will update theme it will set the update to true and so the updation in getDerivedStateFromProps wont take place.Removing that update state will also fail 1 test case

flyingcircle:
@hetanthakkar1 What I'm asking is if a user updated the theme using `updateTheme`and then updates it later by changing the `theme` prop, the theme wouldn't update the second time right? I don't see where in this code that `update` gets switched back to `false`.

hetanthakkar1:
@flyingcircle  I got your point and I have made the necessary changes. Can you please review

hetanthakkar1:
@flyingcircle Okay i got your point. I am thinking on it. Please ignore the last commit . I pushed it by mistake that it will fix but realized later that it isn't correct and didnt wat to delete commit by force pushing

PR #2839, opened 2021-03-12
migrated input to functional component

blackwolf08:

Input component to Functional Component

**Summary**

GSOC 2021

**Other information**
We have to rewrite our Input component tests as Functional components do not use `.instance()` after shallow mounting, therefore tests failing @flyingcircle can you help with this?

### DIFF
--- a/src/input/Input.tsx
+++ b/src/input/Input.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { forwardRef, useImperativeHandle, useRef } from 'react';
 import {
   Text,
   View,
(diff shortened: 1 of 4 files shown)


blackwolf08:
@flyingcircle use `useImperativeHandle` with forwarded ref for instance methods, tests failing as `.instance()` doesn't work with Functional Components, we have to rewrite our test, its functional, tested it on demo app

flyingcircle (changes requested):
Well I'd definitely want the unit tests fixed before accepting this PR. But when I was testing this out in the sample app I got this error when I clicked inside any of the inputs.

Simulator Screen Shot - iPhone 8 - 2021-03-13 at 20 39 04

Issue #2797, opened 2021-03-07
Rebuild Tooltip into a Tooltip2


flyingcircle:
`Tooltip` is the component that seems to attract the highest number of bug tickets. Part of this is that it is trying to tackle a rather difficult problem of using absolute positioning and using Layout positioning which seems to have its own problems with implementing a consistent solution.

What is needed is a replacement component that for now we will call `Tooltip2` so that we can eventually deprecate `Tooltip`/ replace it with this new `Tooltip`.

Requirements list:
Functional component
All state vars use hooks
Verify that all currently open `Tooltip` issues are resolved:  #2782, #2751, #2462, #2414, #2392, #2210
Uses latest animation API
Looks sleeker: Follow Material UI styles and pattern

jeeva2812:
Hello, @flyingcircle I am willing to work on this issue. Can you please assign me for this?

jeeva2812:
I am extremely sorry @flyingcircle looks like it's much more extensive work than I thought it would be. I am unassigning myself.

abhishekkumar08:
@flyingcircle can you assign me this? And can you guide me how should I start on this, the changes would be the same like the given PR's?

deevo-sage:
@flyingcircle can I be assigned the task. I converted the Button component to a Functional component in a recent PR and would love to do the same with The Tooltip if possible.

flyingcircle:
This is quite a difficult task, I would say try and see how far you can get with it. It may take multiple people working on it so feel free to collaborate.

deevo-sage:
I fail to understand why the test case is failing otherwise, both the components (Tooltip, Tooltip2) are rendering and working the same when run on an android emulator in a practice project.

jalajcodes:
> I fail to understand why the test case is failing otherwise, both the components (Tooltip, Tooltip2) are rendering and working the same when run on an android emulator in a demo project.

Can you provide some more information about the error?

deevo-sage:
@jalajcodes 
**  Tooltip component › does not render pointer

    TypeError: Cannot read property 'state' of null

      69 |     ).root;
      70 |     component.findAllByType(TouchableOpacity)[0].props.onPress();
    > 71 |     expect(component.instance.state.isVisible).toBe(true);
         |                               ^
      72 |     try {
      73 |       component.findByType(Triangle);
      74 |     } catch (e) {

      at Object.state (src/tooltip/__tests__/Tooltip2.js:71:31)


[...]

PR #2842, opened 2021-03-12
added Tooltip2 Functional component

deevo-sage:
This PR addresses #2797 

This PR adds Tooltip2 which is Migration of Tooltip to a functional component

yes, I added tests similar to the ones for Tooltip. Except 1 which accessed ```component.instance``` which does not exist in  Functional components.

no, it doesn't.

### DIFF
--- a/src/tooltip/Tooltip2.tsx
+++ b/src/tooltip/Tooltip2.tsx
@@ -1,4 +1,10 @@
-import React, { Fragment, useState, useEffect, useRef } from 'react';
+import React, {
+  Fragment,
+  useState,
+  useEffect,
+  useRef,
+  useCallback,
+} from 'react';
 import {
   TouchableOpacity,
   Modal,
@@ -65,9 +71,6 @@ const Tooltip: React.FunctionComponent<TooltipProps> = (props) => {
   const [elementWidth, setElementWidth] = useState(0);
   const [elementHeight, setElementHeight] = useState(0);
 
-  useEffect(() => {
-    requestAnimationFrame(getElementPosition);
-  });
   const {
     withOverlay = true,
     overlayColor = 'rgba(250, 250, 250, 0.70)',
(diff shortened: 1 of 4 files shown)


renjfk:
It's missing external interaction which class component had so it should have a prop to control internal state externally. You can either consider props presence and use it instead of internal state or just watch (use-effect) prop and delegate to internal state.

src/tooltip/Tooltip2.tsx
```diff
@@ -0,0 +1,322 @@
...
+    onClose = () => {},
+    onOpen = () => {},
+    skipAndroidStatusBar = false,
+    ModalComponent = Modal,
+    closeOnlyOnBackdropPress = false,
+    pointerColor,
+  } = props;
+
+  const toggleTooltip = () => {
+    getElementPosition();
```
flyingcircle, line 93: Why is this called here?

flyingcircle (commented):
I'm also not a fan of how `Tooltip` is currently written. It builds up components using function calls... It just doesn't look clean to me or easy to understand. This is a good first step, but I want `Tooltip2` to be more than just a `FunctionComponent` of the first one. If we were just converting the code from class to function then `Tooltip2` wouldn't be necessary. The functions need additional refactoring and rethinking. There are simply too many functions jumping all over the place, the styling and components are spread out throughout instead of being organized in a single place, the order of execution is difficult to follow.

I hope this is enough to go on. You've done a good job so far, I'd like to see `Tooltip2` be easier to read than the original `Tooltip`.

src/tooltip/Tooltip2.tsx
renjfk: That part is actually useful for ensuring absolute position one last time before showing e.g. in a changing layout such as flat list however it's a bit unnecessary before hiding.

renjfk: Actually now that I think about it if we consider this usage as safe and not overkill then the one called at mount becomes redundant so basically tooltip will hold render until next coordinates are available before each show. This can be easily achieved by clearing coordinates before hide and set before show.

Or alternatively this kind of behavior can be actually configurable in case of static layouts, either dynamic positioning which calculates before each trigger or static; use always same the one calculated during mount.

arpitBhalla:
@deevo-sage May I continue on this?, Please

Issue #2798, opened 2021-03-07
Fix all @ts-ignore tags


flyingcircle:
In order to convert over to typescript there were some particularly difficult issues that weren't able to be resolved without using @ts-ignore. Resolve all remaining @ts-ignore tags.

- [x] index.ts
- [x] Accessory
- [x] Avatar
- [x] Button
- [x] ButtonGroup
- [x] Card - subcomponents
- [x] Checkbox
- [x] ThemeProvider
- [x] withTheme
- [x] Header
- [x] Icon
- [x] Image
- [x] Input
- [x] Pricing
- [ ] Searchbars
- [x] SocialIcon
- [x] FeaturedTile
- [x] Tile
- [x] Tooltip

nlok5923:
@flyingcircle i want to work on this issue but please can you explain me a bit more about this issue as i am new to this project.

flyingcircle:
Yeah, so I recently converted the library to using TypeScript. However, in order to have everything compile, I had to add `// @ts-ignore` to different lines. This tag means that the typescript compiler ignores the line directly below it. 

So the goal here is to delete every instance of `@ts-ignore` and fix the typescript error that it is hiding.

nlok5923:
@flyingcircle sir can you please assign this issue to me so that i can create a pr for resolving this issue,

nlok5923:
@flyingcircle sir please can you let me know how can i compile package .

arpitBhalla:
@pranshuchittora May I work on this?

nlok5923:
@arpitBhalla thanks for asking but currently I am working on it so please you may find another one.

flyingcircle:
@arpitBhalla  yes, just say which component you're working on. There's lots of work here, so plenty of room for multiple people to work on it.

flyingcircle:
I cleared the assignees from this ticket because it is quite large and easily divided into smaller tasks. Please state which component's ignore tags you're going to work on as I have them listed above so that no one works on top of each other.

nlok5923:
@flyingcircle i had removed //@ts-ignore from accessory component as currently removing it returns not error.

flyingcircle:
@nlok5923 well, actually the error was being hidden because of an issue in `withTheme`. I pushed the fix so hopefully nothing is hidden anymore. But the fix in `withTheme` also ended up fixing `Accessory` as well.

sarthakm21:
@flyingcircle I would like to work on the button component

[...]

PR #2847, opened 2021-03-13
fix vector icons types package version difference

flyingcircle:
Relates to: #2798


src/social/SocialIcon.tsx
```diff
@@ -160,7 +160,6 @@ const SocialIcon: React.FunctionComponent<SocialIconProps> = (props) => {
     >
       <View style={styles.wrapper}>
         <Icon
-          //@ts-ignore
```
arpitBhalla, line 163: @flyingcircle It showing error to me

snitin315: Try installing deps again. -`yarn install`

PR #2876, opened 2021-03-18
Header Elevation

arpitBhalla:
new prop to elevate header component 

Yes
Yes
**Summary**
It looks nice😅

No

**Other information**


flyingcircle:
I'll be honest. I'm not too much of a fan of this change. There are a few places where having props for different styles might be useful, but usually an app only has a single header. So I just don't see a need to shorthand some styles in a prop that people might just end up wanting to tweak anyways. What are your thoughts?

arpitBhalla:
@flyingcircle it is in Material UI so I thought, it may help user

src/header/Header.tsx
```diff
@@ -144,6 +146,16 @@ const Header: React.FunctionComponent<HeaderProps> = (props: HeaderProps) => {
             alignItems: 'center',
           },
           backgroundColor && { backgroundColor },
+          elevated && {
```
flyingcircle, line 149: The style needs to be moved to the bottom in the `StyleSheet`

flyingcircle (changes requested):

src/header/Header.tsx
arpitBhalla: Sure Sir

flyingcircle:
I brought this up in the demo app and I thought it just looked a bit strange. What do you think about this instead:
```js
elevatedHeader: {
        shadowColor: '#000',
        shadowOffset: {
            width: 0,
            height: 6,
        },
        shadowOpacity: 0.6,
        shadowRadius: 8.0,
        elevation: 24,
    },
```

PR #2877, opened 2021-03-18
Add a Switch Component and Documentation

khushal87:
This PR adds a new Switch Component which was missing in the package

Yes, I have used appropriate testing tools to test.
Not yet, looking forward for PR verification and then I will do that for sure.
**Summary**
This add a new Switch component Capture23 Capture24 in the package which will  be useful. It inherits all the props of the native switch and provide theme support with better view.

Yes, this can be a breaking change in terms of feature

**Other information**
NIL

Waiting for your review and feedback of mentors. @flyingcircle Can you review it. Thanks a lot. Reference issue - #2877


src/switch/switch.tsx
```diff
@@ -0,0 +1,81 @@
...
+  Switch as NativeSwitch,
+  StyleProp,
+  ViewStyle,
+  ViewProps,
+  Platform,
+} from 'react-native';
+import { withTheme } from '../config';
+import { Theme } from '../config/theme';
+
+export type SwitchProps = {
```
flyingcircle, line 12: I don't understand why you rewrote the props here. Shouldn't this just be `SwitchProps` from React-Native?

src/switch/switch.tsx
```diff
@@ -0,0 +1,81 @@
...
+  onValueChange,
+  color,
+  style,
+  theme,
+  ...rest
+}) => {
+  // switchedOnColor deals with picking up a color provided as props by user or picks up default theme
+  const switchedOnColor = color || theme.colors.primary;
+
+  const onTintColor =
```
flyingcircle, line 33: So 
```js
Platform.OS === 'ios' || !disabled ? switchedOnColor : theme.colors.disabled
```

src/switch/switch.tsx
```diff
@@ -0,0 +1,81 @@
...
+      ? switchedOnColor
+      : disabled
+      ? theme.colors.disabled
+      : switchedOnColor;
+
+  const thumbTintColor =
+    Platform.OS === 'ios'
+      ? undefined
+      : disabled
+      ? theme.colors.disabled
```
flyingcircle, line 44: I don't understand this. If there's a `theme.colors.disabled` color, then choose between `grey1` or `grey4`?

flyingcircle (changes requested):

src/switch/switch.tsx
khushal87: Yes it can be, but I thought of using the types here, I am working on your review and will change it.

src/switch/switch.tsx
khushal87: Shall I change to this?

src/switch/switch.tsx
khushal87: That's where my confusion of color lied for the users to use it, if you find irrelevant its okay to remove it. Thanks.

flyingcircle (changes requested):

website/docs/props/switch.md
```diff
@@ -0,0 +1,62 @@
+## Props
```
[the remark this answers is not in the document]
khushal87: Done @flyingcircle

PR #2881, opened 2021-03-19
Removed ts-ignore

arpitBhalla:
Removed ts-ignore from:


src/input/Input.tsx
```diff
@@ -42,7 +43,7 @@ export type InputProps = TextInputProps & {
   labelProps?: object;
   theme?: Theme;
   renderErrorMessage?: boolean;
-};
+} & ThemeProps<InputProps>;
```
flyingcircle, line 46: Theme Props do not get added here. They get added at `withTheme`

flyingcircle (changes requested):

Issue #253, opened 2017-03-30
Dropdown or Picker component


kayazinc:
Is there any combobox (aka dropdown, picker) component?
eg. http://shoutem.github.io/docs/ui-toolkit/components/dropdown-menu

binoy14:
Not currently. Will add it to the roadmap. Thanks for reporting

dyaacov:
Any estimation? the lack of this components blocks me from using your great library :(

binoy14:
@dyaacov any suggestions on how it should look? Having some screenshots would be helpful for anyone building this feature

dyaacov:
https://storage.googleapis.com/material-design/publish/material_v_11/assets/0Bzhp5Z4wHba3UE5aelZUMnZUZzA/components_menus_behavior4.png

martinezguillaume:
Good idea ! Need this component ! 👏👏

felemedeiros:
I would add [https://material.io/guidelines/components/menus.html#menus-simple-menus](https://material.io/guidelines/components/menus.html#menus-simple-menus) as a reference as well.

Monte9:
Also adding https://www.npmjs.com/package/react-native-modal-dropdown as a reference. Thanks @xstable

And in the meantime, please check out the Picker component supported by React Native here: https://facebook.github.io/react-native/docs/picker.html

camux:
I would add this
[(https://github.com/toystars/react-native-multiple-select)]
 is very flexible

jstansbe:
Hey, was anything implemented?

iRoachie:
Not currently @jstansbe

geminiyellow:
oh~ hope we can use it tomorrow.

martinezguillaume:
https://github.com/react-native-training/react-native-elements/pull/1216

jstansbe:
@martinezguillaume that looks great!

camux:
When this component is released on beta?

martinezguillaume:
@camux Maybe the pr will be finished during the next month

Ryan-ED:
Any updates on this component? Thanks.

escalepion:
+1 !!!!!!

martinezguillaume:
Sorry guys 😔 I’ve been busy last weeks, I’ll try to do my best to finish this pr in the next days !!

iRoachie:
Same here 🙈

ggdegreat:
still waiting ..

[...]

PR #2889, opened 2021-03-20
Added new component Dropdown

venishpatidar:

Introducing a new component Dropdown

Yes

**Summary**

Dropdown picker will be added
dropdown


flyingcircle:
Hmm. I appreciate the PR. I'm not sure how in demand a dropdown list is. It seems like you'd want to use the `ListItem` that we already have to build something like this out. The other concern I have with this is that it looks out of place. It doesn't look like any of the other components.

venishpatidar:
#253 `https://github.com/react-native-elements/react-native-elements/issues/253`
It would be pretty handy as RNE lacks of dropdown, therefore it needed to install another package specifically for dropdown. Is there anything that you want to be get upgraded, removed or edited, I would like to accept any suggestion.

flyingcircle:
Yeah a dropdown would be a fine addition. But right now, it looks like a DOS menu or something. It needs to look like an RNE component. Try using `ListItem` in the component instead.

venishpatidar:
I am on it 👍.

HariAcidReign:
> Yeah a dropdown would be a fine addition. But right now, it looks like a DOS menu or something. It needs to look like an RNE component. Try using `ListItem` in the component instead.

@flyingcircle even though there is an NPM package for the dropdown menu, are you still looking to implement it using RNE? IF yes, can I be assigned to it?

venishpatidar:
@HariAcidReign I am already half way on it, Just matter of fact I was busy last few days so it had taken so long. Try solving open issues.

Issue #2879, opened 2021-03-19
Proposal for Creation of new Chip component


Dhananjayan-PN:
## **Is your feature request related to a problem? Please Describe.**

The React Native Elements UI Toolkit currently lacks a Chip component.

## **Describe the solution you'd like**

I will be creating a new Chip component that can be used as a button, for filtering results, displaying tags, and a lot of other fun ways of displaying information. The Chip component is extremely essential for a UI library.

## **Describe alternatives you've considered**

Although the Chip component is very similar to the button component, I believe it can add a lot to it with a better UI look and feel than a button. I'm also considering the addition of Dismissable Chips which are a useful component too.

## **Additional context**

My current plan is to create a basic Chip component first and then once the PR is approved I will work on adding new fun features and props to the Chip component. The final product will look very similar to the following images

![Screenshot 2021-03-19 at 3 41 26 PM](https://user-images.githubusercontent.com/46382245/111764767-92f0f100-88c9-11eb-85fb-e9d65eeef71a.png)
![Screenshot 2021-03-19 at 3 42 14 PM](https://user-images.githubusercontent.com/46382245/111764858-aef49280-88c9-11eb-9888-8dcbf13ab6ce.png)
![Screenshot 2021-03-19 at 3 43 14 PM](https://user-images.githubusercontent.com/46382245/111764966-d3506f00-88c9-11eb-8d77-4774d6fb5aeb.png)

Dhananjayan-PN:
@flyingcircle @pranshuchittora It'll be great if you can acknowledge this proposal! Thanks!

PR #2890, opened 2021-03-20
Design and Implementation of new Chip Component

Dhananjayan-PN:
## **What kind of change does this PR introduce?**

This PR introduces a new Chip component.

## **Did you add tests for your changes?**
I haven't written any tests for this new component as I'm still figuring out how the test scripts work

## **If relevant, did you update the documentation?**
I haven't updated the documentation as I believe there are more features that can be incorporated into this component.

## **Summary**
Most UI toolkits have the Chip component as it is a very elegant and useful UI element that adds to the user experience. I believe this PR aims to solve the absence of such a component. Link to Open Issue -> #2879 

I have also submitted a PR to the RNE-app repository for the showcase of the new Chip component for users to fiddle with and understand. (https://github.com/react-native-elements/react-native-elements-app/pull/173)

Below is a screenshot of the implementation:
Screenshot 2021-03-21 at 2 23 28 AM

@flyingcircle @pranshuchittora Kindly review this PR and let me know if there is anything I can add to it or make it more complete. Looking forward to a merge :)


flyingcircle:
If this is basically just a specialized button, why not just use the Button component to make it and have it inherit the ButtonProps? (and stick it under the button dir) That seems like it would simplify down the code a bit.

Dhananjayan-PN:
Great Idea! I'm working on it!

Dhananjayan-PN:
@flyingcircle I refactored the new component just as you suggested! It sure did simplify the code by quite a bit! Thanks for the suggestion! Kindly review the new commits :)

src/buttons/Chip.tsx
```diff
@@ -0,0 +1,47 @@
...
+  ViewStyle,
+  TextStyle,
+} from 'react-native';
+import { withTheme } from '../config';
+import Button from './Button';
+import { IconNode } from '../icons/Icon';
+import { Theme } from '../config/theme';
+import { TextProps } from '../text/Text';
+
+export type ChipProps = TouchableOpacityProps &
```
flyingcircle, line 15: Inherit from the `ButtonProps` instead of rebuilding.

flyingcircle (changes requested):

src/buttons/Chip.tsx
Dhananjayan-PN: Should I still export ChipProps or just keep the ButtonProps and pass that into the Chip component?

Dhananjayan-PN: Sure thing! Working on it!

Dhananjayan-PN: Done!

Dhananjayan-PN:
Refactored the ChipProps to inherit from ButtonProps, omit properties irrelevant to the Chip component, and redefine the type property.

flyingcircle:
Cool just need unit tests and documentation.

Dhananjayan-PN:
All done! Documentation and Unit Tests added! Ready to Merge! 😃

Dhananjayan-PN:
@flyingcircle Kindly review!

flyingcircle:
The only thing I'm unsure about here is shouldn't the default be that the `Chip` isn't pressable? I think the default should be that they don't do anything right?

Dhananjayan-PN:
Yes, that's true but since we're currently inheriting ButtonProps and utilizing a Button to create the Chip, the pressable characteristic is added by default. Let me see if I can override it by passing additional props. Thanks for the suggestion!

Dhananjayan-PN:
All done! Now the chip is only pressable (changes opacity in iOS/ripples in android) when you pass an onPress() property to the Chip. I overrode the TouchableComponent of the Button accordingly. Kindly review the changes :)

src/buttons/Chip.tsx
```diff
@@ -0,0 +1,28 @@
...
+  'loading' | 'loadingStyle' | 'loadingProps' | 'buttonStyle'
+> & {
+  type?: 'solid' | 'outline';
+};
+
+const Chip: React.FunctionComponent<ChipProps> = (props: ChipProps) => {
+  return (
+    <Button
+      {...props}
+      titleStyle={{ fontSize: 14, paddingHorizontal: 2 }}
```
flyingcircle, line 17: The way these props are done are not quite correct. 2 items:
1. `{...props}` should always come at the bottom so that people can overwrite the defaults.
2. `titleStyle` and other props should extend the `titleStyle` passed in as a prop so that they can extend the style and not just completely overrwrite it.

flyingcircle (changes requested):

src/buttons/Chip.tsx
Dhananjayan-PN: Thanks a lot for that! I've updated the props accordingly.

src/buttons/Chip.tsx
```diff
@@ -0,0 +1,34 @@
...
+> & {
+  type?: 'solid' | 'outline';
+};
+
+const Chip: React.FunctionComponent<ChipProps> = (props: ChipProps) => {
+  return (
+    <Button
+      titleStyle={StyleSheet.flatten([
+        props.titleStyle,
+        { fontSize: 14, paddingHorizontal: 2 },
```
flyingcircle, line 18: This is the same issue. Here, someone won't be able to overwrite `fontSize` or `paddingHorizontal`. the `props.titleStyle` has to come last.

flyingcircle (changes requested):

src/buttons/Chip.tsx
Dhananjayan-PN: Ahh okay, I see what you mean. I misunderstood how StyleSheet.flatten works (anything we add should come before what the user passes in, got it). Thanks! I've updated the props! Kindly review and let me know if there are any other issues I've overlooked.

src/buttons/Chip.tsx
```diff
@@ -0,0 +1,34 @@
...
+const Chip: React.FunctionComponent<ChipProps> = (props: ChipProps) => {
+  return (
+    <Button
+      titleStyle={StyleSheet.flatten([
+        { fontSize: 14, paddingHorizontal: 2 },
+        props.titleStyle,
+      ])}
+      buttonStyle={StyleSheet.flatten([
+        { borderRadius: 30 },
+        props.buttonStyle,
```
flyingcircle, line 22: You have to separate out `buttonStyle` and `titleStyle` from `props` or else, the style gets overwritten below when `{...props}` is used. Separate it out in the parameters list like other components.

flyingcircle (changes requested):

src/buttons/Chip.tsx
Dhananjayan-PN: All done! Kindly review

Dhananjayan-PN:
@flyingcircle Kindly review this PR for any further changes!

website/docs/props/chip.md
```diff
@@ -0,0 +1,173 @@
+## Props
+
+> Also receives all
+> [TouchableNativeFeedback](http://reactnative.dev/docs/touchablenativefeedback.html#props)
+> (Android) or
+> [TouchableOpacity](http://reactnative.dev/docs/touchableopacity.html#props)
+> (iOS) props
+
+- [Props](#props)
+- [Reference](#reference) - [`containerStyle`](#containerstyle) - [`disabled`](#disabled) - [`disabledStyle`](#disabledstyle) - [`disabledTitleStyle`](#disabledtitlestyle) - [`icon`](#icon) - [`iconContainerStyle`](#iconcontainerstyle) - [`iconRight`](#iconright) - [`linearGradientProps`](#lineargradientprops) - [`onPress`](#onpress) - [`raised`](#raised) - [`title`](#title) - [`titleProps`](#titleprops) - [`titleStyle`](#titlestyle) - [`TouchableComponent`](#touchablecomponent) - [`type`](#type) - [`ViewComponent`](#viewcomponent)
```
flyingcircle, line 10: All of these props are already listed in the `Button` component. Don't repeat them, just put a link to the other component. You can see and example of this in other components that inherit their props (like `Searchbar`)

flyingcircle (changes requested):

website/docs/props/chip.md
Dhananjayan-PN: Oh okay, I'll update that. I was thinking maybe I'll put it all in one place but yea you're suggestion sounds better!

Dhananjayan-PN: All done! Do let me know if I should leave the prop names in the Reference section as Searchbar does or not.

Dhananjayan-PN:
@flyingcircle I redid this PR and found that the lack of the postinstall script made the app to keep crashing upon launch. Kindly run `./fix.sh && tsc --p ./node_modules/react-native-elements` after installation for testing out the app. Do let me know if you know any alternatives for this. Looking forward to a merge!

PR #2891, opened 2021-03-21
Theme can now be changed by changing the theme prop

hetanthakkar1:
What kind of change does this PR introduce?

Added ability to change the theme for a component tree via state change
Fixed issue #2754


hetanthakkar1:
@flyingcircle Sorry for reopening this PR again. I have made the necessary changes and made the code much simpler. Can you please review!? I closed that PR because I accidentally messed up with git and deleted that branch. Will be cautious next time. Sorry for trouble

src/config/ThemeProvider.tsx
```diff
@@ -54,7 +54,10 @@ export default class ThemeProvider extends React.Component<
 
   static getDerivedStateFromProps(props, state) {
     const { useDark } = props;
-    if (useDark !== state.useDark) {
+    const isTheme = (theme) => {
+      return !(Object.keys(theme).length === 0 && theme.constructor === Object);
```
flyingcircle, line 58: I'm not sure I understand why this function? And I don't know what `isTheme` is even supposed to mean. What is this function trying to look for? That the theme is present? I thought we were supposed to look for whether the `theme` object changed or not.

flyingcircle (changes requested):

hetanthakkar1:
@flyingcircle  yeah! you are right. We need to detect that. I think I need to get some sleep. I am making too many silly mistakes.

flyingcircle:
CI is saying you're failing the unit tests. Please fix first.

hetanthakkar1:
@flyingcircle  I have also added `isTheme` condition in the if statement to check if the theme has been supplied. It is necessary to check as a test case in test file was failing as its style got changed in getDerivedStateFromProps because `state.theme!==props.theme` always holds true in the beginning if the theme is not supplied

hetanthakkar1:
@flyingcircle I have also added `isTheme` condition in the if statement to check if the theme has been supplied. It is necessary to check as a test case in test file was failing as its style got changed in getDerivedStateFromProps because `state.theme!==props.theme` always holds true in the beginning if the theme is not supplied

src/config/ThemeProvider.tsx
```diff
@@ -54,7 +54,13 @@ export default class ThemeProvider extends React.Component<
 
   static getDerivedStateFromProps(props, state) {
     const { useDark } = props;
-    if (useDark !== state.useDark) {
+    const isTheme = (theme) => {
```
flyingcircle, line 57: This function still doesn't make sense to me. Are we just checking `theme !== {}`. That seems a bit easier to understand than this.

flyingcircle (changes requested):

src/config/ThemeProvider.tsx
hetanthakkar1: Yeah! we are checking that if theme function is not empty

hetanthakkar1: That function is a more formal way of checking the same thing precisely

flyingcircle: Easy to read code is preferable.

hetanthakkar1: shall i replace that by theme !== {}?

hetanthakkar1:
@flyingcircle  Please review!

src/config/ThemeProvider.tsx
```diff
@@ -54,12 +54,10 @@ export default class ThemeProvider extends React.Component<
...
   static getDerivedStateFromProps(props, state) {
     const { useDark } = props;
-    const isTheme = (theme) => {
-      return !(Object.keys(theme).length === 0 && theme.constructor === Object);
-    };
+
     if (
       useDark !== state.useDark ||
-      (isTheme(props.theme) && props.theme !== state.theme)
+      (props.theme !== {} && props.theme !== state.theme)
```
arpitBhalla, line 60: We can't compare Objects in js

src/config/ThemeProvider.tsx
flyingcircle: @hetanthakkar1 I see that I was wrong on this one. Can you change it back and just add a comment explaining the code?

hetanthakkar1:
@flyingcircle Reverted the changes

hetanthakkar1:
@flyingcircle Please review! I have made the necassary changes

src/config/ThemeProvider.tsx
hetanthakkar1: @flyingcircle  done!

hetanthakkar1:
@flyingcircle Can you please review!

anthowm:
This will be available in next release ?

PR #2892, opened 2021-03-22
Fix SearchBar placeholder tsc issue

khushal87:
It **fixes the Searchbar Tsc warning for placeholder**

Not required, but the code is tested well locally
Not needed
**Summary**
error_placeholder
solved_placeholder

No.

**Other information**
Reference issue - https://github.com/react-native-elements/react-native-elements/issues/2870

Looking forward for suggestions and review.

### DIFF
--- a/src/searchbar/SearchBar.tsx
+++ b/src/searchbar/SearchBar.tsx
@@ -35,6 +35,7 @@ export type SearchBarBaseProps = {
   onChangeText?(text: string): void;
   onCancel?(): void;
   theme?: Theme;
+  placeholder?: string;
 };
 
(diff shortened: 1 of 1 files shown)


flyingcircle:
This doesn't make sense. `placeholder` is already inherited through `TextInputProps`

khushal87:
But it still gives a tsc issue. Let me find a fix for it then. @flyingcircle

flyingcircle:
No, I'm against patchwork fixes like this. There is some reason why these types aren't lining up correctly, but adding `placeholder` to SearchBar just hides the issue in a harder to find way.

khushal87:
Okay, I will work and look at some other fixes.

PR #2894, opened 2021-03-22
Removed ts-ignore tags from searchbar

hetanthakkar1:
Fixed ref type of input


flyingcircle:
I'm not sure I understand this. `ref` isn't a prop.

hetanthakkar1:
The ref props used in searchbar was not defined in the the list of props. so i defined them in the input component.

flyingcircle:
This isn't the correct fix for this. This probably means that there needs to be either a change in withTheme's return type or a fix with how `SearchBar` is using the forwardedRef.

PR #2895, opened 2021-03-22
removed @ts-ignore from Image component and migrated to functional co…

k-guy-dev:
…mponent

Removed the @ts-ignore tag and remove typescripts errors
Migrated Component to Functional Component

No, previous image component test are used

**Summary**

@ts-ignore was used in the component which may result in further type discrepancies, hence the code has to be update.
The component was a class based component therefore update it to new standard function component. 

**Other information**
Some test for other dependent component such as Avatar and Tile are throwing errors. (Writing this to get some insights on what might be the problem)

### DIFF
--- a/src/image/Image.tsx
+++ b/src/image/Image.tsx
@@ -24,29 +24,26 @@ export type ImageProps = RNImageProps & {
 };
 
 type ImageState = {
-  placeholderOpacity: Animated.Value;
+  placeholderOpacity?: Animated.Value;
 };
 
-class Image extends React.Component<ImageProps, ImageState> {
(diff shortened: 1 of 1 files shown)


flyingcircle:
I can appreciate the thought that went into this, but I also don't want to break the build. If fixing this is showing errors in other spots, then the PR needs to be combined with those fixes as well.

k-guy-dev:
Yes, that wouldn't be a good idea 😅 .  I can't seem to find what's going wrong with the other components, I'll make another PR without errors if I get the solution. Thanks.

PR #2897, opened 2021-03-22
Removed ts-ignore tags

hetanthakkar1:
Removed ts-ignore from Image, AirbnbRating, Rating, Input, Searchbar, Tooltip.


hetanthakkar1:
oops! I pushed the fetched changes by mistake. Please ignore those commits. This happened because i accidentally pushed before committing. Will take care next time. Please review @flyingcircle

arpitBhalla (commented):
See this https://github.com/react-native-elements/react-native-elements/pull/2881#discussion_r597857387

hetanthakkar1:
@arpitBhalla sorry i didn't got you!

arpitBhalla:
We don't have to add `theme` here it's added in `withTheme`

hetanthakkar1:
I think we do need to add them

hetanthakkar1:
i think the problem lies in type checking of parameters  in withTheme function
`  WrappedComponent: React.ComponentType<P & ThemeProps<T>>,
`
Assigning value to parameters is the first line of the function and that's where problem arises so i think we need to add theme, updateThme and reaplaceThme

arpitBhalla:
@hetanthakkar1 only class components have this problem...

hetanthakkar1:
> @arpitBhalla I think the problem lies in type checking of parameters in withTheme function
> ` WrappedComponent: React.ComponentType<P & ThemeProps<T>>,`
> Assigning value to parameters is the first line of the function and that's where problem arises so i think we need to add theme, updateThme and reaplaceThme

@arpitBhalla  yeah! did you get this!? why we need to include in class components!?

flyingcircle:
@arpitBhalla is correct. I do not want these types individually added to every component. A better solution is needed.

PR #2902, opened 2021-03-24
Removed @ts-ignore tags

hetanthakkar1:
Removed ts-ignore tags from various files

### DIFF
--- a/src/config/ThemeProvider.tsx
+++ b/src/config/ThemeProvider.tsx
@@ -7,9 +7,9 @@ import { FullTheme, Theme } from './theme';
 type RecursivePartial<T> = { [P in keyof T]?: RecursivePartial<T[P]> };
 
 export interface ThemeProps<T> {
-  theme: Theme<T>;
-  updateTheme: (updates: RecursivePartial<FullTheme>) => void;
-  replaceTheme: (updates: RecursivePartial<FullTheme>) => void;
+  theme?: Theme<T>;
+  updateTheme?: (updates: RecursivePartial<FullTheme>) => void;
+  replaceTheme?: (updates: RecursivePartial<FullTheme>) => void;
 }
 
 export const ThemeContext: React.Context<ThemeProps<{}>> = React.createContext({
--- a/src/input/Input.tsx
+++ b/src/input/Input.tsx
@@ -23,26 +23,27 @@ const renderText = (content, defaultProps, style) =>
     style: StyleSheet.flatten([style, defaultProps && defaultProps.style]),
   });
 
-export type InputProps = TextInputProps & {
-  containerStyle?: StyleProp<ViewStyle>;
-  disabled?: boolean;
-  disabledInputStyle?: StyleProp<TextStyle>;
-  inputContainerStyle?: StyleProp<ViewStyle>;
-  leftIcon?: any;
-  leftIconContainerStyle?: StyleProp<ViewStyle>;
-  rightIcon?: any;
-  rightIconContainerStyle?: StyleProp<ViewStyle>;
-  inputStyle?: object | any[];
-  InputComponent?: typeof React.Component;
-  errorProps?: object;
-  errorStyle?: object | any[];
-  errorMessage?: string;
-  label?: React.ReactNode;
-  labelStyle?: object | any[];
-  labelProps?: object;
-  theme?: Theme;
-  renderErrorMessage?: boolean;
-};
+export type InputProps = React.ComponentPropsWithRef<typeof TextInput> &
+  TextInputProps & {
+    containerStyle?: StyleProp<ViewStyle>;
+    disabled?: boolean;
+    disabledInputStyle?: StyleProp<TextStyle>;
+    inputContainerStyle?: StyleProp<ViewStyle>;
+    leftIcon?: any;
+    leftIconContainerStyle?: StyleProp<ViewStyle>;
+    rightIcon?: any;
+    rightIconContainerStyle?: StyleProp<ViewStyle>;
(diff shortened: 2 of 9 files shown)


hetanthakkar1:
@flyingcircle and @arpitBhalla  Do this approach seem correct?

hetanthakkar1:
@flyingcircle  Do this approach seem correct?

src/config/ThemeProvider.tsx
```diff
@@ -7,9 +7,9 @@ import { FullTheme, Theme } from './theme';
 type RecursivePartial<T> = { [P in keyof T]?: RecursivePartial<T[P]> };
 
 export interface ThemeProps<T> {
-  theme: Theme<T>;
-  updateTheme: (updates: RecursivePartial<FullTheme>) => void;
-  replaceTheme: (updates: RecursivePartial<FullTheme>) => void;
+  theme?: Theme<T>;
```
flyingcircle, line 10: Why make these optional? This is incorrect.

hetanthakkar1:
@flyingcircle I've not included ThemeProps in every component but made them optional.I think making them optional make sense because not every component need them.And alart from that I've made other changes too, please review

src/tooltip/Tooltip.tsx
```diff
@@ -44,7 +44,7 @@ const defaultProps = {
   highlightColor: 'transparent',
   withPointer: true,
   toggleOnPress: true,
-  toggleAction: 'onPress',
+  toggleAction: 'onPress' as const,
```
flyingcircle, line 47: I'm not sure I understand the necessity of this change.

src/input/Input.tsx
```diff
@@ -23,26 +23,27 @@ const renderText = (content, defaultProps, style) =>
...
-  errorProps?: object;
-  errorStyle?: object | any[];
-  errorMessage?: string;
-  label?: React.ReactNode;
-  labelStyle?: object | any[];
-  labelProps?: object;
-  theme?: Theme;
-  renderErrorMessage?: boolean;
-};
+export type InputProps = React.ComponentPropsWithRef<typeof TextInput> &
```
flyingcircle, line 26: This is closer to what I think is probably needed, but it still doesn't seem correct. It might be that there is some additional `forwardRef` that is needed to make this work, but I'm unsure about this. Did you get this idea from a particular article or website?

flyingcircle (changes requested):

src/tooltip/Tooltip.tsx
hetanthakkar1: That's necassary! it's just new syntax that needs to be written in order to do that! 
Actually that wasn't previously written and the error which might arise due to that was hidden because of ts-ignore.Basically that's just a syntax!

src/input/Input.tsx
hetanthakkar1: Actually almost similar issue was there on react-native-paper. and the PR which solved this had done the exact same thing. I think we can go with this. I'll link the PR here

hetanthakkar1:
@flyingcircle can we merge the removal of ts-ignore tag from image style in Stylesheet.flatten? and I've made changes in the slider component can you please review that as well?

flyingcircle:
You can't merge half a PR.

hetanthakkar1:
@flyingcircle  I will open a new for the same

hetanthakkar1:
@flyingcircle  I got that  React.ComponentPropsWithRef idea from here: https://github.com/callstack/react-native-paper/pull/1791/files

hetanthakkar1:
@flyingcircle  You said :
@arpitBhalla is correct. I do not want these types individually added to every component. A better solution is needed.
There are certain components which don't have the theme props. Can i include the theme props in those component? 
There are either 2 ways: making theme props optional which you said is incorrect or to include them in the component.So should is it permissible to include the theme prop in the component which are missing that prop

flyingcircle:
No, there are not just 2 ways. I am against both making Theme's props optional and adding unnecessary props to components. There are some components that make use of the default theme. I'm starting to wonder if the theming has just been setup poorly. Perhaps it is time to take a fresh look at `withTheme` and rethink how to redo theming to fix these issues. Cause this is getting overly complicated in my opinion.

hetanthakkar1:
@flyingcircle  I got your point! So currently its a little abstract as to what could be done to fix that theme prop issue so will have to leave it for now! Did you got any insight from that react native paper PR?

hetanthakkar1:
@flyingcircle  I am thinking if we can insert some default temporary theme in the withTheme method to the component which doesnt possess the theme?

src/input/Input.tsx
flyingcircle: At the very least `TextInputProps &` seems unnecessary with this addition.

hetanthakkar1: @flyingcircle we can replace TextInputProps with React.ComponentPropsWithRef<typeof TextInput> and it will work fine! Shall i make a new PR for this Input and Image!?

hetanthakkar1:
> @flyingcircle I am thinking if we can insert some default temporary theme in the withTheme method to the component which doesnt possess the theme?

@flyingcircle  the exact problem lies when the component is passed to withTheme funtion and the WrappedComponent tries to take that value.

As there is type of WrappedComponent is `React.ComponentType<P & ThemeProps<T>>`  It is required that the component possess the ThemeProps 

If we dont want the compnent to have the ThemeProps madatory nor do we want to add theme we can do this :
`  WrappedComponent: React.ComponentType<P | ThemeProps<T>>,`  and this works fine too

PR #2905, opened 2021-03-24
Removed ts-ignore from various components

hetanthakkar1:
Fixed ts-ignore issue resulted from theme prop

### DIFF
--- a/src/badge/Badge.tsx
+++ b/src/badge/Badge.tsx
@@ -9,7 +9,7 @@ import {
   ViewStyle,
 } from 'react-native';
 import { withTheme } from '../config';
-import { ThemeProps } from '../config/ThemeProvider';
+import { ThemeProps } from '../config';
 import { renderNode } from '../helpers';
 
 export type BadgeProps = {
--- a/src/badge/Badge.tsx
+++ b/src/badge/Badge.tsx
@@ -9,7 +9,7 @@ import {
   ViewStyle,
 } from 'react-native';
 import { withTheme } from '../config';
-import { Theme } from '../config/theme';
+import { ThemeProps } from '../config/ThemeProvider';
 import { renderNode } from '../helpers';
 
 export type BadgeProps = {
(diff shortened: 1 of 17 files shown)


src/buttons/Button.tsx
```diff
@@ -50,10 +51,11 @@ export type ButtonProps = TouchableOpacityProps &
     disabledStyle?: StyleProp<ViewStyle>;
     disabledTitleStyle?: StyleProp<TextStyle>;
     raised?: boolean;
-    theme?: Theme;
   };
 
-const Button: React.FunctionComponent<ButtonProps> = (props: ButtonProps) => {
+const Button: React.FunctionComponent<ButtonProps> = (
```
flyingcircle, line 56: I think you accidentally added this to the prop type instead of the component type. Or it atleast should be added to both.

src/badge/Badge.tsx
```diff
@@ -9,7 +9,7 @@ import {
   ViewStyle,
 } from 'react-native';
 import { withTheme } from '../config';
-import { Theme } from '../config/theme';
+import { ThemeProps } from '../config/ThemeProvider';
```
flyingcircle, line 12: I think `ThemeProps` should be added to the `config/index.ts` and also exported in `src/index.ts` can you change that around and then simplify the imports so that they only have to import from `../config`?

src/header/Header.tsx
```diff
@@ -84,12 +84,13 @@ export type HeaderProps = ViewProps & {
   centerContainerStyle?: StyleProp<ViewStyle>;
   leftContainerStyle?: StyleProp<ViewStyle>;
   rightContainerStyle?: StyleProp<ViewStyle>;
-  theme?: Theme;
   children?: JSX.Element[];
   elevated: boolean;
 };
 
-const Header: React.FunctionComponent<HeaderProps> = (props: HeaderProps) => {
+const Header: React.FunctionComponent<HeaderProps> = (
```
flyingcircle, line 91: Same as `Button`

flyingcircle (changes requested):

flyingcircle:
Awesome thanks!

hetanthakkar1:
@flyingcircle  Welcome!!

Issue #2854, opened 2021-03-14
Create a Toast component


maxikgreat:
# Title

Creating a new **Toast UI** component.

# Motivation

As I have been developing different apps for different purposes and have been used a wide specter of UI libraries for pretty interfaces and UX, in most cases, this type of component was really helpful to provide a convenient way to show information and make the interface more friendly for end-user.

# Propose

I think the better way to implement that functionality is to add hook ```useToast``` with basic function (and probably the only one 😉 ) - ```show```

The additional configuration may be passed throw the third parameter of that function. E.g 
```show('Ooops. Something went wrong', 'error', {config})```

or it can be globally configured with ThemeProvider:

```
<ThemeProvider
    theme={
        Toast:  {
            position: 'top' | 'bottom',
            backgroundColor: 'red',
        }
    }
>

</ThemeProvider>
```

# Design Details

The design must be fittable to the rest of the already existing components and looks similar to them.

# Inpired by 

- [react-native-paper](https://callstack.github.io/react-native-paper/index.html)
- [react-toastify](https://www.npmjs.com/package/react-toastify)
- [react-native-base](https://nativebase.io/)

So what we're waiting for? 😜

abhishekkumar08:
The idea looks really nice, looking forward to it.

flyingcircle:
It seems to me that this API is a little too complicated for this library. This library is a simple set of base components that are easily extendable. Compare this to the Compare it to the API here: https://www.npmjs.com/package/react-native-toast-message where the API, setup, styling, and extendibility are all readily apparent and easy to understand. I don't think we want to overly complicate this library.

PR #2906, opened 2021-03-24
feature: introduce new toast component

maxikgreat:
## What kind of change does this PR introduce?

This PR introduces new functionality with a couple of new components:
- **ToastProvider** as a head component created with context
- **Toast** as container for all messages 
- **Message** as component for particular message
- **useToast** as hook to return **showMessage** function

The main flow based on React Context strategy to pass params and call functions from any place of application.

Linked issue #2854 

## Did you add tests for your changes?

Yes.

## If relevant, did you update the documentation?

I've created documentation related only to this component.

## Does this PR introduce a breaking change?

The changes don't touch the previous functionality of the library.

## Next steps

Of course, that's only the beginning of developing this component, like MVP.
In the future, it'll be able to evolve with types of animation, pre-styled types of messages, extra icons, and actions.

Welcome to review 🐈


flyingcircle:
The API to this seems rather complicated for a single component. Compare it to the API here: https://www.npmjs.com/package/react-native-toast-message

As a developer I would rather choose that over having to wrap my entire component in another "Provider". I can see you put a lot of work into this, but it seems overly complicated and is solved by another library in a much simpler fashion. Multiple people have brought up making a "Toast" component, but I think this PR shows that we probably shouldn't be going down this rabbit hole.

maxikgreat:
> The API to this seems rather complicated for a single component. Compare it to the API here: https://www.npmjs.com/package/react-native-toast-message
> 
> As a developer I would rather choose that over having to wrap my entire component in another "Provider". I can see you put a lot of work into this, but it seems overly complicated and is solved by another library in a much simpler fashion. Multiple people have brought up making a "Toast" component, but I think this PR shows that we probably shouldn't be going down this rabbit hole.
I got you and understood your point. I thought that wouldn't be a bad approach to add this type of component, cause I'm pretty like libraries that completely staffed with all essential elements and have an ecosystem of own component.

Anyway, thanks for your response and feedback 👍 I gonna delete this branch

PR #2912, opened 2021-03-25
[Tooltip] Class to FC & Performance improved

### LINKED ISSUE #2210: Tooltip Direction cannot be horizontal screen
react-native 0.58
Direction cannot be horizontal screen

### LINKED ISSUE #2392: Tooltip Position is not aligned (iOS and Android)
**Explain what you did**
I am using react-native-navigation

I used the  <Tooltip> element:
```
import {Text, Tooltip} from 'react-native-elements';
...
<Tooltip
          backgroundColor={'#fff'}
          popover={<Text>Currently Accessed</Text>}>
          <FontAwesome5 size={25} name="thumbs-up" color="#252cf0" />
</Tooltip>
```

**Expected behavior**
A tooltip box should appear with the pointed part touching the icon.

**Describe the bug**
The tooltip appeared visibly higher up, with a duplicate of the child icon

**Screenshots**
*before clicking*
before click

*after clicking*
afterclick

*android*
android after

*withAndroidStatusBar changed it slightly on Android*
skipAndroidStatusbar

I am including the react-native-navigation version because I read [here](https://github.com/react-native-elements/react-native-elements/issues/1888) and [here](https://github.com/react-native-elements/react-native-elements/issues/1871) that another navigation library caused an issue with determining the right y-axis offset. I tried using the gist described [here](https://github.com/react-native-elements/react-native-elements/issues/1871#issuecomment-485255900) but it didn't help and all of the above are really focused on Android, whereas I am having this issue on iOS and Android.

| react-native-elements |    2.0.2  |
| react-native          |     0.62.2 |
| react-native-navigation    |     6.6.2 |

### LINKED ISSUE #2414: Tooltip - Need KeyboardAvoidingView Support for tooltip

**Need  react-native's"KeyboardAvoidingView" Support for tooltip.** 

- I use a text input field inside the tooltip, 

- When I activate the tooltip from the bottom of the screen, the input keyboard hides the tooltip .

- I tried using the react-native's KeyboardAvoidingView . But the tooltip was not avoiding the keyboard.

- It would be great , if the tooltip supports the react-native's "KeyboardAvoidingView"

Not yet tried

**Video of the error:**

tooltipissue2

### LINKED ISSUE #2462: Tooltip positioning issue on android still exists in version 2.0.4.
**Explain what you did**

**Expected behavior**

**Describe the bug**

**To Reproduce**

We highly recommend that you re-create the bug on [Snack](https://snack.expo.io). If not, list the steps that a reviewer can take to reproduce the behaviour:

**Screenshots**

| react-native-elements |         |
| react-native          |         |

### LINKED ISSUE #2751: Make sure component is mounted before calling setState.

### LINKED ISSUE #2782: Tooltip starts from top left corner and then take actual position of the wrapper element
**Explain what you did (Required)**

Wrapping a Text Component with Tooltip and trying to show when screen loads

**Expected behaviour (Required)**

Tooltip should take the wrapped element position as soon as screen loads

**Describe the bug (Required)**

When screen loads tooltip first displays at top left corner and the take the actual position of wrapper [element]([url](url))

**Screenshots (Required)**

**Your Environment (Required):**

| react-native-elements |     3.2.0    |
| react-native          |    0.63.2
Simulator Screen Shot - iPhone 8 - 2021-02-17 at 15 23 54
Simulator Screen Shot - iPhone 8 - 2021-02-17 at 15 23 40

     |

### LINKED ISSUE #2797: Rebuild Tooltip into a Tooltip2

### LINKED ISSUE #2923: Tooltip closeOnlyOnBackdropPress effect is not working
**Explain what you did (Required)**

 I have been used tooltip with  closeOnlyOnBackdropPress .

**Expected behavior (Required)**

Tooltip has to dismiss when only overlay text is clicked

**Describe the bug (Required)**

But. tootlip overlay (popover) UI is invisible when click tooltip.
**To Reproduce (Required)**

`
   <SafeAreaView style={styles.container}>
            <Text>Open up App.js to start working on your app!</Text>
            <View style={styles.modalContainer}>
                <Tooltip
                    popover={<Text>Hello world</Text>}
                    closeOnlyOnBackdropPress={true}
                    skipAndroidStatusBar={true}
                    toggleAction="onLongPress"
                    backgroundColor="white">
                    <Text style={styles.text}>Hello World</Text>
                </Tooltip>
            </View>
        </SafeAreaView>
`
`container: {
        flex: 1,
        alignItems: 'center',
        justifyContent: 'center',
    },
    modalContainer: {
        flexDirection: 'column',
        alignItems: 'center',
        justifyContent: 'center',
    },
    text: {
        color: 'black',
    },`

**Screenshots (Required)**
init screen
image
when click "hello world" text
image

**Your Environment (Required):**

| react-native-elements |  3.3.2       |
| react-native          |  0.63.4       |

### LINKED ISSUE #2957: Tooltip - attribute "ref" is missing typescript (manually programiccally)
**Explain what you did (Required)**

**Expected behavior (Required)**

**Describe the bug (Required)**

**To Reproduce (Required)**

We highly recommend that you re-create the bug on [Snack](https://snack.expo.io). If not, list the steps that a reviewer can take to reproduce the behaviour:

**Screenshots (Required)**

**Your Environment (Required):**

| react-native-elements |   3.3.2    |
| react-native          |    0.63.4    |

### LINKED ISSUE #3085: Tooltip render error : "top":"<<NaN>>"
Hello,

Sorry for my bad English, I'm French.
Thanks you very much for your fabulous library.

With `tooltip` in `Flatlist` (very basic use case) I have this error when clicking on the `tooltip`  :

> Invariant Violation: [9217,"RCTView",11,{"position":"absolute","top":"<<NaN>>","backgroundColor":0,"overflow":"visible"}] is not usable as a native method argument

Here is a copy of my code, although this one is really basic :

    <View>
        <Tooltip
            popover={<Text>{exerciseEquipment.name}</Text>}
            withOverlay={false}
            backgroundColor="#fff"
            width={300}
        >
            <Image
                source={{ uri: ... }}
                style={styles.image}
            />
        </Tooltip>
    <View>

I think it's the same problem, this bug being from the beginning of 2020 : [https://github.com/react-native-elements/react-native-elements/issues/2237](https://github.com/react-native-elements/react-native-elements/issues/2237)
This issue proposes to set up `getElementPosition` but I'm not sure where to put that.

Have you found a lasting solution ?
If not, it would perhaps be interesting to indicate it in the documentation because the probability of error is very very high.

**Your Environment (Required):**

| react-native-elements | 3.4.2 |
| react-native          | 0.64.2 |

arpitBhalla:
Closes #2797 with

- [ ] Component
  - [x] Functional component
  - [x] All state vars use hooks
  - [ ] Verify that all currently open Tooltip issues are resolved:
    - [ ] Fixes #2782,
    - [ ] Fixes #2751,
    - [ ] Fixes #2462,
    - [ ] Fixes #2414 
    - [ ] Fixes #2392,
    - [ ] Fixes #2210,
    - [ ] Fixes #2923
    - [ ] Fixes #2957,
    - [ ] Fixes #3085
  - [x] Uses latest animation API
  - [x] Looks sleeker: Follow Material UI styles and pattern
- [ ] Tests
- [ ] Docs


arpitBhalla:
@flyingcircle Kindly review

arpitBhalla:
@flyingcircle Please Review the code

renjfk:
This is also missing external interaction, please read my remark here: https://github.com/react-native-elements/react-native-elements/pull/2842#issuecomment-798797626

renjfk:
Another thing caught my attention from your screen cap triangle looks a bit detached which wasn't the case with class component.

arpitBhalla:
@renjfk Sure, will work on your suggestions

arpitBhalla:
> Another thing caught my attention from your screen cap triangle looks a bit detached which wasn't the case with class component.

@renjfk Can you explain this little more please🙄

arpitBhalla:
@renjfk

flyingcircle:
This looks very promising. As far as the API, the only confusing piece is that there is both an `open` prop and an `open` state which is leading to confusing developer interaction. I would lean towards having the developer decide when a Modal like this should be visible. So instead of the 2 `open`s, just have a standard `isVisible` prop, and developers can use their own functions for when to have it disappear.

arpitBhalla:
> This looks very promising. As far as the API, the only confusing piece is that there is both an `open` prop and an `open` state which is leading to confusing developer interaction. I would lean towards having the developer decide when a Modal like this should be visible. So instead of the 2 `open`s, just have a standard `isVisible` prop, and developers can use their own functions for when to have it disappear.

But then for each tooltip user have to create a state ???

renjfk:
> > This looks very promising. As far as the API, the only confusing piece is that there is both an `open` prop and an `open` state which is leading to confusing developer interaction. I would lean towards having the developer decide when a Modal like this should be visible. So instead of the 2 `open`s, just have a standard `isVisible` prop, and developers can use their own functions for when to have it disappear.
> 
> But then for each tooltip user have to create a state ???

I'm also in with this kind of logic since it's not really hard to make a wrapper for click and change visible state. Btw on that case you can wrap inside fragment instead of pressable.

src/tooltip/ToolTip2.tsx
```diff
@@ -0,0 +1,219 @@
...
+import Triangle from './Triangle';
+import { withTheme } from '../config';
+import { ScreenWidth, ScreenHeight } from '../helpers';
+
+export type ToolTip2Props = {
+  variant?: 'grow' | 'fade' | 'zoom';
+  withPointer?: boolean;
+  title?: string;
+  popover?: React.ReactElement<{}>;
+  toggleOnPress?: boolean;
```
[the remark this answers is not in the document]
arpitBhalla: Should I remove it?, I don't understand why its there

src/tooltip/ToolTip2.tsx
```diff
@@ -0,0 +1,219 @@
...
+  toggleAction?: 'onPress' | 'onLongPress';
+  width?: FlexStyle['width'];
+  containerStyle?: StyleProp<ViewStyle>;
+  pointerColor?: ColorValue;
+  onClose?(): void;
+  onOpen?(): void;
+  open?: boolean;
+  overlayColor?: ColorValue;
+  backgroundColor?: ColorValue;
+  closeOnlyOnBackdropPress?: boolean;
```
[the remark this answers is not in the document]
arpitBhalla: I don't get what it do in practical

arpitBhalla:
@renjfk Will user do this??
`js
<Tooltip onOpen={()=>setOpen(true)} onClose={()=>setOpen(false)} isVisible={open}/>
`

flyingcircle:
@arpitBhalla yes. This follows `Modal`'s pattern and since `ToolTip` is just a kind of `Modal`, it would be easiest for developers to have the same pattern. In fact, it might be better to change the prop names to match `Modal`'s `onShow`, `onDismiss`, and `visible`.

renjfk:
Indeed, one thing though it would be nice to have modal component selection prop or some sort to allow further customizations.

arpitBhalla:
@renjfk Modal only create a backdrop

src/tooltip/ToolTip2.tsx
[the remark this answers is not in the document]
renjfk: I think we can safely remove in case of stateless scenario

src/tooltip/ToolTip2.tsx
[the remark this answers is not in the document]
renjfk: I've never used it myself either but it doesn't make sense to have props doing nothing

flyingcircle:
From the developer's point of view (ie a blackbox view) this is just a `Modal`.

renjfk:
Precisely and for example a modal is not a necessity at all to have a tooltip. I'm aware that show and dismiss events are delegated from modal but this is going to be a stateless component and you can control state however you want from outside so in my humble opinion modal should be customizable/optional.

renjfk:
But then again this gets us to a point we need to ensure or provide a way to control positioning because then tooltip can still be visible without being swiping in a changing layout such as flat list. Maybe we should just extract the tooltip component only and let user to have this kind of scenario?

arpitBhalla:
@renjfk Is Backdrop Required to be added?

src/tooltip/ToolTip2.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+  React.useEffect(() => {
+    requestAnimationFrame(
+      () =>
+        element.current &&
+        element.current.measure((_x, _y, width, height, px, py) => {
+          setContainerDimentions({ width, height, px, py });
+        })
+    );
+  }, []);
```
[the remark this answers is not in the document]
arpitBhalla: It's working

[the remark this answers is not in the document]
renjfk: Are you really sure about it? Because during scrolling components inside don't get remounted and this will be only called once which is first mount and your screen absolute coordinates might be wrong.

renjfk:
@arpitBhalla Can you please elaborate a bit?

arpitBhalla:
@renjfk @flyingcircle Kindly Review

src/tooltip/ToolTip2.tsx
```diff
@@ -0,0 +1,235 @@
...
+  React.useEffect(() => {
+    Animated.timing(animation, {
+      toValue: Number(visible),
+      useNativeDriver: true,
+      duration: 200,
+    }).start();
+  }, [visible, animation]);
+
+  React.useEffect(() => {
+    let subs = true;
```
renjfk, line 85: Asking out of curiosity what's that `subs` used for? Causing error when you change state too quickly or something?

src/tooltip/ToolTip2.tsx
```diff
@@ -0,0 +1,235 @@
...
+    requestAnimationFrame(
+      () =>
+        containerElement.current &&
+        containerElement.current.measure((_x, _y, width, height, px, py) => {
```
renjfk, line 89: You can replace with optional chaining for the sake of elegancy

src/tooltip/ToolTip2.tsx
```diff
@@ -0,0 +1,235 @@
...
+    return () => {
+      subs = false;
+    };
+  }, [visible]);
+  React.useEffect(() => {
+    let subs = true;
+    requestAnimationFrame(
+      () =>
+        tooltipElement.current &&
+        tooltipElement.current.measure((_x, _y, width, height) => {
```
renjfk, line 104: You can replace with optional chaining for the sake of elegancy

src/tooltip/ToolTip2.tsx
```diff
@@ -0,0 +1,235 @@
...
+          if (subs) {
+            setTooltipDimension({ width, height });
+          }
+        })
+    );
+    return () => {
+      subs = false;
+    };
+  }, [visible]);
```
renjfk, line 113: Is there any specific reason why these are handled within separate effect? I think they can be merged since they both have exactly same dependencies.

renjfk (changes requested):

src/tooltip/ToolTip2.tsx
arpitBhalla: This will not allow to update the state if component gets unmounted

src/tooltip/ToolTip2.tsx
arpitBhalla: I will surely do this, but if possible can you please run this on your device and let me know if any weird thing happened, I tested on lists too,

src/tooltip/ToolTip2.tsx
```diff
@@ -0,0 +1,226 @@
...
+    requestAnimationFrame(() =>
+      containerElement?.current?.measure((_x, _y, width, height, px, py) => {
+        if (subs) {
+          setContainerDimensions({ width, height, px, py });
+        }
+      })
+    );
+
+    requestAnimationFrame(() =>
+      tooltipElement?.current?.measure((_x, _y, width, height) => {
```
renjfk, line 95: First optional chain is unnecessary.

src/tooltip/ToolTip2.tsx
```diff
@@ -0,0 +1,226 @@
...
+      toValue: Number(visible),
+      useNativeDriver: true,
+      duration: 200,
+    }).start();
+  }, [visible, animation]);
+
+  React.useEffect(() => {
+    let subs = true;
+    requestAnimationFrame(() =>
+      containerElement?.current?.measure((_x, _y, width, height, px, py) => {
```
renjfk, line 87: First optional chain is unnecessary.

renjfk (changes requested):

src/tooltip/ToolTip2.tsx
arpitBhalla: Can you please test on your device

renjfk: Sure I'll give it a try later on but my point here is a ref itself can never be null nor undefined so `ref?.current` is redundant

arpitBhalla: Done

arpitBhalla: @renjfk Does it working properly??

arpitBhalla:
Needs more calculations & Mathematics, will do it soon

src/tooltip/Tooltip.tsx
```diff
@@ -1,357 +1,224 @@
...
-          onShow={onOpen}
+        <Pressable
+          {...(!closeOnlyOnBackdropPress && { [toggleAction]: toggleToolTip })}
         >
-          {this.renderModalContent()}
-        </ModalComponent>
+          {children}
+        </Pressable>
+        <Modal transparent visible={visible}>
+          <TouchableWithoutFeedback onPress={toggleToolTip}>
```
pranshuchittora, line 137: Why not `Pressable`

arpitBhalla: I missed

src/tooltip/Tooltip.tsx
```diff
@@ -1,357 +1,224 @@
...
+};
+
+const styles = StyleSheet.create({
+  content: {
+    position: 'relative',
+  },
+  tooltip: {
+    position: 'absolute',
+    top: 10,
+    left: -10,
```
pranshuchittora, line 209: Any reason for this top and left ?

src/tooltip/Tooltip.tsx
```diff
@@ -1,357 +1,224 @@
...
+  },
+  label: {
+    color: '#fff',
+  },
+  tooltipView: {
+    padding: 10,
+    maxWidth: 200,
+    width: ScreenWidth / 2,
```
pranshuchittora, line 217: What about landscape mode ?

arpitBhalla: Screenshot_1623415930
Screenshot_1623415934

arpitBhalla: There is issue with Modal on web, Can we make `tooltip.web.tsx` and do without it?

pranshuchittora: No, because then users have to change the configs here and there. `.native.js` `.web.js`

arpitBhalla: https://github.com/expo/expo/issues/7034

renjfk: What happens when you switch back to portrait w/o closing the app, does it still work fine?

arpitBhalla: Not now, I will add a listener

arpitBhalla: @renjfk 
rotate

flyingcircle (commented):
I don't know that I have much more to say. I would test the demo app, but I wouldn't do much more than your gif already shows. This code is so much cleaner and nicer looking than the old tooltip though. Fantastic job! I'm sure once we start doing more testing around the edges, we might find some more stuff to deal with, but this is great looking

src/tooltip/Tooltip.tsx
```diff
@@ -1,357 +1,223 @@
...
-        </ModalComponent>
+          {children}
+        </Pressable>
+        <Modal transparent visible={visible}>
+          <TouchableWithoutFeedback onPress={toggleToolTip}>
+            <Animated.View
+              style={[
+                StyleSheet.absoluteFillObject,
+                {
+                  backgroundColor: overlayColor || '#b4b0b087',
```
pranshuchittora, line 144: Can we change the default overlay colour to transparent

src/tooltip/Tooltip.tsx
```diff
@@ -1,357 +1,223 @@
...
+  withPointer = true,
+  overlayColor,
+  backgroundColor = '#616161',
+  pointerColor,
+  toggleAction = 'onPress',
+  onShow = () => {},
+  onDismiss = () => {},
+  visible = false,
+  width: propWidth = 200,
+  closeOnlyOnBackdropPress = false,
```
pranshuchittora, line 49: Is it possible to avoid the backdrop at all. Usually, tooltip shouldn't block any interaction with other components.

src/tooltip/Tooltip.tsx
```diff
@@ -1,357 +1,223 @@
...
-      <View style={TooltipHighlightedButtonStyle}>{this.props.children}</View>
-    );
-  };
-
-  renderHighlightedButton = () => {
-    const { closeOnlyOnBackdropPress } = this.props;
-    if (closeOnlyOnBackdropPress) {
-      return this.renderTouchableHighlightedButton();
+const ToolTip2: React.FunctionComponent<ToolTip2Props> = ({
+  children,
```
pranshuchittora, line 37: Can we add support for showing tooltip only on hover. 

Like this

Screenshot 2021-06-12 at 5 25 03 PM

arpitBhalla: Can be done on web only

pranshuchittora: Can we use pressable state onPressIn and hide it on press out

arpitBhalla: I will try

src/tooltip/Tooltip.tsx
```diff
@@ -1,357 +1,223 @@
...
+      (width > ScreenWidth / 2
+        ? (width || 0) / 2 - (tooltipWidth || 0) / 2
+        : 0) +
+      (px < 10 ? 10 : 0) +
+      (width < 30 ? -20 : 0);
+    if (horizontal > ScreenWidth - tooltipWidth) {
+      horizontal = ScreenWidth - (tooltipWidth || 0);
     }
+    let top = py + (ScreenHeight / 2 < py ? -toolTipHeight - height - 28 : 0);
+    return { top, [I18nManager.isRTL ? 'right' : 'left']: horizontal };
```
pranshuchittora, line 123: Can you explain this entire function via code comments.

src/tooltip/Tooltip.tsx
```diff
@@ -1,357 +1,223 @@
...
+              style={[
+                styles.tooltip,
+                {
+                  opacity: animation,
+                },
+              ]}
+            >
+              {withPointer ? (
+                <Triangle
+                  isDown={ScreenHeight / 2 < containerDimensions.py}
```
pranshuchittora, line 163: `position` prop should be defined by the user 
Something like this -> https://material-ui.com/components/tooltips/#positioned-tooltips

arpitBhalla: Like this?
A

pranshuchittora:
Able to make this work in `ios` but some positioning issues

https://user-images.githubusercontent.com/32242596/121809910-57718300-cc7c-11eb-89a5-9c03c50d4fb4.mov

src/tooltip/Tooltip.tsx
```diff
@@ -1,357 +1,223 @@
...
+      Dimensions.removeEventListener('change', calcDim);
+    };
+  }, []);
+
+  React.useEffect(() => {
+    calcDim();
+  }, [visible]);
+
+  const ToolTipStyle = () => {
+    const { px, width, height, py } = containerDimensions;
```
pranshuchittora, line 107: I recommend using platfrom inputs for this decisioning 
https://github.com/react-native-elements/react-native-elements/blob/91d582a19ba3e34b078740fc374a20a8938a3f6b/src/tooltip/Tooltip.tsx#L284

arpitBhalla: Done

src/tooltip/Tooltip.tsx
```diff
@@ -1,357 +1,223 @@
...
+              {withPointer ? (
+                <Triangle
+                  isDown={ScreenHeight / 2 < containerDimensions.py}
+                  style={{
+                    position: 'absolute',
+                    borderBottomColor: pointerColor || backgroundColor,
+                    top:
+                      containerDimensions.py -
+                      containerDimensions.height *
+                        (ScreenHeight / 2 < containerDimensions.py ? 2 : 0) -
```
pranshuchittora, line 170: Can you abstract conditional checks out of the rendering pipeline

Issue #2900, opened 2021-03-23
Use custom moderateScale in the application


khushal87:
Currently, we are using a package `react-native-size-matters` for scaling the sizes of the Text and other components which is using the normalize function. Therefore, I have a solution that won't use the dependant package and work on every device. This will be dependant only on Native Text provided by React Native. Can I work on it @flyingcircle ? 
This will reduce the 3rd party dependencies in the application.

PR #2913, opened 2021-03-25
Remove react-native-size-matters dependency

khushal87:
Added a custom normalizeText which won't use the react-native-size-matters for the same.

Not required
Not required
**Summary**
This makes the app to be not dependant on the package  react-native-size-matters

No

**Other information**
NA

Reference issue - #2900


flyingcircle:
Is this just copying the library's code over to our library? I don't see the purpose of doing this.

khushal87:
I didn't copy this code @flyingcircle . Almost all of the application uses the same implementation with our without dependency. And even all the package uses same code to implement the feature. Therefore I didn't intend this at all. Rest is on you. You can take appropriate decisions. Thanks.

Issue #2591, opened 2020-09-20
Button(Props): Make the icon prop more intuitive and dynamic


pranshuchittora:
**Description**
The `icon` position props come with a huge surface area. The idea is to decrease the APIs surface area by introducing more functionalities.

**Proposed Change**
Instead of `iconRight` and default left position of the `icon`, we can introduce a new prop called `iconPosition`. 
And the usage looks might look something like this :arrow_down: 
```js
iconPosition: = "left" | "right" | "top" | "bottom"
```

The following functionality can be achieved by using the `flex` container for the text and the icon and changing it's the direction to `row | row-reverse | column | column-reverse` to achieve all the possible combinations.

flyingcircle:
I wonder if it might be simpler to or possible to have button follow the same general pattern as `Card` and `ListItem` have now. Something like:

```js
<Button>
  <Button.Icon />
  <Button.Title />
</Button>
```

This leaves greater flexibility and customization for users without having to sacrifice ease of use. And it also helps prevent prop bloat.

pranshuchittora:
Looks great, this can unlock new possibilities of a discrete set of children. May be an image btn

laxmy:
Hi @pranshuchittora , Can I take this up?

HariAcidReign:
Hi @flyingcircle , this looks like something I can work on..Is it available and can I be assigned to it?

HariAcidReign:
Working on it!

vanya-24:
Hi @flyingcircle, 
Can I also work on this issue as I found this issue to get started with, Or can you guide me with to get started? 
Thank you😊

flyingcircle:
@vanya-24 go for it. I usually try to take the best from multiple solutions. someone can help you get started in the slack channel.

vanya-24:
Thanks, working on it

hetanthakkar1:
> I wonder if it might be simpler to or possible to have button follow the same general pattern as `Card` and `ListItem` have now. Something like:
> 
> ```js
> <Button>
>   <Button.Icon />
>   <Button.Title />
> </Button>
> ```
> 
> **And it also helps prevent prop bloat.**

We are just replacing iconRight with iconPosition so do u still think it will create prop bloat!? And I think the main advantage of using this iconPosition method as opposed to child method is that the users which are currently using iconRight will have to modify their code a lot as compared to just replacing the iconRight with iconPosition.

khushal87:
[...]

PR #2914, opened 2021-03-25
Replaced iconRight with iconPosition

hetanthakkar1:
Solved issue #2591


src/buttons/Button.tsx
```diff
@@ -42,7 +42,7 @@ export type ButtonProps = TouchableOpacityProps &
     containerStyle?: StyleProp<ViewStyle>;
     icon?: IconNode;
     iconContainerStyle?: StyleProp<ViewStyle>;
-    iconPosition?: string;
```
arpitBhalla: don't remove this

hetanthakkar1: okay!

hetanthakkar1: why?

arpitBhalla: it will break code

PR #2917, opened 2021-03-26
remove ts-ignore error

Enesklmc:
i guess one more left for searchBar //@ts-ignore error


flyingcircle (changes requested):
I think what we want here is `React.ComponentPropsWithRef`. See `Input` for an example.

Enesklmc:
okay i've committed with React.ComponentPropsWithRef<typeof TextInput>

flyingcircle (approved):

PR #2924, opened 2021-03-27
show raised & reverse icons in docs

Enesklmc:
Upgrading image. 
Removing space at bottom-right in image and demonstrating raised & reverse button

icons

to

icons


pranshuchittora:
How did you achieve the reverse functionality. There's no such prop.

pranshuchittora:
Thanks

Issue #2878, opened 2021-03-18
SearchBar typescript props missing


kickbk:
I import SearchBar and need to create a ref for it:
```
import { SearchBar, SearchBarProps } from 'react-native-elements';
const searchFieldRef = useRef<SearchBarProps>(null);
const blurSearch = useCallback(() => {
   searchFieldRef.current?.blur();
  }, [searchFieldRef]);

<SearchBar ref={searchFieldRef} onPress={blurSearch} ...
```
I get a typescript error on:
1. `searchFieldRef.current?.blur();` - it doesn't know about blur.
2. `ref={searchFieldRef}` on the SearchField itself

The first issue I can, in an ugly way, fix by adding the methods to the typescript definition like so:

```
type MissingProps = {
  focus: () => void;
  blur: () => void;
  clear: () => void;
  cancel: () => void;
}
const searchFieldRef = useRef<SearchBarProps & MissingProps>(null);
```

I prefer this to be handled correctly of course.
But one thing I am not sure how to fix is `ref={searchFieldRef}`

Could you please advise how to properly import the typescript definition of this component?
Thanks!

darrylyoung:
There are other issues with the `SearchBar` types, too. For example, the [docs](https://reactnativeelements.com/docs/searchbar#props) say I can use `placeholder`, for example, but I'm getting a warning that it's not a valid prop. Looking through the types shipped, it looks like this is missing.

The codebase only just transitioned to TypeScript so I'm guessing it's just a small bug that got through and one that'll be fixed soon. The same for your case, too.

Aymer-el:
Hello,

First I would like to say I am new here. So, I might be totally out of the subject :).

Regarding your questions @darrylyoung I had a discussion about it with @flyingcircle. 

Identification issue: 
I think our issue is just typing as you said too. 
It is true only our component has the action typed (blur etc) and not the type itself. 

Action: 
I can place the type (blur, focus, clear, render, cancel) on the SearchBarProps. as well as adding the placeholder and value props (of type string).

Thanks you

darrylyoung:
Hi, @Aymer-el. 👋 

Thanks for the message.Yeah, that sounds good. I'm not sure internally how the `Searchbar` component is written – that is, if it's just a wrapper around a `TextInput`, for example – but I also noticed that it doesn't accept any accessibility props such as `accessibilityRole="search"` either.

Aymer-el:
[...]

PR #2925, opened 2021-03-28
searchBar props missing : #2878

singhakshita:
fixes the missing props, and recognised if the input box is already blurred,clear,or focused also initialised missing placeholder prop for the issue #2878 

no,not needed for now

no ,they are already present in documentation

**Summary**
placeholder was not available as a state also clear, focused, blurred  are just "on action" events but now they have been added as a state.

No

**Other information**
due to some updation changes yarn.lock and package.json has also been changed a bit , these change are irrevelant for this issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "hoist-non-react-statics": "^3.3.2",
     "lodash.isequal": "^4.5.0",
     "opencollective-postinstall": "^2.0.3",
+    "react-native-elements": "^1.0.0-beta8",
     "react-native-ratings": "^7.3.0",
     "react-native-size-matters": "^0.3.1"
   },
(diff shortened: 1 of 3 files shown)


singhakshita:
i was just not too sure about the placeholder props, also its not mentioned in main issue definition .so i removed it from the commit , but please once check the first commit also. this include on focus,onclear,on blurr action types missing props additions which is sure and tested by my side.

package.json
```diff
@@ -44,6 +44,7 @@
     "hoist-non-react-statics": "^3.3.2",
     "lodash.isequal": "^4.5.0",
     "opencollective-postinstall": "^2.0.3",
+    "react-native-elements": "^1.0.0-beta8",
```
flyingcircle, line 47: Don't change the the package.json or yarn.lock.

flyingcircle (changes requested):

package.json
singhakshita: i didn't change it deliberately ,it changed automatically after yarn install . yeah sure  i will undo it

singhakshita:
sorry i had to create an another pull request , as i was unable to undo yarn and json chnages on this one

package.json
arpitBhalla: do `git add src/searchbar/SearchBar-default.tsx`

Issue #2929, opened 2021-03-30
Strict TypeScript 


flyingcircle:
A good goal for all TS libraries is to get to a point where we can turn on 'strict' mode in the `ts.config`. If you do so as of the time I'm writing this ticket, it shows that there about 170 errors when 'strict' mode has been turned on.

So, there's a lot of possible work here so no need to submit a PR solving everything (but please don't solve just 1 thing either. Try to solve a good chunk before submitting a PR). Don't submit a PR that actually turns on 'strict' mode, just the fixes. Once everything has been fixed, then we'll turn it on.

To turn it on temporarily change the following line in the `tsconfig.json` to `true`
```
"strict": false
```

then run `tsc`. And you should see the additional errors that pop up.

Yunicorn228:
@flyingcircle Hi Jeremy, I'm trying to participate in this project in the Google Summer of code, however, I can not join the slack channel, I'm fairly new to open source projects and really need some help.

abdulsamad:
Hi Jeremy, Is this issue resolved or can I give it a try? Is somebody working on it?
@flyingcircle

PR #2930, opened 2021-03-30
Strict TypeScript #2929

arpitBhalla:
- [x] Button
- [x] FAB
- [x] Checkbox
- [x] Badge
- [x] Divider
- [x] Icon
- [x] Input
- [x] LinearProgress
- [x] SearchInput (except iOS SearchBar)
- [x] Social Icon
- [x] Pricing
- [x] Tile
- [x] ThemeProvider
- [x] withTheme

### DIFF
--- a/src/buttons/ButtonGroup.tsx
+++ b/src/buttons/ButtonGroup.tsx
@@ -63,7 +63,7 @@ const ButtonGroup: RneFunctionComponent<ButtonGroupProps> = (props) => {
     textStyle,
     selectedTextStyle,
     selectedButtonStyle,
-    underlayColor = theme.colors.primary,
+    underlayColor = theme?.colors?.primary,
     activeOpacity,
     onHideUnderlay,
     onShowUnderlay,
@@ -81,7 +81,7 @@ const ButtonGroup: RneFunctionComponent<ButtonGroupProps> = (props) => {
     innerBorderStyle &&
     Object.prototype.hasOwnProperty.call(innerBorderStyle, 'width')
   ) {
-    innerBorderWidth = innerBorderStyle.width;
+    innerBorderWidth = innerBorderStyle.width as number;
   }
   return (
     <View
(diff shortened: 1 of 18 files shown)


src/buttons/FAB.tsx
```diff
@@ -83,7 +84,7 @@ const FAB: RneFunctionComponent<FABProps> = ({
         ])}
         titleStyle={[
           styles.label,
-          { color: theme.colors.white },
+          { color: theme && theme.colors && theme.colors.white },
```
flyingcircle, line 87: Might look nicer to use optional chaining: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

flyingcircle (changes requested):

src/buttons/ButtonGroup.tsx
```diff
@@ -220,7 +220,7 @@ const styles = StyleSheet.create({
   },
   verticalContainer: {
     flexDirection: 'column',
-    height: null,
+    height: undefined,
```
flyingcircle, line 223: This actually changes the behavior of the component. We should either turn off the null check in the tsconfig or find a different solution.

flyingcircle (changes requested):

src/buttons/ButtonGroup.tsx
arpitBhalla: I tried in demo app, it was working fine, but if it really results in some weired behaviour we can add  `"strictNullChecks": false`  in the end

flyingcircle: Yeah, it might just be a corner case, but I do know that `width` and `height` being set to `null` have a special meaning apart from `undefined`. At least for TS changes, it safest to try not to change any of the JS code. There might be a Stackoverflow though that talks about how to handle this situation.

arpitBhalla: I will check for best solution

PR #2931, opened 2021-03-30
fixed:strict typescript errors

abhishekkumar08:
- [x] Card
- [x] Checkbox
 
Refr: #2929

### DIFF
--- a/src/card/CardImage.tsx
+++ b/src/card/CardImage.tsx
@@ -10,7 +10,7 @@ const CardImage: RneFunctionComponent<ImageProps> = ({ style, ...props }) => {
 
 const styles = StyleSheet.create({
   image: {
-    width: undefined,
+    width: null,
     height: 150,
   },
 });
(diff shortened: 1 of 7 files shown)


src/card/CardImage.tsx
```diff
@@ -10,7 +10,7 @@ const CardImage: RneFunctionComponent<ImageProps> = ({ style, ...props }) => {
 
 const styles = StyleSheet.create({
   image: {
-    width: null,
+    width: undefined,
```
arpitBhalla, line 13: it must be null

abhishekkumar08: okay, then finding a better sloution for this.

arpitBhalla: Muzhe bhi batana😅

flyingcircle, line 13: This is changing the behavior of the component. Either we should turn on ignoring null checks or find a different solution.

flyingcircle (changes requested):

src/card/CardImage.tsx
abhishekkumar08: Okay would gonna do that!

abhishekkumar08:
@flyingcircle Please review. Thanks

Issue #2581, opened 2020-09-18
It is recommended that the styles of ImageComponent and View components including children should be separated in the Image composition


flycj:
**Is your feature request related to a problem? Please Describe.**
It is recommended that the styles of ImageComponent and View components including children should be separated in the Image composition. 
Avoid the style property from affecting both ImageComponent and View components

A clear and concise description of what the problem is. 
I set the boxShadow effect in the style property of the Image component to have no effect, so I set the boxShadow effect on the containerStyle property to make it effective.
But in this case, I must set overflow: visible; in containerStyle.
Then, I want to set an Icon in the children of the Image to be located in the center of the Image, which will become very troublesome

```
   <Image style={styles.image}  containerStyle={styles.container} {...attributes}>
      <Icon style={styles.video} name="video" size={50} color="#ef4b4b" />
      {children}
    </Image>
```
Any styles set by the style property of my Image will act on both the ImageComponent and the children's View components, this creates conflict

**Describe the solution you'd like**
It is recommended that the styles of ImageComponent and View components including children should be separated in the Image composition. 

**Describe alternatives you've considered**
My current solution is to set the following again in the style attribute
```
const styles = {
  image: {
     position: 'absolute',
     flex: 1,
     top: 0,
     bottom: 0,
     width: '100%',
     alignItems: 'center',
     justifyContent: 'center',
  }
}
```
I think the style is set separately in ImageComponent, and then childrenStyle is set on the View, so that the separation will not cause style confusion

PR #2933, opened 2021-03-30
bugfix: extra style prop for image children container

maxikgreat:
## Motivation and solution

I solve the problem of opened issue #2581. There are no breaking changes of project and even of component 😄 
Just I've added an extra style property for ```Image``` to avoid collision between children and image styles.

Btw, I didn't expect to see an extra wrapper for children (it's redundant from my perspective) but anyway I solve the problem based on already existing functionality and expectations of constant users.

## Did you add tests for your changes?

I've extended one test case to ensure props pass correctly.

## If relevant, did you update the documentation?
Yep.

### DIFF
--- a/src/avatar/__tests__/__snapshots__/Avatar.js.snap
+++ b/src/avatar/__tests__/__snapshots__/Avatar.js.snap
@@ -731,13 +731,8 @@ exports[`Avatar Component should apply values from theme 1`] = `
       />
     </View>
     <View
-      style={
-        Object {
-          "flex": 1,
-          "height": undefined,
-          "width": undefined,
-        }
-      }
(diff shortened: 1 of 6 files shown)


vinceniko:
I think this change broke my images after I updated to 3.4.0. The images do not get rendered anymore. The React Native Image component still works however. Below is the relevant Elements' Image Component.

```javascript
<Image
  source={{ uri: uri }}
  style={{
    width: "100%",
    height: width * imageScale * (mediaDims.height / mediaDims.width),
  }}
/>
```

arpitBhalla:
Hey @vinceniko Can you please create new issue, with the information of problem that you are facing

PR #2935, opened 2021-03-31
Fix typescript strict-mode issues

khushal87:
This PR fixes some of the critical tsc strict mode issues in the package. The components I have fixed the issues for are:
This PR also fixes issue #2954 related to textStyle on Badge
- [x] Switch.tsx
- [x] Text.tsx
- [x] Header.tsx
- [x] Switch.tsx
- [x] Searchbar-ios
- [x] ListItem

Reference issue - #2929


abhishekkumar08:
@khushal87 check #2931 card component is there.

khushal87:
No worries I will remove my changes @abhishekkumar08

khushal87:
@flyingcircle Please review.

src/buttons/ButtonGroup.tsx
```diff
@@ -220,7 +220,6 @@ const styles = StyleSheet.create({
   },
   verticalContainer: {
     flexDirection: 'column',
-    height: null,
```
flyingcircle, line 223: null means something, you can't just remove it.

src/buttons/__tests__/__snapshots__/ButtonGroup.js.snap
```diff
@@ -1342,7 +1342,7 @@ exports[`ButtonGroup Component vertical ButtonGroup 1`] = `
       "borderRadius": 3,
       "borderWidth": 1,
       "flexDirection": "column",
-      "height": null,
```
flyingcircle, line 1345: same here.

src/card/CardImage.tsx
```diff
@@ -10,7 +10,6 @@ const CardImage: RneFunctionComponent<ImageProps> = ({ style, ...props }) => {
 
 const styles = StyleSheet.create({
   image: {
-    width: null,
```
flyingcircle, line 13: here

src/list/ListItem.tsx
```diff
@@ -30,6 +30,7 @@ export type ListItemProps = TouchableHighlightProps & {
   ViewComponent?: typeof React.Component;
   linearGradientProps?: any;
   children?: any;
+  theme?: FullTheme;
```
flyingcircle, line 33: This shouldn't be necessary. There are other ways to add theme without directly adding it to the component (There are other components that do this)

src/list/ListItemInput.tsx
```diff
@@ -32,14 +32,10 @@ const styles = StyleSheet.create({
   inputContainer: {
     flex: 1,
     borderBottomWidth: 0,
-    width: null,
```
flyingcircle, line 35: All of these

flyingcircle (changes requested):

src/list/ListItem.tsx
flyingcircle: Look at `Input` and how it handles the `theme` prop.

khushal87: I tried the same way, but that doesn't fix the issue, probable reasons are the interface extending other secondary components in the case of ListItem. @flyingcircle

khushal87: fixed @flyingcircle. Please review

src/avatar/Avatar.tsx
```diff
@@ -70,7 +70,7 @@ const AvatarComponent: Avatar = ({
   imageProps,
   placeholderStyle,
   renderPlaceholderContent,
-  ImageComponent = RNImage,
```
arpitBhalla, line 73: Why Remove?

src/searchbar/SearchBar-ios.tsx
```diff
@@ -188,8 +188,12 @@ class SearchBar extends Component<
           }}
           inputContainerStyle={StyleSheet.flatten([
             styles.inputContainer,
-            { backgroundColor: theme.colors.platform.ios.searchBg },
-            hasFocus && { marginRight: this.state.cancelButtonWidth },
+            { backgroundColor: theme?.colors?.platform?.ios?.searchBg },
+            hasFocus && {
+              marginRight: this.state.cancelButtonWidth
+                ? this.state.cancelButtonWidth
+                : 0,
```
arpitBhalla, line 195: this.state.cancelButtonWidth || 0

src/switch/switch.tsx
```diff
@@ -22,18 +23,25 @@ const Switch: React.FunctionComponent<SwitchProps> = ({
...
+      : color;
 
-  const onTintColor =
+  const onTintColor: ColorValue =
     Platform.OS === 'ios' || !disabled
       ? switchedOnColor
-      : theme.colors.disabled;
+      : theme?.colors?.disabled
+      ? theme.colors.disabled
+      : '';
```
arpitBhalla, line 38: Why are you doing in this way?

src/avatar/Avatar.tsx
arpitBhalla: @flyingcircle What do you think?

khushal87: Hey, @arpitBhalla I am passing in ImageComponent, it's just that RNImage is wrapped in the types defined above. This was tested on rne-app, you can also do that and let me know if it doesn't work. Also, we still have ImageComponent as props to the component.

arpitBhalla: Hey @khushal87, according to me, adding types will not ensure default value of `ImageComponent`.

khushal87: So should I add the default value @arpitBhalla ?

flyingcircle: @arpitBhalla is correct, type definitions do not affect the actual behavior of the code. The reason this still works is that the Image component defaults to the same thing.

khushal87: @flyingcircle Got it, sir, will keep it in mind. The behavior was the same so that made me confused too. Thanks for the review

khushal87:
@pranshuchittora Please review.

flyingcircle:
conflicting file here.

khushal87:
@flyingcircle Fixed, please review

PR #2944, opened 2021-04-03
added new component Dropdown

venishpatidar:
This PR will introduces a new component Dropdown to RNE.

Yes all changes are tested.

Yes Documents have been updated

**Summary**
There will be new component to Dropdown to RNE which was long awaited request, and need not to be installed from other npm packages separately. Also this dropdown can be act as Accordion  [https://media.geeksforgeeks.org/wp-content/uploads/20190522140003/accordian.gif](url) 

No

dropdown


venishpatidar:
@flyingcircle, upgraded it.

flyingcircle:
Very nice! Yes, this is more like what I was thinking. I'm wondering though, it looks like you copied a good chunk of the code over. Why just use `ListItem` as is? Is there a compelling reason that We need `DropdownItem` and the associated components? Couldn't people just use `ListItem`?

arpitBhalla:
I have a doubt, it is not called `dropdown`,  It can be Expandable or Accordion

venishpatidar:
@flyingcircle yes, there is a reason, I was thinking as it would add more flexibility and rigidity. As further updated or modification to `ListItem` would not cause any bugs in dropdown. Also other way it would be more flexible to be change anything in particular.

venishpatidar:
@arpitBhalla `Expandable` is great suggestion.

venishpatidar:
@flyingcircle what's your opinion on should it need to be changed to `Expandable` or not.

flyingcircle:
Decided to go with #2953 as it fit in a bit more with existing code.

PR #2952, opened 2021-04-08
Strict Typescript Slider

arpitBhalla:
I added a listener for the getting current value, earlier it was using `__getValue()` method, but wasn't in TS

- [x] Slider


arpitBhalla:
@flyingcircle Please Review

flyingcircle:
Looks good thanks

PR #2962, opened 2021-04-13
Support for Font Awesome 5 Pro

den1n:

Icon component changes:

- Added support for the "font-awesome-5-pro" value for the **type** prop.
- Added **light** prop.

Updated the list of available icon sets.
Added a description of the **light**.

No.

### DIFF
--- a/src/helpers/getIconStyle.tsx
+++ b/src/helpers/getIconStyle.tsx
@@ -21,8 +21,10 @@ export default (type: IconType, extraProps: any) => {
     case 'font-awesome':
       return {};
     case 'font-awesome-5':
+    case 'font-awesome-5-pro':
       return {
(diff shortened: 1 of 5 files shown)


flyingcircle:
Is there a way to actually test this? I don't have a pro account.

den1n:
I think there is no legal way without a Pro account.

The **Icon** component is based on the **react-native-vector-icons** which supports the FontAwesome5Pro font.

* https://github.com/oblador/react-native-vector-icons/blob/master/FontAwesome5Pro.js
* https://github.com/oblador/react-native-vector-icons/blob/master/FONTAWESOME5.md#upgrading-to-pro

Just examine the changes I made and trust the **react-native-vector-icons** with the rest.

PR #2975, opened 2021-04-19
Fixes issues #2878

### LINKED ISSUE #2878: SearchBar typescript props missing

arpitBhalla:
- [x] Fixed #2878

### DIFF
--- a/src/buttons/Button.tsx
+++ b/src/buttons/Button.tsx
@@ -51,6 +51,7 @@ export type ButtonProps = TouchableOpacityProps &
     disabledStyle?: StyleProp<ViewStyle>;
     disabledTitleStyle?: StyleProp<TextStyle>;
     raised?: boolean;
+    iconPosition?: 'left' | 'right' | 'top' | 'bottom';
   };
 
 const Button: RneFunctionComponent<ButtonProps> = (props) => {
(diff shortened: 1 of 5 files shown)


arpitBhalla:
@flyingcircle Please Review

PR #3024, opened 2021-05-10
Revert "Support for Font Awesome 5 Pro"

flyingcircle:
Reverts react-native-elements/react-native-elements#2962


flyingcircle:
Found out that expo doesn't support Font Awesome Pro. If people need it, it's better to just go through https://github.com/FortAwesome/react-native-fontawesome

Issue #94, opened 2016-11-17
Floating labels on FormInput


lefnire:
Consider adding a `floatingLabel={BOOL}` prop on `<FormInput>`. Examples at [js.coach](https://js.coach/all/react-native-floating-label-text-input?search=floating%20labels); could possibly make one of those projects a RNE dep, in case the dev is involved

Monte9:
@binoy14 Let's consider adding this to v0.11.0

Monte9:
In an effort to streamline all RNE issues such that it is easier to get to solutions faster/ submit a new issue we have decided to close any enhancement/roadmap features as "issues". 

Feel free to use this issue to talk about this feature and ping @binoy14 or me (@monte9) if you feel like you can submit a PR for this. Thanks.

Also you can track the status of this feature and all other roadmap features [here](https://github.com/react-native-community/react-native-elements/issues/195#issuecomment-286213704).

sercanov:
Consider this too ; https://github.com/halilb/react-native-textinput-effects

iOSUser110:
am also looking for Floating Input within this library , or any workaround we can integrate with its Input field instead of other library use

RWOverdijk:
Is anyone working on this?

iRoachie:
@RWOverdijk not that i'm aware of

saxenanihal95:
@iRoachie may i work on this one thinking it to implement it like this: https://goshakkk.name/floating-label-input-rn-animated/

flyingcircle:
Closing due to inactivity.

HariAcidReign:
@flyingcircle Is this feature still a requirement for the library? And if yes, can I work on this issue?

Polad10:
+1

PR #3035, opened 2021-05-16
Added New Component: Floating Label

HariAcidReign:
Added a New Component (no external libraries used) #94 

Yes

Not yet. Once the component is approved, I will add the documentation

**Summary**
Satisfies the feature proposal requested in https://github.com/react-native-elements/react-native-elements-universe/issues/8
- Initially, I wanted to use some third-party library to create this component. However, there was a method to implement this without any such libraries, and hence I have made the PR here and not in `React Native Elements Universe`

No

**Other information**

https://user-images.githubusercontent.com/58134096/118386612-51b15f00-b636-11eb-9636-2713d6453049.mp4

Note: The floating label on user input goes down (instead of staying up) as I have not defined the `onChangeText` for the component. So, `props.value` will not change on text input and hence the component assumes that there is no input present, and brings the label down. However, in the user's implementation, the value will change, and hence the component will detect that `isFieldActive && !props.value` is `false` and `handleBlur` will not run that segment of code.


src/index.ts
```diff
@@ -43,6 +43,7 @@ import BottomSheet, { BottomSheetProps } from './bottomSheet/BottomSheet';
 import LinearProgress, { LinearProgressProps } from './linearProgress';
 import Switch, { SwitchProps } from './switch/switch';
 import Tab, { TabItemProps, TabProps } from './tab/tab';
+import FloatingInput, { FloatingInputProps } from './input/floatingInput';
```
arpitBhalla, line 46: Wouldn't it be better if add `variant='floating'` or `variant='outlined'` Prop in input? What do you think?

HariAcidReign: I wanted first to get the core styling of the component done and then build these new props in future iterations. that being said, that prop is indeed necessary so I'll look into it for sure.

khushal87 (commented):
Please adapt the PR according to the latest changes in the repo and resolve conflicts.

khushal87 (commented):
Please check the next branch for the file naming conventions and modify the files accordingly. You will find a common pattern in all the components and also a uniform directory structure. You need to modify your PR according to this. @HariAcidReign

src/index.ts
flyingcircle: ^ I agree with Arpit here. This seems like more of a variant rather than a completely separate new component.

PR #3052, opened 2021-05-26
Added new prop round to image

varunhhhrahul:

Added new prop round, which makes the image round based on its width

Yes.

**Summary**

**Other information**


flyingcircle:
This looks the same as `Avatar` which defaults to round and takes an image. But also this already a prop that can be passed in (borderRadius). So I don't think this simplifies anything to the end developer.

PR #3054, opened 2021-05-27
Added new prop round to button

akshitsuri:

Added new prop round to button

Yes

**Summary**

**Other information**


flyingcircle:
I don't see a strong need for this prop. I've never seen an app switch between round and not round for the same button. This is simply a static style which is a pretty common and easy to implement item.

Issue #3066, opened 2021-06-06
SocialIcon: Overall icon size doesn't change when prop raised={true}


rishabhsxn:
**Explain what you did (Required)**
I was trying to render the SocialIcon with an iconSize of less than default of 24.

**Expected behavior (Required)**
The overall size of icon should decrease.
```javascript
<SocialIcon type="facebook"  iconSize={16} />
```
should look like
![image](https://user-images.githubusercontent.com/48175786/120933128-496cb100-c716-11eb-9a4a-847ace52fbfd.png)

**Actual behavior**
But it only decreased the size of inner symbol:
![image](https://user-images.githubusercontent.com/48175786/120933186-9b153b80-c716-11eb-8bd7-8ed978130abf.png)

**Describe the bug (Required)**
* It shows same unexpected behaviour of outer wrapper unable to adapt to the passed iconSize prop for both: increasing or decreasing the iconSize from default of 24 pixels.
* This unexpected behaviour is not shown when we set raised={false}.
* This issue may be related to issue #3065 

**To Reproduce (Required)**
https://snack.expo.io/@rishabhsxn/socialicon-bug2

**Screenshots (Required)**
![image](https://user-images.githubusercontent.com/48175786/120933676-c8fb7f80-c718-11eb-8bfc-33deabb907ed.png)

**Your Environment (Required):**

| software              | version |
| --------------------- | ------- |
| react-native-elements |    3.4.1     |
| react-native          |      0.63.2   |
| expo           |      41.0.1  |

rishabhsxn:
When I removed !raised and !light from the conditional statement in : 
https://github.com/react-native-elements/react-native-elements/blob/08c0cb0edea6f4c2f12396a93c37cbf107c2d2fa/src/social/SocialIcon.tsx#L147

both the issue #3065 and #3066 were solved in my project. I am new to open-source contribution, I would love if anyone can guide me.
Please let me know if my suggested solution is correct or not. If yes, how could I submit it for PR?

khushal87:
@rishabhsxn How can I help you with raising a PR, where are you facing issues, can you be specific?

rishabhsxn:
I was confused with the different branches and test suites. I figured it out and made a PR.
Thank you @khushal87 for being open to help. I'll ping you if I get stuck. :)

souviknsl07:
[...]

Issue #3065, opened 2021-06-06
SocialIcon: Renders as button with 'raised={false}' and 'light={true}' props


rishabhsxn:
**Explain what you did (Required)**
I was trying to render the light themed 'facebook' SocialIcon without the elevation.
I checked on both android and iOS, facing the same issue.

**Expected behavior (Required)**
```javascript
<SocialIcon` type="facebook" raised={false} light />
```
should render
![image](https://user-images.githubusercontent.com/48175786/120930214-fdb40a80-c709-11eb-9ffe-7272a6fe50ae.png)

**Actual behavior**
But, it renders:  
![image](https://user-images.githubusercontent.com/48175786/120930431-d6117200-c70a-11eb-9a5b-f201f08647a8.png)

**Describe the bug (Required)**
* This behaviour is shown only when raised={false} and  light={true}, not in any other combination of these two props.
* Doesn't show expected behaviour even when you pass button={false} with raised={false} and light={true}.

**To Reproduce (Required)**
https://snack.expo.io/Bez6JulMd

**Screenshots (Required)**
![image](https://user-images.githubusercontent.com/48175786/120930989-46b98e00-c70d-11eb-9442-d4a594a2d255.png)

**Your Environment (Required):**

| software              | version |
| --------------------- | ------- |
| react-native-elements |    3.4.1     |
| react-native          |     0.63.2    |
| expo          |     41.0.1    |

arpitBhalla:
```javascript 
      <SocialIcon
        style={{
          width: 42,
          height: 42,
        }}
        type="facebook"
        raised={false}
        light={true}
      />
```
This can be reconsidered
https://github.com/react-native-elements/react-native-elements/blob/08c0cb0edea6f4c2f12396a93c37cbf107c2d2fa/src/social/SocialIcon.tsx#L147

PR #3073, opened 2021-06-09
Fix: Remove problem causing conditional checks in styling

rishabhsxn:
* Due to these conditional checks height, width and borderRadius are not calculated dynamically, instead are using hardcoded values
* Fix: SocialIcon rendered as button when props raised={false} and light={true}
* Fix: Whole icon not resizing according to passed prop iconSize when raised={true} (Only inner symbol resizes)

Bug fix for issue #3065 and #3066 

No. I am new to open-source contribution and not know much about testing.
However, the test suites need to be modified as the snapshots contain the old borderRadius value i.e. 30 (which was hardcoded). Now the borderRadius will be dynamically calculated depending upon the value of prop iconSize, so that the whole icon will resize based on iconSize.

Not relevant.

**Summary**
This PR implements bug fixes for above mentioned issues. 
Motivation: I was trying to resize the SocialIcon in my project but only inner symbol was resizing. So I checked the source code and tried to figure out what styles were being applied. I found out that the whole icon was properly resizing when I set raised={false}. In the process I also found that if I make raised={false} and light={true}, the component becomes a button type which should not happen if button={true} is not set.
I already opened the issues.

No.

**Other information**
Screenshot of test suite that needs to be modified to accomodate the dynamic calculation of borderRadius:
image


khushal87 (commented):
@rishabhsxn Please make sure all the checks pass. To update the snapshot you can run `yarn test -u` , this will automatically update the snapshots. Make sure after this your `yarn test` passes. Feel free to reach out for issues.

rishabhsxn:
I updated the test snapshot and ran ```yarn test```
Looks like all tests were passed.
image

Let me know if something is incorrect.

khushal87:
@rishabhsxn I think you need to solve the prettier issues:
```
/home/travis/build/react-native-elements/react-native-elements/src/social/SocialIcon.tsx
  147:11  error  Delete `··`  prettier/prettier
  148:1   error  Delete `··`  prettier/prettier
  149:11  error  Delete `··`  prettier/prettier
  150:1   error  Delete `··`  prettier/prettier
```

rishabhsxn:
@khushal87 Looks like this issue is related to line endings (LF and CRLF) or whitespaces. I am using VS code on a windows machine.
I can't generate this error on my local machine. Is there any other way to check if the error is solved without commiting and pushing the code?

khushal87:
@rishabhsxn Now there are no issues.

flyingcircle:
This solution doesn't make sense to me. If you add `button`, then the problem still appears with odd shaped social icons. Plus the `iconSize` prop would no longer be resizing the `icon` button the entire component, which doesn't seem correct to me.

rishabhsxn:
@flyingcircle  Could you elaborate " then the problem still appears with odd shaped social icons"  ?

Why would anyone want to increase or decrease the size of inner symbol of SocialIcon with iconSize and not changing the size of whole SocialIcon? A user would consider SocialIcon as an icon not as a component containing an icon.

PR #3081, opened 2021-06-13
[Refactor] Directory structure of components

khushal87:
This PR refactors the directory structure of the components as in the list below. This includes both simple and compound components. There are other refactoring including importing the Rating components(TapRating, SwipeRating), etc. All the directories and files are using a common pattern including usage of props and the way it's imported and used.

Yes, I have done the required changes in the tests if needed and have maintained the imports for the components as and when required.

Not required for this change.

**Summary**
I have changed the directory structure considering the point in mind that it will unlock new possibilities when it comes to working with the code. The directory structure along with the components files is now uniform. An example of the pattern can be found here:
```
Suppose X is a component:
Directory structure:
X
 - X.tsx
 - index.tsx
 - __tests__
```
Now, if there's a child component that is used by some other component then I have created a `components` sub-directory for it. I have done this considering the point that **`src` directory will only be exposed to the users and developers using index.ts and specific UI component's `index.tsx`**. This change is done for the following components as they had child components:
- Tile
- Header
- Checkbox
- PricingCard
- SpeedDial
- ListItem
- Slider
- Tooltip
An example of this pattern of modification is:
```
Suppose X is a component:
Directory structure:
X
 - components
       - Y.tsx
 - X.tsx
 - index.tsx
 - __tests__
```

Moreover, I have tried my best to make the components code linear within the components files. This includes export, import, usage structure.

Yes, this makes the directory structure uniform such that new developers can find it easy to work on the project and this also allows adding ideas like Docs Autogen easier as this will help a common parser/code to parse the components and extract information out of it.

**Other information**
Components whose directory structure are refactored:
##### Simple Components
The simple components follow the architecture as mentioned above in the example. 

- Badge
- BottomSheet
- Checkbox
- Divider
- Header
- Icon
- LinearProgess
- Overlay
- PricingCard
- Searchbar
- Slider
- SocialIcon
- Switch
- Text
- Tile

##### Compound Components
They use Base components which are exported along with the other components in the `index.tsx` file. The Themed export is also done in the `index.tsx` file itself.
- Avatar
- Card
- Dialog
- ListItem
- Tab
- TabView

##### Class based Components
Some of the components are still using the traditional class-based components. I have done the folder restructuring including adding `index.tsx` file and doing the import-export work for the components. These include:

- Image
- Input
- SearchBar
- Tooltip

##### Buttons

There is a folder restructure for Button components. Previously components like Chip, ButtonGroup, SpeedDial, etc. whereinto the same directory as `buttons`, and now all the components have their own directory. This solution will unlock structured and readable code including other benefits. Components for which the following change is done:

- Button
- ButtonGroup
- Chip
- FAB
- SpeedDial

##### Ratings

The package currently directly uses Rating components such as TapRating, SwipeRating directly from the package `react-native-ratings`. To generate documentation automatically we need to provide the components as a separate entity in the code. Therefore, I have created two different folders with the same names as users would import them:

- AirbnbRating as TapRating
- Rating as SwipeRating The name of the folder is as per the usage.

**Tags**


pranshuchittora:
Review Checklist
- [ ] TSC Build
- [ ] Tests
- [ ] Website working & building

Issue #3004, opened 2021-05-01
[RFC] Visual Preview of components in VS Code


arpitBhalla:
![Screenshot from 2021-05-01 16-46-53 (1)](https://user-images.githubusercontent.com/55053424/116780993-9ddc9b00-aa9d-11eb-97a0-5f6513d85c5d.jpg)

Can be achieved by adding this above the component code
```javascript
/**
 * Documentation: https://reactnativeelements.com/docs/searchbar
 *
 * Preview: ![Search Box](https://user-images.githubusercontent.com/17592779/31586716-f6e8ff9c-b1d4-11e7-918f-2a7e11d51b08.gif)
 *
 * Demo Code: ```
   <SearchBar
        placeholder="Type Here..."
        onChangeText={this.updateSearch}
        value={search}
      />
 * ```
 */
```

@pranshuchittora @flyingcircle @dhruvdutt

hetanthakkar1:
it's great

arpitBhalla:
Solved by https://marketplace.visualstudio.com/items?itemName=rne.snippets

PR #3112, opened 2021-06-28
[Add] props description, descriptions to the component.

khushal87:
This PR introduces adding descriptions and props descriptions to the components for auto-generation of docs in the future.

Yes, I have added and configured the tests accordingly.

Not required.

**Summary**
This PR added descriptions and props description to the components except for the class-based components:
- Image
- Input
- Searchbar
- Tooltip

No.

**Other information**
Nothing as such. I have used a common pattern for commenting on the descriptions in the components.

**Tags**


arpitBhalla:
@khushal87 Could you add this feature too? https://github.com/react-native-elements/react-native-elements/pull/3105#discussion_r657893869

khushal87:
@arpitBhalla This is not decided yet, I will have to talk to mentors. Adding all this will unnecessarily make the component long with descriptions in my opinion.

arpitBhalla:
@khushal87 Its already decided in #3004

khushal87:
I can't see any mentors/admin feedback on it.

arpitBhalla:
definitely adding single line for preview would not make description lengthy

src/AirbnbRating/AirbnbRating.tsx
```diff
@@ -7,6 +7,11 @@ import { RneFunctionComponent } from '../helpers';
 
 export type TapRatingProps = RatingProps;
 
+/**Ratings are used to collect measurable feedback from users.
```
dhruvdutt, line 10:
```suggestion
/** Ratings are used to collect measurable feedback from users.
```

src/Avatar/Accessory.tsx
```diff
@@ -14,10 +14,15 @@ import { RneFunctionComponent } from '../helpers';
 
 export type AccessoryProps = Partial<IconProps> &
   Partial<ImageProps> & {
+    /**Add underlay color to the accessory of avatar. */
```
dhruvdutt, line 17:
```suggestion
    /** Add underlay color to the accessory of avatar. */
```

dhruvdutt: @khushal87 Can you make this change everywhere? You can find and replace all.

khushal87: @dhruvdutt Done, thanks for pointing it out. 🙌

flyingcircle (approved):

arpitBhalla:
@dhruvdutt Can we add preview images and documentation links to description, it would be easy for developer to quickly access docs.

khushal87:
@arpitBhalla The problem with this is that it will also be visible in documentation after parse unless and until we remove it by adding additional conditions. I don't think making the code complex is a good thing, where we already have an underlying props filter complexity which we would be having. Docs are enough to show what the components are used for. This is my opinion. @dhruvdutt what do you say?

arpitBhalla:
> I don't think making the code complex

removing 2 lines would not be complex, you can do it after generating markdown files, I guess 🙄

dhruvdutt (approved):

dhruvdutt:
@arpitBhalla Adding preview image & description link can be picked up separately as an enhancement later. Let's track it at #3115 and add it for all components together since this would need some changes in the comments parsing for filtering during auto-generation of docs.

PR #3136, opened 2021-07-09
[Add] Docgen auto generation of docs scripts

khushal87:

This PR introduces adding scripts to generate docs automatically. It uses `react-docgen-typescript` to achieve the same.

No. 	`React docgen Typescript`, for now, has its own tests and we can rely on them as they are stable. We can add tests to our code in the future. I plan to add them if needed in the future with a single PR.

Will be done later with a single PR.

**Summary**

To run the scripts for now and generate the documentation run the command:
```
yarn build:docs
```

This will trigger `scripts/docgen/index.tsx` script. It calls the function `getAndStoreComponentDocs` which calls `getComponentFiles` which is a utility to get all the file paths of the components in the form of an array. This is served as input to the `generateComponentDocs` which generate the component docs using `react-docgen-typescript`. The result is stored inside the `.docgen` directory in the respective folders.
The scripts have the respective files:

- **componentMap.ts** - This is a mapping of all the components including compound and non-compound components and is served as an input to get the array of all files paths. This will be removed later once we add regex or automate the process.
- **docgenParser.ts** - This file adds configurations to the react-docgen-typescript and according to the configurations, the props are filtered.
- **generateComponentDocs.ts** - This calls the docgen parser passing the array of files paths as input. This also sorts the props in sorted order of the key. Finally, a global object of all the component docs is returned.
- **getComponentFiles.ts** - Converts the mapped object of files in `componentMap.ts` to a single array of files having all the component file paths. 

No.

**Other information**
*Disclaimer*: These are the initial scripts to generate JSON just to make sure that the data is received for every prop type and descriptions are fetched. There will be changes in future PRs once the Markdown is generated.

**Tags**


src/Card/CardDivider.tsx
```diff
@@ -20,4 +20,4 @@ const styles = StyleSheet.create({
   },
 });
 
-CardDivider.displayName = 'CardDivider';
+CardDivider.displayName = 'Card.Divider';
```
pranshuchittora, line 23: Why not `CardDivider`

khushal87: The name should be the same as the use-case, therefore I have changed it. As users are going to use it as Card.Divider. Also, this will help detect the parent directory.

Issue #3130, opened 2021-07-05
Slider allowTouchTrack prop is not working


hiji:
**Explain what you did (Required)**

I set the Slider allowTouchTrack prop to false, but it doesn't get disabled.

**Expected behavior (Required)**

Slider allowTouchTrack prop setting is enabled.

**Describe the bug (Required)**

Documentation explains that Slider's `allowTouchTrack` prop default is `false`, but the behavior is the same as when `true` is set.
Also, setting allowTouchTrack prop to false is ignored.

**To Reproduce (Required)**

**Screenshots (Required)**

**Your Environment (Required):**

| software              | version |
| --------------------- | ------- |
| react-native-elements |     3.4.1    |
| react-native          |     0.63.4    |

PR #3137, opened 2021-07-10
allowTouchTrack prop is not working issue (#3130) fixed

saif-malik-01:
Open issue  - #3130 Fixed

There is a bug in slider component props called allowTouchTrack, according to 
docs, allowTouchTrack props default value is false but it behave like that the value is 
true , even if you set the value to true it just make no change to behaviour.

I simply solve this bug by just adding one more condition in ' if ' condition.

After applying changes , i did some test of my own and it work fine.

### DIFF
--- a/src/Slider/Slider.tsx
+++ b/src/Slider/Slider.tsx
@@ -306,7 +306,7 @@ export const Slider: RneFunctionComponent<SliderProps> = ({
     if (!allowTouchTrack && !TRACK_STYLE) {
       return thumbHitTest(e);
     }
-    if (!trackStyle) {
+    if (!trackStyle && allowTouchTrack) {
       setCurrentValue(getOnTouchValue(e));
     }
     fireChangeEvent('onValueChange');


flyingcircle:
Nope this didn't work. `allowTouchTrack={true}` wasn't turning on touch track capability. Tested on iOS

saif-malik-01:
https://user-images.githubusercontent.com/76686929/126952058-98472863-29be-4cc3-964a-2c567574c6b0.mp4

I don't know why but after changing the allowTouchTrack prop value I have to explicitly refresh the the app (expo) to make it work fine. May be because of useRef handleStartShouldSetPanResponder not get refresh after changing the Value for the same . Can anyone explain a bit about it. Thank you.

src/Slider/Slider.tsx
```diff
@@ -306,7 +306,7 @@ export const Slider: RneFunctionComponent<SliderProps> = ({
     if (!allowTouchTrack && !TRACK_STYLE) {
       return thumbHitTest(e);
     }
-    if (!trackStyle) {
+    if (!trackStyle && allowTouchTrack) {
```
harshal-moneylion, line 309: Shouldn't this be a ||

@saif-malik-01 
@flyingcircle Does that work if you try with ||??

harshal-moneylion:
This PR should also fix https://github.com/react-native-elements/react-native-elements/issues/3044

src/Slider/Slider.tsx
saif-malik-01: Nope , it will not. It actually make no sense  to the statement as no matter what the value of prop is, this statement is going to be always evaluate as true .

flyingcircle: This doesn't logically change anything. Combined with the if statement above, adding this doesn't do anything.

PR #3140, opened 2021-07-14
Added support for screen orientation

katatumuri-maimai:

 a bugfix

**Summary**
Previously, the modal component specification changed the orientation of the phone screen when the tooltip was opened.
Screen orientation doesn't change with added support for screen orientation.

**Other information**
I will attach an image.
I'm not good at English, so maybe I'm doing something wrong.
If so, I'm sorry.

## Before this change
Tooltip-before
## After this change
Tooltip-after


flyingcircle:
`patch` is not the correct location for a PR. Check the documentation on how to contribute.

PR #3187, opened 2021-08-28
Migrate to Pressable

arpitBhalla:
**What kind of change does this PR introduce?**
Convert Touchable(Highlight|Opacity) based components to Pressable.  

- [x] Avatar
- [x] Accessary
- [x] Badge
- [x] ButtonGroup
- [x] SpeedDial
- [x] Checkbox
- [x] Icon
- [x] Image
- [x] SearchBar
- [x] Social Icon
- [x] Overlay
- [x] FeaturedTile
- [x] Tile
- [x] ListItem

**Did you add tests for your changes?**
Yes, I have updated the tests and migrated them to RN-testing Library.

**If relevant, did you update the documentation?**
No changes are required.

**Does this PR introduce a breaking change?**
No, this would not break anything.

**Other information**
GSoC Task of @arpitBhalla.
- [x] Self Review for any kind of error in code.

#### I have tested the PR manually in demo app / Snack
- [x] Android
- [ ] iOS
- [x] Web 

Closes #2939

PR #3145, opened 2021-07-16
[Fix, Improve] Doc-gen issues and automation

khushal87:
This PR fixes existing issues and covers the document generation tasks which include the following:

**Issues:**
- Fixing the typescript compilation issue which was due to the presence of scripts in the root directory and having a different target to which it was compiled into.
- Fixing the issues related to preview of markdown on the docusaurus pages.
- Improving the performance of the script by improving the file input strategy to the docgen parser.
- Adding relevant Comments and Descriptions wherever needed.
- Made `SpeedDial` structured to parse it as a Compound Component.

**Features**
- Adding relevant scripts in `package.json` to run the docgen scripts.
- Upgrading the version of the docusaurus package and related presets and plugins to `2.0.0-beta.4`.
- Improve the Generate Markdown script to accommodate the markdown generation for Compound Components.
- For compound components, the children's would be stored within the parent of the Component within the `childrens` array.
Eg: ListItem.Accordion
The JSON for the Accordion will be inside the ListItem `childrens` key.
- Fixing the broken links within the markdown which were not working the Docusaurus preview.
- Added relevant Prop Filters in the `docgenParser.ts` to fix issues with the broken markdown. These issues were due to the presence of improper JSX in markdown, irrelevant props which would be difficult to understand for users.
- Added pre-push hook to automate the process.
- Added Google Summer of Code - khushal87 blog to the documentation website.

**Tags**
#gsoc #docs-autogen #khushal87

PR #3193, opened 2021-09-02
(website): Update script for InlinePressableProps

khushal87:
This PR fixes the issue with the `InlinePressableProps` which is under the `src/helpers` directory. This makes the pressableProps to be rendered with the documentation.


arpitBhalla:
Screenshot-20210830160902-659x431

khushal87:
> Multiple _prop description_
> Screenshot-20210830160902-659x431

@arpitBhalla I suppose I comment on the PR #3187 to fix them in docgenParser.ts.  I don't know the best possible types for these props as you have worked on them. So please add conditions for them in your PR or create a new PR if that is merged to fix them. Adding conditions for them is the best possible way for now to fix them. And, I am not aware of the exact type if this is replaced. So, I asked you to work on them in your PR. Please look at it. The docgen is working fine as it is supposed to be.

khushal87:
> Multiple _prop description_
> Screenshot-20210830160902-659x431

This is because these props are already present in pressable props provided by React Native. The description inside node_modules and the added description by you are merged. Please remove your added prop to make this work. As the props are overlapping with the same name and do not make any sense.

arpitBhalla:
> Please remove your added prop to make this work

?

khushal87:
Screenshot from 2021-09-03 11-20-40
@arpitBhalla The history shows that you have added the description here. What I mean is that, if react docgen typescript detects two same props(intersecting props).

Here:
One inside `PressableProps` inside /react-native-elements/node_modules/@types/react-native/index.d.ts:
```
export type InlinePressableProps = Inline<
  PressableProps,
  'onPress' | 'onLongPress' | 'onPressIn' | 'onPressOut'
>;
```
having description `Called when a long-tap gesture is detected.`
and the other in lines 25-26 of Image.tsx
```
/** Callback function when long pressing component. */
    onLongPress?(): void;
```
having description `Callback function when long-pressing component.`.

As there are the same props with two descriptions react-docgen-typescript is merging their descriptions. 
The solution is to remove the `onLongPress` from the component props as it is already in `PressableProps` you have imported.

arpitBhalla:
> @arpitBhalla The history shows that you have added the description here.

Kindly check again and see carefully in #3145 

https://github.com/react-native-elements/react-native-elements/pull/3145/files#diff-9d5b5c3f693360668dda73e10258f9cf27ca4788fd30821a0ba70aadfe4d46f3R21-R25

khushal87:
> > @arpitBhalla The history shows that you have added the description here.
> > why do we need it again in component props?
> 
> Kindly check again in https://github.com/react-native-elements/react-native-elements/pull/3145/files#diff-9d5b5c3f693360668dda73e10258f9cf27ca4788fd30821a0ba70aadfe4d46f3R21-R25
> 
> Can we make something to overwrite the parent description?

I will have to think about that. Do you have a solution?

arpitBhalla:
No ;)

khushal87:
> No ;)

Can we remove the prop from Image.tsx as it is already present?

arpitBhalla:
@khushal87 Can you merge `next` ?

PR #3195, opened 2021-09-02
Update input.md

SteveGreenley:
According to the RN documentation, when setting an input field's value with setNativeProps, the property to set is 'text' not 'value'. Setting value rather than text does nothing.

**Summary**

**Other information**


khushal87 (commented):
Please don't update the docs by updating the markdown files at `versioned_docs`. They won't be reflected on the next version/release. The right way of doing it is by updating the files inside docs. Please go through https://reactnativeelements.com/blog/2021/08/12/auto-generation-of-docs

PR #3198, opened 2021-09-10
fix: change TextProps from type to interface to enable declaration me…

wkoutre:
…rging

Please see my comment [here](https://github.com/react-native-elements/react-native-elements/issues/2310#issuecomment-916561196) and the [follow up](https://github.com/react-native-elements/react-native-elements/issues/2310#issuecomment-916908717).

This is a Typescript-related bug fix.

No.

Yes!

**Summary**

Please refer to links above.

Not as far as I can tell in my review of where `TextProps` is used.


wkoutre:
Hi, any word on this?

mdjarv:
I ran into this issue when trying to follow the documentation to augment `TextProp` with a few more styles.

Until this is merged the documentation below is incorrect
https://reactnativeelements.com/docs/customization#typescript-definitions-extending-the-default-theme

arpitBhalla:
Could you please resolve conflict.

wkoutre:
@arpitBhalla All set.

arpitBhalla:
Don't we need to change types to interface for all components?

wkoutre:
Doing so would enable declaration merging for everything, yes. Would you like me to do so in this PR? I could otherwise submit another with these broader changes, whichever is preferred.

arpitBhalla:
For sure you can do it in another PR

PR #3203, opened 2021-09-16
fix(Slider): Fix issues with scrub and allowTouchTrack

### LINKED ISSUE #3087: Slider props value changes not rendering correctly
Hello,

I'm having problems with Slider every time props value changes. I have already update react-native-elements to last version (3.4.2) to resolve  this bug [#2995](https://github.com/react-native-elements/react-native-elements/issues/2995) but I'm still having problems. Now every time props value changes, the position of the thumb changes correctly but it seems props value is not changing. Here it is a video that explains it better.

https://user-images.githubusercontent.com/51318357/122426276-657f1680-cf90-11eb-85ab-5d3d4a5a5933.mp4

To control Slider value, there is a parent component who stores in state the value and listens to every element that change this. Here it is the parent code:

```import React, { useState } from 'react';
import { View, StyleSheet } from 'react-native';
import { Button, Slider } from 'react-native-elements';

const LightDimmerDetailScreen = () => {
  const [currentValue, setCurrentValue] = useState(0);

  const handleOnSliderChange = (newValue: number) => {
    setCurrentValue(newValue);
  };

  return (
    <View style={styles.screen}>
      <Slider
        value={currentValue}
        minimumValue={0}
        maximumValue={100}
        step={1}
        minimumTrackTintColor={'red'}
        onSlidingStart={(newSliderValue) => console.log('on sliding start: ', newSliderValue)}
        onSlidingComplete={(newSliderValue) => handleOnSliderChange(newSliderValue)}
      />
      <Button onPress={() => setCurrentValue(50)} title={'set to 50%'} />
    </View>
  );
};

export default LightDimmerDetailScreen;

  screen: {
    marginTop: 100,
    width: 300,
    marginHorizontal: 57,
  },
});
```

I'm using:
- react-native-elements: 3.4.2
- react-native: 0.64.1

Thanks in advanced for your help!

NiwotSmitty:
Fixes #3087 and #3130

Yes, existing test cases were fixed, snapshots updated, and new test cases added.

**Summary**
This PR is a significant refactor of the Slider component.  The primary driver of this change is to fix the `onValueChange`, `onSlidingStart`, and `onSlidingComplete` callback props, as they weren't sending in the current position.  I believe they were broken when the Slider component was ported to a function component.

Before starting this update, I also realized the test scripts were failing on the Slider component.  So, this PR also fixes the tests, updates the snapshots, and adds a couple new test.  Unfortunately, the test coverage is lower than I'd like, but the primary reason is the Pan Gesture handlers, as they're commonly cumbersome and tedious to test with Jest.

Additionally, major kudos to @menssen for proving a great [outline](https://github.com/react-native-elements/react-native-elements/issues/3087#issuecomment-883609255) of what needed to be done, and providing a patch.  As @menssen pointed out in his comments, his changes had been de-typescripted, so I use his changes as a starting point, added-in typescript where appropriate, and fixed/optimized other items as I came across them..

Here are some notes about the changes (again, thanks to @menssen for a fantastic starting point):
* Added `useCallback()` to functions within the component, and added the appropriate dependencies array. Note this change required reordering of many of these functions to eliminate compiler complaints.  Unfortunately, this makes the diff a little harder to follow.
* Moved function `getBoundedValue()`, `handlePanResponderRequestEnd()`, and `handleMoveShouldSetPanResponder()` out of the component, as they're not dependent on any component state variables.
* Added `SizableVars` enum to describe possible vars managed within `handleMeasure()`
* Extracted setting `allMeasured` into a stand-alone `useEffect()` so it gets automatically updated whenever any of the `SizableVars` structures change.
* Destructured all props used within the FC so they're not passed-down unnecessarily (otherwise they get passed into the container View).
* Removed `didMountRef` as it really wasn't needed.
* Remove numerous cases where an optional prop was used as `someProp || defaultValue`, as that's not needed because in those cases the `defaultProps` ensure default values exist at initial render.
* Always range-bound `props.value` at the top of the function.  This eliminated the need to bound it later (as was done multiple times).
* Changed `isVertical` from ref to calculated value, as it will now update if props are changed.
* Review @menssen comments re his [changes](https://github.com/react-native-elements/react-native-elements/issues/3087#issuecomment-883609255); those are mostly included in this PR.

I think that captures the gist of the changes.  I realize this is a large PR, so thanks in advance for review.  IMO, this is a pretty important PR, as the existing implementation doesn't work in the 3.x stream.  In my app, I had to remain on the 2.3 stream until. I fixed these issues.

Please let me know if there are any questions or issues.


freemanUA:
is this alive?

NiwotSmitty:
I've never received any feedback from the maintainers re this PR.  I'm happy to update as needed, but I have no feedback.

NiwotSmitty:
Done.  I simple merged the latest 'next' branch.  There were no conflicts.  Thanks.

arpitBhalla:
The changes looks good to me, Thanks for great work @NiwotSmitty ;)

Ashutosh-Tiwari:
@arpitBhalla In which version of react-native elements will this fix be available? I am using 3.4.2 currently and I still don't see it fixed in my case.

KrisLau:
Omg I just saw this I thought it was my fault. I kept adding random things like dynamic keys to try and force a re-render. Thanks so much for this.

PR #3207, opened 2021-09-23
Fix npm and GitHub spelling

friederbluemle:
Just a couple of capitalization fixes: `Github` -> `GitHub` and `NPM` -> `npm`.

npm should never be capitalized:
https://github.com/npm/cli#is-it-npm-or-npm-or-npm


khushal87 (commented):
The PR doesn't make much sense. Though thanks for the efforts :)

friederbluemle:
> The PR doesn't make much sense. Though thanks for the efforts :)

Why does fixing typos not make much sense? ;)

PR #3212, opened 2021-10-01
Fix typo in Divider

UdithaIshan:
In the **Usage** section, every **Divider** component ends with a semicolon. And it differs from the other components usage sections.
This PR fixes that and keeps the consistency.


khushal87 (changes requested):
The `versioned_docs` is not the right place to update documentation. That doesn't get reflected with the `next` releases. Please check the `next` branch and update the documentation in the `docs/main/usage` folder if there's an issue as such and if you find the same there.

Issue #3196, opened 2021-09-04
Add accessibilityHint and accessibilityLabel support to ButtonGroup buttons


mitchdowney:
**Is your feature request related to a problem? Please Describe.**

Add accessibilityHint and accessibilityLabel support to ButtonGroup buttons. Currently when I try to pass in a custom button component like this to the ButtonGroup buttons prop:

```
const buttons = [
  (
    <Text
      accessibilityHint={translate('ARIA HINT - Tap to search by podcast title')}
      accessibilityLabel={translate('Search by podcast title')}>
      {translate('Podcast')}
    </Text>
  ),
  (
    <Text
      accessibilityHint={translate('ARIA HINT - Tap to search by podcast host')}
      accessibilityLabel={translate('Search by podcast title')}>
      {translate('Host')}
    </Text>
  )
]
```

Neither the hint or label are read. I think it may be because the ButtonGroup buttons are wrapped in their own <Touchable> that we cannot override or configure with props.

**Describe the solution you'd like**

When accessibilityHint and accessibilityLabel are provided for custom ButtonGroup buttons, they should be read by iOS and Android screen readers. This solution may include overriding the <Touchable> wrapping the ButtonGroup buttons.

**Describe alternatives you've considered**

I can't think of another way to accomplish the same thing.

ruiaraujolink:
I'm facing the same problem.

ruiaraujolink:
@mitchdowney 
My workaround is this:

```javascript
buttons={[
  <View accessibilityLabel='Ver gráfico' collapsable key='chart'>
    <FontAwesomeIcon
      color={selectedTabIndex === 0 ? theme.colors?.highlight : theme.colors?.primary}
      icon={faChartLine}
      size={20}
    />
  </View>,
  <View accessibilityLabel='Ver tabela' collapsable key='table'>
    <FontAwesomeIcon
      color={selectedTabIndex === 1 ? theme.colors?.highlight : theme.colors?.primary}
      icon={faTable}
      size={20}
    />
  </View>,
]}
```

Make a View collapsable enable me to put the accessibilityLabel and this approach allow me to pass the accessibility Scanner test on Android.

mitchdowney:
@ruiaraujolink oh interesting, thanks for sharing.

Any idea why `collapsable` is needed here for Android only? How the heck did you figure that out? lol

Also, is it OK to keep this property on iOS too? It won't cause any conflicts?

ruiaraujolink:
[...]

PR #3220, opened 2021-10-10
fix(ButtonGroup): add and improve types and add tests

grgr-dkrk:
Improve Types(ButtonGroup)

yes.

it is already exists.

**Summary**

- The documentation says that ButtonGroup needs to specify `{element: componentName}` for `buttons` props, but TypeScript doesn't have that type, thus I added a type.
- Several types in ButtonGroups have been improved.
- Added a test case when `buttons` is entered in an array other than a string.

no.

**Other information**
I was thinking the #3196 problem could be solved with `{element: componentName}`.
However, improving #3196 required modification of the parent component. This PR does not directly resolve it.


arpitBhalla:
Can you merge `next` ?

PR #3229, opened 2021-10-22
[Icon] Fixed ViewStyle  Color type

dbrab:

Improve Types(IconType)

yes

**Summary**

- The react-native View Style documentation says that color must be of type string, even in the element documentation it says the same, but currently it is of type Number | Color, this gives conflict in typescript
[https://reactnativeelements.com/docs/icon/#color](url)

No

**Other information**

### DIFF
--- a/src/Icon/Icon.tsx
+++ b/src/Icon/Icon.tsx
@@ -1,25 +1,25 @@
+import Color from 'color';
 import React from 'react';
 import {
   Platform,
-  View,
-  StyleSheet,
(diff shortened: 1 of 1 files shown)


khushal87 (commented):
What I can see here is `string` only. Where did you find `number | string` as the type? Which version are you looking at?

flyingcircle:
I'm with @khushal87 I see the type error, but this doesn't seem to solve the issue.

PR #3233, opened 2021-10-24
Update link to React Native TextInput element

shystruk:
Update link to React Native TextInput element

**Summary**

**Other information**


khushal87 (approved):

khushal87 (changes requested):
The only change I would ask here is that please don't update docs inside `versioned_docs` but inside the respective docs folder so that it gets reflected in the next branch and consequent releases. Thanks 😄

PR #3250, opened 2021-11-17
Add noopener and noreferrer to links opening a new tab

Goff-Davis:

Minor bugfix

No, this change should not require additional or updated tests.

N/A

**Summary**

Links that contain `target="_blank"` as a best practice should also include the `rel` property with `noopener` and `noreferrer` to prevent tabnabbing. Since these are links to pages that are owned by the maintainers of this repo this change isn't super important, however.

N/A

**Other information**

N/A

### DIFF
--- a/website/src/components/OpenCollective.tsx
+++ b/website/src/components/OpenCollective.tsx
@@ -26,6 +26,7 @@ const OpenCollective: React.FC<{}> = () => (
           <a
             href="https://opencollective.com/react-native-elements#backers"
             target="_blank"
+            rel="noopener noreferrer"
           >
             <img src="https://opencollective.com/react-native-elements/backers.svg?width=695" />
           </a>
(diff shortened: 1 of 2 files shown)


website/src/components/OpenCollective.tsx
```diff
@@ -26,6 +26,7 @@ const OpenCollective: React.FC<{}> = () => (
           <a
             href="https://opencollective.com/react-native-elements#backers"
             target="_blank"
+            rel="noopener noreferrer"
           >
             <img src="https://opencollective.com/react-native-elements/backers.svg?width=695" />
```
LucasBerce71, line 31: how to alt into image?

LucasBerce71 (changes requested):

website/src/components/OpenCollective.tsx
Goff-Davis: I'm not sure what you're asking. Are you asking if the image on line 30 should have an `alt` attribute?

That's not really in the scope of the change I was making, but I'm happy to add it if there's some +1s on the request. Omitting the `alt` attribute indicates that the image is a key part of the content and there isn't a textual equivalent. I don't have enough context on this site to decide that.

PR #3263, opened 2021-12-16
Release 4.0.0-beta.0

flyingcircle:
**What kind of change does this PR introduce?**

**Did you add tests for your changes?**

**If relevant, did you update the documentation?**

**Summary**

**Does this PR introduce a breaking change?**

**Other information**

Issue #3184, opened 2021-08-26
Minimize props filter for documentation auto-gen


khushal87:
As the project uses Docusaurus as the documentation website, the format used in docusaurus to render docs is Markdown. Now, with the latest work, automatic generation of documentation is processed. There are some props that don't render well in markdown and break on compilation while running the documentation website.

To deal with this, I have added the props filter in the docgenParser in the file `website/scripts/docgenParser.ts`. This was the solution that was taken forward after considering other possibilities,

This issue is created in continuation to the fact that we can make the code more maintainable in the future by minimizing those props filters. This should be done taking in mind the fact that the docgen parser works well and the rendering of the generated markdown is not hampered.

Looking for contributions:
I am also creating a task list for the same:

- [ ] To replace all the '|' in props with 'or'. _(Done for the purpose of readability)._
- [ ] To replace all the '&' in props with 'and'. _(Done for the purpose of readability)._
- [ ] To deal with the props of type StyleProp<ViewStyle> and StyleProp<TextStyle> which breaks the markdown. _(This was done as the markdown was breaking due to the presence of angular braces)._
- [ ] To deal with the props of type `() => void` or `() => any`.
- [ ] To deal with prop's default value of type `() => null` or `() => {}`.
- [ ] To deal with the props of type Partial<> which breaks the markdown.
- [ ] Replace the occurrences of type Component/ViewComponent to React Component.
- [ ] To replace the containerStyle props with 'typeof styles'(Object) to Style Object.
- [ ] To replace the animation prop with a valid type of ListItem.Accordion.
- [ ] To deal with the platform-specific props default value in ButtonGroup.
- [ ] To replace '\r\n' which breaks the markdown for certain props to ''.
- [ ] To deal with the Badge Component with prop name onPress.
- [ ] To deal with the platform-specific props default value in Icon.
- [ ] To deal with the ViewComponent prop default value in Header.
- [ ] Replace all the props with default value theme?.colors?.primary to Color(Primary).
- [ ] To deal with the prop of default value onPress || onLongPress ? TouchableOpacity: View in Avatar.

arpitBhalla:
Hopefully we can do it using TSdoc tags

PR #3266, opened 2021-12-18
Fix issues with introduction of Pressable props in docs autogen

khushal87:
This PR fixed the pressable props issue in the autogeneration of docs and also fixes the multiple exports from the LinearProgress component such that the `yarn docs:builds` runs successfully and the docs are updated automatically.

Reference - #3263

### DIFF
--- a/src/Avatar/Avatar.tsx
+++ b/src/Avatar/Avatar.tsx
@@ -115,11 +115,9 @@ export const Avatar: RneFunctionComponent<AvatarProps> = ({
 }) => {
   const width =
     typeof size === 'number' ? size : avatarSizes[size] || avatarSizes.small;
-  const ComponentProp = Component
-    ? Component
-    : onPress || onLongPress || onPressIn || onPressOut
-    ? Pressable
-    : View;
+  const ContainerComponent =
+    Component ||
+    (onPress || onLongPress || onPressIn || onPressOut ? Pressable : View);
 
   const height = width;
   const titleSize = width / 2;
@@ -162,7 +160,7 @@ export const Avatar: RneFunctionComponent<AvatarProps> = ({
   }
 
   return (
-    <ComponentProp
+    <ContainerComponent
       style={StyleSheet.flatten([
         styles.container,
         { height, width },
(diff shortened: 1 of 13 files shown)


khushal87:
@flyingcircle Awaiting your feedback!

flyingcircle:
Is moving the default parameter to a function variable a requirement for this fix?

khushal87:
> 

We would need to add some checks in that case to handle the parsing of such a prop. I will have to think about how can we parse it using the condition.

I chose this approach because ideally, I have noticed default props to be simpler everywhere and not with a complex condition(which appears like a code smell). Anyways if we have a `Component` prop that the user provides then it can be handled here. On the other side, this checked the presence of other props(onPress, onLongPress, etc.) and takes the default value, that's the reason I extracted the logic out of default logic. Ideally, we don't have any default component if the props(onPress, onLongPress, etc.) aren't present.

flyingcircle:
I agree that it's a code smell, but the code has also been like this since I've worked on it. I think it's worth solving, but I think it needs its own PR.

flyingcircle:
If there's no other way to solve it easily, then I'll accept this PR. Let me know.

khushal87:
Okay, I will have a look at alternate approaches to the problem and revert soon.

khushal87:
To solve it the correct way I would say it would need internal code modification. If we can achieve this in a single different PR we also solve #3184

src/Avatar/Avatar.tsx
```diff
@@ -117,6 +115,9 @@ export const Avatar: RneFunctionComponent<AvatarProps> = ({
 }) => {
   const width =
     typeof size === 'number' ? size : avatarSizes[size] || avatarSizes.small;
+  const ContainerComponent =
```
arpitBhalla, line 118: I don't think this is right way of doing it, you just moved default prop within the component which *may* cause performance issues, and still this is unreadable(as per you https://github.com/react-native-elements/react-native-elements/pull/3263#issuecomment-997229483)

It has nothing to do with my Pressable PR, as title of this PR say. You need to take care in autoDoc Parser.

I guess we can use `@defaultValue` property of TSDoc.

@khushal87 @flyingcircle

PR #3279, opened 2021-12-30
Refactor base and themed package

arpitBhalla:
- Docs build script update
- Separate out base and themed packs


arpitBhalla:
@flyingcircle Please review

PR #3281, opened 2021-12-31
chore: Added props folder with data abstracted from root mdx files.

jugshaurya:
- updated MDX files to include respective Props component
- build-related change of playground repository.

**Summary**
- Issue and solution discussed here:- https://github.com/react-native-elements/playground/issues/121#issuecomment-1002652840

**Other information**


arpitBhalla (changes requested):
We don't have to change mdx files, these are auto generated by script, make changes in that script. I have raised https://github.com/react-native-elements/playground/pull/122, it will solve the problem by removing content above `## Props`

jugshaurya:
> We don't have to change mdx files, these are auto generated by script, make changes in that script. I have raised [react-native-elements/playground#122](https://github.com/react-native-elements/playground/pull/122), it will solve the problem by removing content above `## Props`

Ohh that is awesome. Got your point.

PR #3322, opened 2022-01-31
fix(FAB): add small size style 

sumj25:
What kind of change does this PR introduce?

Fix Fab button with title size bug.

Summary
Fab button with the title does not change on size small. Fix this issue.

Other information


arpitBhalla:
Thanks for your contribution :tada:

PR #3397, opened 2022-03-17
fix(Tab, TabView): conditional rendering

### LINKED ISSUE #3390: Conditional rendering Tab.Item / TabView.Item

I'm using the tab inside a screen, I need to hide a certain tab item based on some condition. When inside the <Tab> and I render the <Tab.Item> conditionally, it will pop up an error saying that React.cloneElement(...): The argument must be a React element, but you passed null.

I hope the Tab.Item could support the conditional rendering. Thx.

My code is like: 
```
<Tab>
    <Tab.Item title="item1" />]
    {condition && <Tab.Item title="item2" />}
    <Tab.Item title="item3" />
</Tab>
```

The tab item / tab view item can be conditionally rendered.

I have to based on the condition to render different <Tab> component, which needs to duplicate a lot of codes.

**Additional context**

WVAviator:
## Motivation

Fixes #3390

## Type of change

# How Has This Been Tested?

- [x] Jest Unit Test

## Checklist

## Additional context

Switching to the use of React.Children.toArray() instead of React.Children.map() effectively ignores falsey children rather than throwing an error. However, assuming that a developer may want the ability to enable/disable tabs while the component is mounted created a couple other issues to solve:
1. The "length" value used in the onPanResponderRelease function continues to reference the old length value, which enables the user to navigate past the last TabView.Item if an item was removed, or prevents them from navigating to the last TabView.Item if an item is added. This was fixed by making the onPanResponderRelease function a useCallback function with a dependency on length, and switching the panResponder object from useRef to useMemo with a dependency on onPanResponderRelease
2. If the user were on the last TabView.Item when another item is conditionally removed, they would be stuck past the last navigable TabView.Item. This was fixed by adding a useEffect function with a dependency on length that verifies whether the user's currentIndex is greater than the length - 1, and fixes it if not.
3. Changes in the number of TabView.Items at indices less than the user's currentIndex will cause their TabView to change since all views will shift downward. This cannot be done from within the TabView component since there's no real way to know how the user's index should change from there. The users index will need to be managed outside the TabView component alongside the index state. This may require an update to the documentation, which I will be happy to add - just let me know.


packages/base/src/TabView/TabView.tsx
```diff
@@ -56,33 +56,41 @@ export const TabViewBase: RneFunctionComponent<TabViewBaseProps> = ({
...
+  useEffect(() => {
+    if (currentIndex.current > length - 1) {
+      onChange(length - 1);
     }
-    if (Math.abs(dy) > Math.abs(dx)) {
-      return;
-    }
-    const position = dx / -window.width;
-    const next = position > value ? Math.ceil(position) : Math.floor(position);
-    onChange?.(currentIndex.current + next);
```
WVAviator, line 77: For some reason this logic broke with conditional children. I fixed by simply adding or subtracting 1 from the index based on the sign of dx

packages/base/src/TabView/TabView.tsx
```diff
@@ -56,33 +56,41 @@ export const TabViewBase: RneFunctionComponent<TabViewBaseProps> = ({
...
+        (dx > 0 && currentIndex.current <= 0) ||
+        (dx < 0 && currentIndex.current >= length - 1)
+      ) {
+        return;
+      }
+      if (Math.abs(dy) > Math.abs(dx)) {
+        return;
+      }
+
+      onChange?.(currentIndex.current + (dx > 0 ? -1 : 1));
```
WVAviator, line 81: This works just as well as the previous logic (as tested in the example app) and is able to support conditional rendering.

packages/base/src/TabView/TabView.tsx
```diff
@@ -56,33 +56,41 @@ export const TabViewBase: RneFunctionComponent<TabViewBaseProps> = ({
...
-    { dx, dy }: PanResponderGestureState
-  ) => {
-    if (
-      (dx > 0 && currentIndex.current <= 0) ||
-      (dx < 0 && currentIndex.current >= length - 1)
-    ) {
-      return;
+  useEffect(() => {
+    if (currentIndex.current > length - 1) {
+      onChange(length - 1);
```
WVAviator, line 65: This keeps the user from getting stuck beyond the last view if length changes

packages/base/src/TabView/TabView.tsx
WVAviator: After a second look, the original code was setting "value" only once upon initial render when panResponder was initialized with useRef. Since value was, by default, set to 0 - this code was effectively doing the same thing as my fix. This code could have caused bugs in its current state - for example, if the developer mounted the component with a starting value for "value" of a number greater than 0, it could have prevented the index from being incremented since position would never be greater than value. The reason it broke with dynamic children was because "value" was permitted to update whenever the dependency on length was added. 

Would it be worthwhile to consider added a minimum dx threshold to change TabViews? Right now, as long as Abs(dx) is greater than Abs(dy), even if it's a very small number, a pan will always initiate the view change. Is that desirable? Or should we be checking if the pan is more than, say, 25% of the screen width?

arpitBhalla: I am making changes here https://github.com/react-native-elements/react-native-elements/pull/3380. Feel free to review it.

WVAviator: Looks great - exactly what I was thinking of. It looks like to implement my fix in your new code, `React.Children.count(children)` on line 76 will need changed to `React.Children.toArray(children).length`, and that same value will need to be included ~~in the dependency array on 101 as well~~ as a dependency for  panResponder which means converting it from useRef to useMemo.

Nice work.

arpitBhalla: Hey, no issue, would merge your PR first

arpitBhalla:
@WVAviator Can you merge latest `next`

WVAviator:
> @WVAviator Can you merge latest `next`

All set - tested again as well. Decided to omit the useEffect statement I had created before that kept the user from getting stuck past the last TabView - the index really needs to be managed in the parent component if/when the number of Views changes so I decided it wouldn't be a good idea to hardcode that behavior.

arpitBhalla (approved):

arpitBhalla:
Thank you for your contribution 🎉

PR #3461, opened 2022-04-15
Update pitfall.md to show Example 3 fails as expected

jamonholmgren:
The final comment is incorrect -- example 3 is intended to show that it does not work with mixed styles.


arpitBhalla:
I believe it must work 🤔

PR #3473, opened 2022-04-26
fix(makeStyles): return theme that is extended

lucksp:
## Motivation

As reported in https://github.com/react-native-elements/react-native-elements/issues/3472, when creating an extended theme, `makeStyles` does not return the full extended them **at run time**.  

This is a followup to https://github.com/react-native-elements/react-native-elements/pull/3450.

Fixes # (issue) https://github.com/react-native-elements/react-native-elements/issues/3472

## Type of change

# How Has This Been Tested?

- [ ] Jest Unit Test - unable to verify with Jest using a custom `themed.d.ts` file for a test

## Checklist

- [ ] I have made corresponding changes to the documentation using `yarn docs-build-api`. ---  n/a
- [ ] I have added tests that prove my fix is effective or that my feature works --- see note above

### DIFF
--- a/packages/themed/src/config/makeStyles.ts
+++ b/packages/themed/src/config/makeStyles.ts
@@ -30,9 +30,7 @@ export const makeStyles =
 
     return useMemo(() => {
       const css =
-        typeof styles === 'function'
-          ? styles({ colors: theme.colors, mode: theme.mode }, props)
-          : styles;
+        typeof styles === 'function' ? styles({ ...theme }, props) : styles;
       return StyleSheet.create(css);
-    }, [props, theme.colors, theme.mode]);
+    }, [props, theme]);
   };
--- a/website/docs/installation.mdx
+++ b/website/docs/installation.mdx
@@ -58,7 +58,7 @@ npm install react-native-elements/react-native-elements#themed
 
 ```bash
 # @rneui/base
-yarn add @rneui/themed@github:react-native-elements/react-native-elements#base
+yarn add @rneui/base@github:react-native-elements/react-native-elements#base
 
 # @rneui/themed
 yarn add @rneui/themed@github:react-native-elements/react-native-elements#themed
--- a/website/versioned_docs/version-4.0.0-rc.1/installation.mdx
+++ b/website/versioned_docs/version-4.0.0-rc.1/installation.mdx
@@ -58,7 +58,7 @@ npm install react-native-elements/react-native-elements#themed
 
 ```bash
 # @rneui/base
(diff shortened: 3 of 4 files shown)


packages/themed/src/config/makeStyles.ts
```diff
@@ -30,9 +30,7 @@ export const makeStyles =
 
     return useMemo(() => {
       const css =
-        typeof styles === 'function'
-          ? styles({ colors: theme.colors, mode: theme.mode }, props)
```
lucksp, line 34: when making a custom theme, the theme shape could look like, for example:

```
colors: {...default RNE colors},
mode: {...default RNE mode},
myCustomColors: {
  text: {
    primary: string;
    secondary: string;
  }
}
```

when consuming `makeStyles` for a component, the current implementation of the `styles()` here ignores any custom fields that have been extended from the `FullTheme`.

This change proposes returning the full `theme` so that users have access to their custom/extended themes.

arpitBhalla: We can extend the `Colors` itself, IMO, having complete theme object is not necessary for `makeStyles`
Use `never` to remove existing colors if they aren't required.

lucksp: Can you explain further how this would work? 

I ask because when we define an extended theme (as per https://github.com/react-native-elements/react-native-elements/pull/3450), in our new `themed.d.ts` file, the custom "theme" name is not a known name (unless RNE adds a specific field to be used for this purpose).  I couldn't think how else to guarantee a return of an unknown field name, except to return the entire theme, which is extended off of `FullTheme`:

```
declare module '@rneui/themed' {

  export interface FullTheme {
    colors: RecursivePartial<Colors>;
    fluidColors: ColorShape;
  }

}
```

As you can see below, `makeStyles()` does not return the full theme since it only returns `theme.colors, mode: theme.mode` and is missing `theme.fluidColors`

image

arpitBhalla: Can you try https://github.com/react-native-elements/react-native-elements/issues/3472#issuecomment-1110752617 ?

arpitBhalla: I understood your point, but what I want to say is, having complete theme object will also contain default Component Props, which I think is overkilling the makestyles, because we only need colors/mode for styling, I believe. Also in your case we only need to add some more colors which can be done by extending `Colors`.

lucksp: https://github.com/react-native-elements/react-native-elements/issues/3472 solves the issue.  thank you

PR #3608, opened 2022-08-02
add children to BottomSheetProps

jeton-th:
## Motivation

Fixes # (issue)

## Type of change

# How Has This Been Tested?

## Additional context

Add `children` as an optional prop in `BottomSheetProps` to avoid typescript complaining about the non-existent property.


arpitBhalla:
Already fixed by #3583

PR #3648, opened 2022-10-01
feat: add `@rneui/layout`

### LINKED ISSUE #3771: Stack Component in Documentation
As [discussed](https://github.com/react-native-elements/react-native-elements/discussions/3627) it would be great to have direction on how to use the `Stack` component (which is used used in the documentation) as users are looking to use it.

An [explanation](https://github.com/react-native-elements/react-native-elements/discussions/3627#discussioncomment-3702065) on how to access the component clearly stated in the docs if it's not available as an export.

**Additional context**

arpitBhalla:
[npm](https://www.npmjs.com/package/@rneui/layout)

[Snack](https://snack.expo.dev/@arpitbhalla/rne-layout)


bartvanandel:
We've been using that package in a WIP app, but we switched to a custom View component instead after running into too many issues with this package. I haven't kept a list unfortunately, but we did find that changing styling of a parent component could cause items to disappear from the screen when using `<Stack />` for instance. I wouldn't consider this production ready (hence the alpha in the version, presumably).

One reason for adding this package though would be that some of the examples rely on it, and don't actually work without it currently.

arpitBhalla:
Hey @bartvanandel, I just release another alpha for this pkg, would it be possible for you to test `Stack` and `Box` component?

PR #3708, opened 2022-12-12
fix(withTheme): refs on functional components wrapped

marcshilling:
## Motivation

My team was attempting to attach a `ref` to one of our custom components that is wrapped with the `withTheme` higher order component, but it wasn't working. Upon digging into the HOC, I quickly noticed that react-native-elements is only doing ref forwarding for class components, but seemingly not for functional components. I don't see any reason for this, and after patching this change locally everything works fine for us and we are now able to use `ref`s on our components wrapped with `withTheme`.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [ ] Jest Unit Test

## Checklist


arpitBhalla (approved):

pranshuchittora (approved):
Overall looks good to me. Haven't run the changes locally. Merge it, if you have tested this. 
Also can we write some updated tests for this, probably some snapshot tests.

Issue #3655, opened 2022-10-20
ListItem backgroundColor persists after it's changed in the theme


Nantris:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Explain what you did

Define a backgroundColor for `ListItem.containerStyle` like below:

```js
const rneTheme = createTheme({
  components: {
    ListItem: {
      containerStyle: {
        backgroundColor: background,
      },
    },
  }
})
<ThemeProvider theme={rneTheme}>{children}</ThemeProvider>
```

### Expected behavior

When the color in the theme changes, the `backgroundColor` for `ListItem` should change, but instead it persists in the old color.

### Describe the bug

The `backgroundColor` for `ListItem` fails to update when the theme is updated. This can be easily tested in any setup that uses theming and defines separate colors for `light`/`dark` modes.

In our case, we noticed the issue occurred when toggling from `light` to `dark`, or the reverse. The color will remain as it was in the originally applied theme, never changing.

### Steps To Reproduce

```Markdown
const rneTheme = createTheme({
  components: {
    ListItem: {
      containerStyle: {
        backgroundColor: background,
      },
    },
  }
})
<ThemeProvider theme={rneTheme}>{children}</ThemeProvider>
```

And then use a `ListItem` somewhere, and toggle dark/light mode in your app.

### Screenshots

Light mode colors persist on `ListItem` even in dark mode. If app is started in dark mode, then the problem is reversed.

![image](https://user-images.githubusercontent.com/6835891/196827750-0f70e60f-5983-4174-9857-46f0b6fd3319.png)

### Your Environment

<details>
<summary>`npx @rneui/envinfo`</summary>

      ```
        System:
    OS: Linux 5.15 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
    CPU: (4) x64 AMD Ryzen 7 5800X 8-Core Processor
    Memory: 2.08 GB / 19.51 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 16.17.0 - ~/.nvm/versions/node/v16.17.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.17.0/bin/yarn
    npm: 8.15.0 - ~/.nvm/versions/node/v16.17.0/bin/npm
    Watchman: 4.9.0 - /usr/bin/watchman
  Managers:
    Apt: 2.4.7 - /usr/bin/apt
[...]

Issue #3665, opened 2022-10-25
Button styles wrongly persist after toggling from light to dark, or dark to light


Nantris:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Explain what you did

Theming buttons with the theme provider and toggling the dark/light mode.

### Expected behavior

The colors should be changed to those defined under `theme.components`

### Describe the bug

Same bug as #3655, but for buttons

### Steps To Reproduce

```Markdown
Add a button with styles defined in the RneTheme, then toggle dark/light mode and note that the styles do not update
```

### Screenshots

Toggling from light mode to dark mode, or reverse, persists the originally loaded styles for background and text colors

## Correct button light mode

![image](https://user-images.githubusercontent.com/6835891/197883102-dfdf746f-dc11-4210-bed7-83d55850587a.png)

## Incorrect button dark mode (after loading app in light mode and switching to dark)
![image](https://user-images.githubusercontent.com/6835891/197883180-c707c0eb-8b98-4c80-ba62-14f7945da41f.png)

----

## Correct button dark mode

![image](https://user-images.githubusercontent.com/6835891/197883329-c54a39a1-4988-4980-8e37-98c808faf02a.png)

## Incorrect button light mode (after loading app in dark mode and switching to light)

![image](https://user-images.githubusercontent.com/6835891/197883447-3d4a7961-cf47-4018-86df-9bd0f5172e0b.png)

### Your Environment

<details>
<summary>`npx @rneui/envinfo`</summary>

      ```
        System:
    OS: Linux 5.15 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
    CPU: (4) x64 AMD Ryzen 7 5800X 8-Core Processor
    Memory: 2.08 GB / 19.51 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 16.17.0 - ~/.nvm/versions/node/v16.17.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.17.0/bin/yarn
    npm: 8.15.0 - ~/.nvm/versions/node/v16.17.0/bin/npm
    Watchman: 4.9.0 - /usr/bin/watchman
  Managers:
    Apt: 2.4.7 - /usr/bin/apt
    pip3: 22.2.2 - ~/.local/bin/pip3
  Utilities:
    CMake: 3.22.1 - /usr/bin/cmake
[...]

PR #3732, opened 2023-01-07
fix(ThemeProvider): update theme with prop change

arpitBhalla:
Related #3665


Nantris:
Thanks again for all your efforts @arpitBhalla. I think that the `ThemeProvider` changes seem good to go in my view.

If merged, it should resolve: #3655 and #3665

Nantris:
@arpitBhalla I just wanted to thank you again for your great work. Thanks so much for what you do!

Nantris:
@arpitBhalla is it planned to merge this? It's seemed ready to go for many months but it still isn't merged, not to mention released. Could you provide some clarity?

example/App.tsx
```diff
@@ -5,12 +5,15 @@ import { cacheImages, cacheFonts } from './src/helpers/AssetsCaching';
...
 import { SafeAreaProvider } from 'react-native-safe-area-context';
 import * as SplashScreen from 'expo-splash-screen';
+import { useColorScheme } from 'react-native';
 
 SplashScreen.preventAutoHideAsync();
 
 export default () => {
   const [isReady, setIsReady] = useState(false);
 
+  theme.mode = useColorScheme();
```
khushal87: Don't store the result of `useColorScheme()` directly to theme.mode. Rather do this:

```suggestion
  const colorScheme = useColorScheme();
  theme.mode = colorScheme;
```

This looks better, imo.

Nantris:
Friendly bump.

I think the code changes look good. I noticed there's been no new release candidate for a while. Is it planned to release one soon?

Thank you again for all of your hard work, and please don't mistake my bump for impatience.

@arpitBhalla

markrickert:
This PR broke my `updateTheme()` functionality when upgrading from `rc.7` to `rc.8`.

I'm using a hook to determine if fonts should be scaled in the app:
```jsx
// These constants determine mow much bigger the font size should get based on the user's
// accessibility settings. Even if they turn the dial all the way to 11, we will only ever
// scale the fonts by these factors. This is to prevent the font size from getting too large
// and completely breaking the layout.
export const MAX_FONT_SCALE = 1.2;
export const MIN_FONT_SCALE = 0.8;

// Returns fontScaling props for Text and TextInput components
export const useFontScaling = (): Partial<TextProps> => {
  const { preferenceStore } = useStores();

  const fontScaling: Partial<TextProps> = React.useMemo(() => {
    const afs = preferenceStore?.allowFontScaling || false;
    return {
      minimumFontScale: afs ? MIN_FONT_SCALE : 1, // This prevents the font from getting too small.
      maxFontSizeMultiplier: afs ? MAX_FONT_SCALE : 1, // This prevents the font from getting too big.
      allowFontScaling: afs, // This allows the font to be scaled.
    };
  }, [preferenceStore?.allowFontScaling]);

  return fontScaling;
};
```

Then whenever it changes, i try and update the current theme:

```jsx
const { updateTheme } = useTheme();
const fontScalingProps = useFontScaling();

React.useEffect(()=>{
  updateTheme({
    components: {
      Text: {
        ...fontScalingProps,
      },
      // Other rneui components as necessary
    }
  });
},[fontScalingProps]);
```

`rc.8` completely broke this functionality because of this PR. I manually reverted changes in the `ThemeProvider.tsx` and the updates resumed working as intended.

Am I updating the theme incorrectly or is this an actual regression? Some guidance would be appreciated, thanks!

Nantris:
@markrickert it happens because you need to maintain object equality for the theme object after this PR (which if I'm not mistaken is how version 3.x handled it)

An approach to handling this might be like:

```
const [theme, setTheme] = React.useState({
    components: {
      Text: {
        ...fontScalingProps,
      },
      // Other rneui components as necessary
    }
});

React.useEffect(() => {
    const updatedTheme = {
        components: {
            Text: {
                ...fontScalingProps,
            },
            // Other rneui components as necessary
        }
    };
    
    setTheme(updatedTheme);
    updateTheme(updatedTheme);
}, [fontScalingProps]);
```

PR #3831, opened 2023-08-26
fix: multiple components render issue in Avatar

### LINKED ISSUE #3824: Both Icon and image show up in the picture
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Explain what you did

Adding the Avatar component

### Expected behavior

The Avatar should show an icon if there is no image. When an image is added as a uri through the source prop, the icon should go away.

### Describe the bug

I was testing the Avatar component in my react native app and I coded something like this:

```
           <Avatar
              source={{ uri: imageUri }}
              size="large"
              rounded
              icon={{ name: 'user', type: 'font-awesome', color: colors.gray40 }}
              containerStyle={{ backgroundColor: colors.gray10 }}
            />
```

However, the icon does not go away, and so what I see is both the avatar and the image squishing into the avatar container itself

Note, this was working a couple of days ago, but I had to do a fresh install of all my dependencies and suddenly I'm seeing this. When I remove the icon prop, the image shows properly, but according to the docs, the icon should only serve as a placeholder when using the source prop. 

### Steps To Reproduce

```Markdown
1. import { Avatar } from '@rneui/themed'

2. Add Avatar component with both an icon prop and a source prop

3. Ensure the source prop has a uri to the image file
```

### Screenshots

image

### Your Environment

<details>
<summary>`npx @rneui/envinfo`</summary>

      ```
        Output from `npx @rneui/envinfo` goes here.
      ```

</details>

khushal87:
## Motivation

Fixes #3824 

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [ ] Jest Unit Test

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings

### DIFF
--- a/packages/base/src/Avatar/Avatar.tsx
+++ b/packages/base/src/Avatar/Avatar.tsx
@@ -169,6 +169,50 @@ export const Avatar: RneFunctionComponent<AvatarProps> = ({
     delete imageProps.containerStyle;
   }
 
+  let componentToRender;
+
+  if (source) {
+    componentToRender = (
+      <Image
+        testID="RNE__Avatar__Image"
+        containerStyle={imageContainerStyle as StyleProp<TextStyle>}
+        source={source}
+        borderRadius={rounded ? width / 2 : undefined}
+        {...imageProps}
+        style={StyleSheet.flatten([
+          styles.avatar,
+          imageProps && imageProps.style,
+          avatarStyle,
+        ])}
+        ImageComponent={ImageComponent}
+      />
+    );
+  } else if (title) {
+    componentToRender = (
+      <Text
+        style={StyleSheet.flatten([
+          styles.title,
+          { fontSize: titleSize },
+          titleStyle,
+        ])}
+      >
+        {title}
+      </Text>
+    );
+  } else if (icon) {
+    componentToRender = (
+      <Icon
+        style={StyleSheet.flatten([iconStyle])}
+        color={icon.color || 'white'}
+        name={icon.name || 'account'}
+        size={icon.size || iconSize}
+        type={icon.type || 'material-community'}
+      />
+    );
+  } else if (renderCustomContent) {
+    componentToRender = renderNode(undefined, renderCustomContent);
+  }
+
   return (
     <Component
       style={StyleSheet.flatten([
@@ -186,42 +230,7 @@ export const Avatar: RneFunctionComponent<AvatarProps> = ({
         ...rest,
       }}
     >
-      {renderCustomContent && renderNode(undefined, renderCustomContent)}
-      {icon && (
-        <Icon
-          style={StyleSheet.flatten([iconStyle])}
-          color={icon.color || 'white'}
-          name={icon.name || 'account'}
-          size={icon.size || iconSize}
-          type={icon.type || 'material-community'}
-        />
-      )}
-      {title && (
-        <Text
-          style={StyleSheet.flatten([
-            styles.title,
-            { fontSize: titleSize },
-            titleStyle,
-          ])}
-        >
-          {title}
-        </Text>
-      )}
-      {source && (
-        <Image
-          testID="RNE__Avatar__Image"
-          containerStyle={imageContainerStyle as StyleProp<TextStyle>}
-          source={source}
-          borderRadius={rounded ? width / 2 : undefined}
-          {...imageProps}
-          style={StyleSheet.flatten([
-            styles.avatar,
-            imageProps && imageProps.style,
-            avatarStyle,
-          ])}
-          ImageComponent={ImageComponent}
-        />
-      )}
+      {componentToRender}
       {children}
     </Component>
   );
@@ -229,8 +238,6 @@ export const Avatar: RneFunctionComponent<AvatarProps> = ({
 
 const styles = StyleSheet.create({
   container: {
-    display: 'flex',
-    flexDirection: 'column',
     justifyContent: 'center',
   },
   avatar: {
--- a/packages/base/src/Avatar/Avatar.tsx
+++ b/packages/base/src/Avatar/Avatar.tsx
@@ -222,6 +222,10 @@ export const Avatar: RneFunctionComponent<AvatarProps> = ({
           )) ||
           (icon && <AvatarIcon icon={icon} iconStyle={iconStyle} size={size} />)
         }
+        placeholderStyle={StyleSheet.flatten([
+          styles.placeholderStyle,
+          imageProps && imageProps.placeholderStyle,
+        ])}
         borderRadius={rounded ? width / 2 : undefined}
         {...imageProps}
         style={StyleSheet.flatten([
@@ -282,6 +286,9 @@ const styles = StyleSheet.create({
     backgroundColor: 'transparent',
     textAlign: 'center',
     zIndex: 1,
+  },
+  placeholderStyle: {
+    backgroundColor: 'transparent',
   },
 });
 
--- a/packages/base/src/Avatar/Avatar.tsx
+++ b/packages/base/src/Avatar/Avatar.tsx
@@ -177,6 +177,17 @@ export const Avatar: RneFunctionComponent<AvatarProps> = ({
         testID="RNE__Avatar__Image"
         containerStyle={imageContainerStyle as StyleProp<TextStyle>}
         source={source}
+        PlaceholderContent={
+          icon && (
+            <Icon
+              style={StyleSheet.flatten([iconStyle])}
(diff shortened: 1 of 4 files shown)


packages/base/src/Avatar/Avatar.tsx
```diff
@@ -169,6 +169,50 @@ export const Avatar: RneFunctionComponent<AvatarProps> = ({
     delete imageProps.containerStyle;
   }
 
+  let componentToRender;
+
+  if (source) {
```
haeniya, line 213: If I unserstand correctly, in 4.0.0-rc.7, the title (if present) was shown in two different situations:
1. As placeholder content when the image was rendering or could not be loaded
2. As a fallback, if there was no image source passed (e.g. when the image source is undefined)
See here: https://github.com/react-native-elements/react-native-elements/blob/v4.0.0-rc.7/packages/base/src/Avatar/Avatar.tsx#L206-L230

I wonder if this behavior is still the same with the latest changes.

There is a workaround for this (https://github.com/react-native-elements/react-native-elements/issues/3824#issuecomment-1694228942) but maybe we could keep the behavior the same? If this in on purpose and there are reasons to have it that way, it's fine for me. Just wanted to point that out.

khushal87: Hey @haeniya I totally understand your concern. I was checking the implementation of Placeholder content but regardless of this change, it doesn't seem to render the content even if I pass a placeholder content directly to the Image component(I doubt this ever worked before). The issue might be different and can be addressed differently.

This PR sought to fix the multiple components render issue(that was a mistake) and rendering placeholder still remains an issue.

khushal87: I think I would say using the imageProps would be a more viable option since we have it in Avatar. Users can use the `imageProp` and pass PlaceholerContent into it. No point in introducing a new prop in avatar for the same.

haeniya: It worked in 4.0.0-rc.7 but doesn't in the 4.0.0-rc.8 release. But you're right, it could be handled in a different issue/PR. I was just wondering if we need to refactor the code anyway, why not fix that too (to maybe speed up the process a bit).

> I think I would say using the imageProps would be a more viable option since we have it in Avatar. Users can use the `imageProp` and pass PlaceholerContent into it. No point in introducing a new prop in avatar for the same.

I think the idea would not be to introduce a new prop but to make the `title` prop the `PlaceholderContent` if present. The linked workaround works for me but I need to put the title twice, as `PlaceholderContent` and as `title`. Was quite convenient to have this fallback automatically.

khushal87: You want the `title` to be the placeholder or the `icon`? I think the document stated the icon to be the placeholder.

khushal87: Hey @haeniya, I have pushed the `icon` as the placeholder component as per the docs. Hope that helps. Please review the PR, if you can and let me know if it works. 😄

haeniya: Thanks for the update. Before, it was either the icon or the title that was used as a placeholder depending on what was passed. They couldn't be passed together as the docs state https://reactnativeelements.com/docs/components/avatar#props. Would be nice to have it the same.

khushal87: Done ✅

haeniya: Great thanks a lot. Will test it tomorrow.

haeniya: Hi @khushal87, thanks for the update. I just tested it and it looks good in general. 👍 The title is now correctly used as placeholder if the image is loading or can not be loaded. However, there is a styling issue:
image
If the `title` is used as placeholder, the background is different to when there is no `source` passed.


**How to reproduce in the example app:**
In the `dataList` where all the image URLs are defined, change one URL to empty string and another to an not existing URL (just add some text to the URL string).
Below, where the avatars are rendered, use the following code:
```
<Avatar
  size={64}
  rounded
  title="T"
  containerStyle={{ backgroundColor: 'grey' }}
  source={l.image_url ? { uri: l.image_url } : undefined}
  key={`${chunkIndex}-${i}`}
/>
```

**How to fix it:**
This happens because there is a default styling for the placeholder content in the `Image` component:
https://github.com/react-native-elements/react-native-elements/blob/next/packages/base/src/Image/Image.tsx#L150

To fix it, we need to override this default styling to `backgroundColor: 'transparent'` so that the background color that is passed as `containerStyle` is correctly used.

```
<Image
  ...
  placeholderStyle={StyleSheet.flatten([
    styles.placeholderStyle,
    imageProps && imageProps.placeholderStyle
  ])}
  ...
/>

  ...
  placeholderStyle: {
    backgroundColor: 'transparent',
  },
});
```

You might need to reload the app to see the changes. Let me know if that makes sense for you or if you see any issues with this. Thanks.

khushal87: Simulator Screenshot - iPhone SE (3rd generation) - 2023-08-27 at 21 52 56

Hey @haeniya, for me it's the same. Not sure if I am doing something wrong.

haeniya: Just tested again. Having one `image_url` in `dataList` as empty string, another one as for example "https://cdn.pixabay.com/photo/2019/11/03/20/11/portrait-4599523453__340.jpg" and then later render the Avatar like this:
```
<Avatar
  size={64}
  rounded
  title="T"
  containerStyle={{ backgroundColor: 'grey' }}
  source={l.image_url ? { uri: l.image_url } : undefined}
  key={`${chunkIndex}-${i}`}
/>
```
results in the follwing:
image
Have you tried reloading the app? I sometimes needed to do this, not sure why.

khushal87: Yup this is how my list looks like:
```
{
    image_url: '',
  },
  {
    image_url:
      'https://cdn.pixabay.com/photo/2019/11/03/20/11/portrait-4599523453__340.jpg',
  },
```

And yes, I reloaded

khushal87: Still couldn't reproduce it:
```
{
    image_url: '',
  },
  {
    image_url:
      'https://cdn.pixabay.com/photo/2019/11/03/20/11/portrait-4599523453__340.jpg',
  },
```

haeniya: And you use this exact code on line 54 in `avatars.tsx` to render the Avatar?
```
<Avatar
  size={64}
  rounded
  title="T"
  containerStyle={{ backgroundColor: 'grey' }}
  source={l.image_url ? { uri: l.image_url } : undefined}
  key={`${chunkIndex}-${i}`}
/>
```
That's really weird.

haeniya: I guess something needs to be wrong. Those two changes are the only changes I do to reproduce it (image_urls and Avatar). If you add these changes and you're on the correct branch, you should see the difference. I'm also able to reproduce it in my own app with the current 4.0.0-rc.8 release. The issue makes sense to me as stated here: https://github.com/react-native-elements/react-native-elements/pull/3831#discussion_r1306661931.

Also, if you use `containerStyle={{ backgroundColor: 'grey' }}` on the Avatar, the correct color should look like this:

image

It seems to be a lot lighter in your screenshot.

khushal87: Hey @haeniya, I am sorry for the confusion. I was luckily able to reproduce it now and I have pushed the fix. Hope that helps. Thanks 😄

khushal87: @haeniya if you have tested it well, can we merge it?

haeniya: @khushal87 I don't see any updates on the PR. Have you pushed your changes?

khushal87: Oh sorry my bad @haeniya, I thought I pushed it but I missed it somehow. Now I have done. Thanks 😄

haeniya: @khushal87 I just tested it and it looks good to me. Thanks for the update. 👍

l2aelba:
How to install or update to this fix ? @khushal87

PR #3833, opened 2023-08-29
feat(SearchBarIos): new prop option `omitCancel`

coderpr0grammer:
## Motivation

I'm always frustrated when the showCancel Prop (for platform="ios" only) on the SearchBar component from @rneui/base only defines whether or not to persist the cancel button beside the searchbar when unfocused. I wanted to see an option that let the developer omit the cancel button from the SearchBar, resulting in a cleaner UI when the button is not necessary. I created a new prop called `omitCancel`:`boolean` that defaults as `false`. When `omitCancel={true}`, it will completely omit the cancel button on SearchBars with `platform="ios"`, so it will never appear regardless of it being focused or not.

Fixes: Creates new prop called `omitCancel`:`boolean` that defaults as `false`

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# How Has This Been Tested?

- [x] Checked with my own app

## Checklist

- [x] I have made corresponding changes to the documentation using `yarn docs-build-api`

## Additional context
In the below image, the cancel button is omitted while maintaining the rest of the native iOS styles.
image


coderpr0grammer (commented):
I'm not sure if we are supposed to also modify the documentation pages that are in the repo (which is what I did), I'm not sure if this may have caused it not to pass some checks by Netlify. Please let me know if there is anything wrong with the PR!

pranshuchittora:
I feel adding more props would lead to more confusion. I would rather update the current showCancel prop to accept more values with current boolean value.

PR #3859, opened 2023-11-17
fix(makeStyles): make props mandatory when declared non-empty

darkbasic:
## Motivation

Currently `makeStyles` always returns a function which optionally requires `props`.
I think this is a typings bug because `props` should be required if the props are needed to create the styles.
With my change if you specify a non empty type for the props the argument of the returned function would be required, otherwise optional.

```ts
export interface MyComponentStyleProps {}
export interface MyComponentProps extends MyComponentStyleProps {
  [...]
}

const MyComponent: FC<MyComponentProps> = ({[...]}) => {
  const styles = useStyles();

  return (
    <AnotherComponent />
  );
};

const useStyles = makeStyles((theme, props: MyComponentStyleProps) => ({
  [...]
}));
```

Since `MyComponentStyleProps` is an empty object `props` would not be required.

```ts
const MyComponent: FC<MyComponentProps> = ({[...]}) => {
  const styles = useStyles();

  return (
    <AnotherComponent />
  );
};

const useStyles = makeStyles((theme, props) => ({
  [...]
}));
```

If you don't specify a type `props` wouldn't be required either.

```ts
export interface MyComponentStyleProps {
  customStyle: any;
}
export interface MyComponentProps extends MyComponentStyleProps {
  [...]
}

const MyComponent: FC<MyComponentProps> = ({
  customStyle,
  [...]
}) => {
  const styles = useStyles({customStyle});

  return (
    <AnotherComponent />
  );
};

const useStyles = makeStyles((theme, props: MyComponentStyleProps) => ({
  [...]
}));
```

Now you will have to specify `props` otherwise `useStyles` will error out.

I would have rather preferred to let the user specify the generic (ex. `makeStyles<MyComponentStyleProps>((theme, props)`) but that would force you to specify all generics so you won't be able to infer the stylesheet type for the second generic.
Once https://github.com/microsoft/TypeScript/pull/54047 gets merged it should be possible to do so.

## Type of change

# How Has This Been Tested?

- [ ] Jest Unit Test
- [x] Checked with production app

## Checklist

- [x] I have made corresponding changes to the documentation using `yarn docs-build-api`

### DIFF
--- a/packages/themed/src/config/makeStyles.ts
+++ b/packages/themed/src/config/makeStyles.ts
@@ -14,8 +14,8 @@ export const makeStyles =
           } & Theme,
           props: V
         ) => T)
-  ) =>
-  (props: V = {} as any): T => {
+  ): keyof V extends undefined ? (props?: V) => T : (props: V) => T =>
+  (props: V | undefined): T => {
     const { theme } = useTheme();
 
     return useMemo(() => {


darkbasic:
Please note that the build is currently broken in the next branch, see https://github.com/react-native-elements/react-native-elements/pull/3858

arpitBhalla:
Fixed by https://github.com/react-native-elements/react-native-elements/pull/3844. Let me know if it's working for the snippets you provided.

PR #3990, opened 2025-10-28
Fix: fractions Runtime Validation in SwipeRating

codewithshinde:
…Updated the fractions type to number

## Motivation
This section is to address the above note from Copilot:
Using @ts-ignore to suppress TypeScript errors is not ideal. The fractionsType function should be properly typed instead of being suppressed.

Location: packages/base/src/AirbnbRating/SwipeRating.tsx lines 56-73, 171

Issue: The [fractionsType] validator is never executed. The component uses fractions with .toFixed() which throws a RangeError if fractions < 0 or fractions > 100.

Fixes # (issue)

## Type of change

- [ x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [ ] Jest Unit Test
- [ x] Checked with `example` app

## Checklist

- [ x] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings

## Additional context

Screenshot 2025-10-27 at 7 14 32 PM


packages/base/src/AirbnbRating/SwipeRating.tsx
```diff
@@ -486,6 +485,10 @@ const SwipeRating: React.FC<SwipeRatingProps> = ({
   );
 };
 
+SwipeRating.propTypes = {
```
deepktp: https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-proptypes-and-defaultprops
Componenet.propTypes and Componenet.defaultProps are now removed in React 19. 

As you have used prop-types library,  is there a good option we can use other then this one.

codewithshinde: @deepktp , Thank you for the Insight on this.
I'll check and try to change the library to something other than prop-types. 
If not, I'll fall back to @theianmay mentioned, a solution using useEffect

theianmay:
Thanks for working on this PR! It has been correctly identified that the `fractionsType` validator was never being executed and removing it is the right approach. However, after further research/analysis, I have a few suggestions related to your current implementation:

## Analysis

### 1. Runtime validation doesn't prevent crashes
The `console.error` you added will log a warning, but it won't prevent the component from crashing. The actual crash occurs on lines 287-288 and 291-292 where `.toFixed(fractions)` is called:

```typescript
currentRating = fractions
  ? Number(currentRating.toFixed(fractions))
  : Math.ceil(currentRating);
```

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed), `toFixed()` throws a `RangeError` if the argument is not between 0 and 100. Your validation checks for 0-20, but even within that range, if someone passes a negative number or a value > 100, the component will crash.

### 2. Validation happens too late
The validation is inside a `useEffect` that runs after the component has already started using the `fractions` value. This means the crash could happen before the warning is logged.

## Suggested Fix

Instead of just logging an error, clamp the value to ensure it's always safe:

```typescript
const SwipeRating: React.FC<SwipeRatingProps> = ({
  // ... other props
  fractions = 0,
  // ... other props
}) => {
  // Clamp fractions to valid range to prevent toFixed() RangeError
  const safeFractions = Math.max(0, Math.min(20, fractions));
  
  // Then use safeFractions throughout the component instead of fractions
```

Then replace all uses of `fractions` in the component with `safeFractions`. You can still add a `console.warn` if the value was clamped:

```typescript
useEffect(() => {
  if (fractions < 0 || fractions > 20) {
    console.warn(
      `[SwipeRating] fractions must be between 0-20, received ${fractions}. Value has been clamped to ${safeFractions}.`
    );
  }
}, [fractions, safeFractions]);
```

This way, the component is defensive and won't crash even with invalid input, while still informing developers about the issue.

## Additional Notes
- The dependency array in the `useEffect` at line 258 should include `fractions` (which you've added, new line 244)
- Consider adding a test case for invalid `fractions` values to prevent regression, but please test in the example app after making these updates

theianmay (changes requested):
Thanks for working on this PR! It has been correctly identified that the `fractionsType` validator was never being executed and removing it is the right approach. However, after further research/analysis, I have a few suggestions related to your current implementation:

## Analysis

### 1. Runtime validation doesn't prevent crashes
The `console.error` you added will log a warning, but it won't prevent the component from crashing. The actual crash occurs on lines 287-288 and 291-292 where `.toFixed(fractions)` is called:

```typescript
currentRating = fractions
  ? Number(currentRating.toFixed(fractions))
  : Math.ceil(currentRating);
```

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed), `toFixed()` throws a `RangeError` if the argument is not between 0 and 100. Your validation checks for 0-20, but even within that range, if someone passes a negative number or a value > 100, the component will crash.

### 2. Validation happens too late
The validation is inside a `useEffect` that runs after the component has already started using the `fractions` value. This means the crash could happen before the warning is logged.

## Suggested Fix

Instead of just logging an error, clamp the value to ensure it's always safe:

```typescript
const SwipeRating: React.FC<SwipeRatingProps> = ({
  // ... other props
  fractions = 0,
  // ... other props
}) => {
  // Clamp fractions to valid range to prevent toFixed() RangeError
  const safeFractions = Math.max(0, Math.min(20, fractions));
  
  // Then use safeFractions throughout the component instead of fractions
```

Then replace all uses of `fractions` in the component with `safeFractions`. You can still add a `console.warn` if the value was clamped:

```typescript
useEffect(() => {
  if (fractions < 0 || fractions > 20) {
    console.warn(
      `[SwipeRating] fractions must be between 0-20, received ${fractions}. Value has been clamped to ${safeFractions}.`
    );
  }
}, [fractions, safeFractions]);
```

This way, the component is defensive and won't crash even with invalid input, while still informing developers about the issue.

## Additional Notes
- The dependency array in the `useEffect` at line 258 should include `fractions` (which you've added, new line 244)
- Consider adding a test case for invalid `fractions` values to prevent regression, but please test in the example app after making these updates

louveshh:
Thanks for the changes, @codewithshinde.

Applied requested changes in `packages/base/src/AirbnbRating/SwipeRating.tsx`:
- Clamp `fractions` to `safeFractions` (0-20) before any usage.
- Use `safeFractions` for all `toFixed()` paths to avoid RangeError.
- Warn when clamping and include the clamped value.
- Update `getCurrentRating` dependencies to use `safeFractions`.

Tests added:
- Out-of-range `fractions` warnings.
- Pan responder path with clamped values.

Tests run:
- `corepack yarn workspace @rneui/base test SwipeRating.test.tsx` (PASS)

Changed code:
```tsx
const SwipeRating: React.FC<SwipeRatingProps> = ({
  // ...
  fractions = 0,
  // ...
}) => {
  // Clamp fractions to valid range to prevent toFixed() RangeError
  const safeFractions = Math.max(0, Math.min(20, fractions));

  useEffect(() => {
    setCurrentRating(startingValue);
    if (fractions < 0 || fractions > 20) {
      console.warn(
        `[SwipeRating] fractions must be between 0-20, received ${fractions}. Value has been clamped to ${safeFractions}.`
      );
    }
  }, [startingValue, setCurrentRating, fractions, safeFractions]);

  const getCurrentRating = useCallback(
    (localValue: number) => {
      // ...
      currentRating = safeFractions
        ? Number(currentRating.toFixed(safeFractions))
        : Math.ceil(currentRating);
      // ...
      currentRating = safeFractions
        ? Number(localStartingValue.toFixed(safeFractions))
        : Math.ceil(localStartingValue);
      // ...
    },
    [ratingCount, minValue, imageSize, jumpValue, safeFractions]
  );

  const panResponderOnRelease = useCallback(
    () => {
      const rating = getCurrentRating(ratingBackdropValue.current);

      if (rating >= minValue) {
        if (!safeFractions) {
          // 'round up' to the nearest rating image
          setCurrentRating(rating);
        }
        onFinishRating(rating);
      }
    },
    // eslint-disable-next-line react-hooks/exhaustive-deps
    [] // [getCurrentRating, safeFractions, onFinishRating, setCurrentRating, minValue]
  );
```

Tests:
```tsx
import { act, render, waitFor } from '@testing-library/react-native';

// ...

it('should warn and clamp fractions above the allowed range', async () => {
  const warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
  const { getByTestId } = render(<SwipeRating fractions={200} />);
  const panResponderView = getByTestId('RNEUI__SwipeRating-pan');
  const gestureState = { x0: 0, dx: 0 };

  act(() => {
    panResponderView.props.onPanResponderMove?.({}, gestureState);
  });

  await waitFor(() => {
    expect(warnSpy).toHaveBeenCalledWith(
      expect.stringContaining('fractions must be between 0-20')
    );
  });

  warnSpy.mockRestore();
});

it('should warn when fractions is below the allowed range', async () => {
  const warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
  render(<SwipeRating fractions={-3} />);

  await waitFor(() => {
    expect(warnSpy).toHaveBeenCalledWith(
      expect.stringContaining('fractions must be between 0-20')
    );
  });

  warnSpy.mockRestore();
});
```

@theianmay

